PR #24, opened 2014-10-08
Increase throughput

fd0:


davecheney:
Hi,

Sorry for the delay.

There are a few problems, some minor, some not.

1. There is a merge conflict, I have no idea, why, could you re push this change.

2. I don't have an objection to this change, but I do think it should be expanded so that _all_ commands issued to the server are made asynchronously.

If you can fix the merge conflict I'll merge this.

davecheney:
Superseded by #41

PR #33, opened 2015-02-11
Make MaxReadPacket and MaxWritePacket configurable

DAddYE:
Some sftp servers seems they don't allow always the default value 32K so
in some case this must be set manually to a lower value.

### DIFF
--- a/client.go
+++ b/client.go
@@ -571,7 +571,7 @@ func (f *File) Close() error {
 func (f *File) Read(b []byte) (int, error) {
 	var read int
 	for len(b) > 0 {
-		n, err := f.c.readAt(f.handle, f.offset, b[:min(len(b), maxWritePacket)])
+		n, err := f.c.readAt(f.handle, f.offset, b[:min(len(b), MaxReadPacket)])
 		f.offset += uint64(n)
 		read += int(n)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


client.go
```diff
@@ -1,4 +1,4 @@
-package sftp
+package sftp // import "github.com/daddye/sftp"
```
shurcooL: This import path comment should not be a part of this PR.

client.go
```diff
@@ -592,8 +592,11 @@ func (f *File) Stat() (os.FileInfo, error) {
 	return fileInfoFromStat(fs, path.Base(f.path)), nil
 }
 
+// clamp reads to less than 32k
+var MaxReadPacket = 32 * 1024
+
 // clamp writes to less than 32k
-const maxWritePacket = 1 << 15
+var MaxWritePacket = 32 * 1024
```
shurcooL: Now that this variable is exported, it should be better documented.

The comment should begin with `// MaxWritePacket ...` (see [here](https://github.com/golang/go/wiki/CodeReviewComments#doc-comments)) and explain what it does, e.g., `// MaxWritePacket controls the maximum size in bytes of a single write packet.` or so.

shurcooL:
LGTM (fwiw).

I would imagine it's likely the package maintainers will want to ask you to squish the 3 commits into one.

davecheney:
Thanks for your proposal. I am sorry I have been so tardy in responding to it.

I'm sorry but this is not the correct way to make this parameter adjustable, package level variables like this would affect every instance of sftp.Client in the application, and changing those values will almost certainly result in a data race. 

These configuration parameters should be passed to the client on creation.

shurcooL:
I think this can be closed now, I'm pretty sure this functionality is no longer needed for its original purpose (unless I'm mistaken).

davecheney:
Superceeded by 33, if you want to split out MaxPacket into MaxReadPacket / MaxWritePacket, please send a PR.

shurcooL:
> Superceeded by 33

I think you meant #39.

PR #39, opened 2015-05-23
Add function options to sftp.NewClient and a MaxPacket option

davecheney:
Update #33

### DIFF
--- a/client.go
+++ b/client.go
@@ -5,6 +5,7 @@ import (
 	"encoding"
 	"encoding/binary"
 	"errors"
+	"fmt"
 	"io"
 	"os"
 	"path"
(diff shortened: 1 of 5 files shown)


iporer:
@davecheney One sftp server I used set the MTU to 16k, thus I will encounter "internal inconsistency" exception. 
Could remove the 32k minimum limit of maxPacket?  I can't use func MaxPacket to set the maxPacket to 16k until remove the the limit.

davecheney:
I’m not sure, from memory those values come from the sftp spec (such that it is) I don’t know what the implications of lowering those values would be.

puellanivis:
The eternal question, when presented with a standards-breaking implementation, do you support it anyways, or do you demand standards compliance. :(

puellanivis:
```   The maximum size of a packet is in practice determined by the client
   (the maximum size of read or write requests that it sends, plus a few
   bytes of packet overhead).  All servers SHOULD support packets of at
   least 34000 bytes (where the packet size refers to the full length,
   including the header above).  This should allow for reads and writes
   of at most 32768 bytes.
```

So, MaxPacket should not be actually limited to a minimum of 32 KiB, but rather every server should (but need not necessarily) support at least 34 kB in order to allow clients to more readily assume 32 KiB is acceptable as a MaxPacket size.

But in theory, a client could use really any max packet size that it wants… Since it is going to be facing overhead in picking a value too small, it is in the clients own self-interests to also pick a reasonable value. (I imagine a de facto minimum could be chosen, under which the pkg/sftp Server would simply refuse to accept.)

Issue #15, opened 2014-09-12
Slow downloads compared to FileZilla


vanbroup:
I noticed some serious difference when comparing the download speed via this package with the same download via FileZilla.

When downloading with Filezilla I can download at 30Mpbs+ with 10 concurrent files (over a VPN), but when downloading with the GO program directly on a high speed internet connection I can't get above the 20Mpbs with 16 concurrent downloads (or about 10 with 10 concurrent).

_The readme already lists that currently all traffic with the server is serialized and I can imagine that this is influencing the download performance. But maybe also other factors as packet size and or compression can play a role here._

The code that is facilitating the download:
```
log.Printf("[%s] Starting to download", f)

// check if file already exists
if _, err := os.Stat(file); !os.IsNotExist(err) {
	log.Printf("[%s] Already downloaded", f)
	return
}

// open local file to save download
out, err := os.Create(file)
if err != nil {
	log.Printf("[%s] Error (%s)", f, err.Error())
	return
}
defer out.Close()

// open file to download
dl, err := s.Open(f)
if err != nil {
	log.Fatal(err)
}
defer dl.Close()

// copy the bytes
nb, err := io.Copy(out, dl)
if err != nil {
	log.Println("error downloading file", err)

	// delete
	err = os.Remove(file)
	if err != nil {
		log.Println("error deleting file after download error", err)
	}
	return
}

log.Printf("[%s] Download completed (%d bytes)", f, nb)
```

davecheney:
I believe this was addressed in #41

PR #41, opened 2015-06-02
Increase throughput

ggriffiniii:
I was running into the same problem as described in #15 and it appears fd0's work in #24 has been dropped. I've created a new patchset that merges cleanly with current head and provides an ~20x throughput increase for large file transfers. I've included the benchmark output in each of the commit messages to give an idea what impact each individual change has.

### DIFF
--- a/client.go
+++ b/client.go
@@ -58,7 +58,7 @@ func NewClientPipe(rd io.Reader, wr io.WriteCloser, opts ...func(*Client) error)
 		w:         wr,
 		r:         rd,
 		maxPacket: 1 << 15,
-		inflight:  make(map[uint32]chan result),
+		inflight:  make(map[uint32]chan<- result),
 	}
 	if err := sftp.applyOptions(opts...); err != nil {
 		wr.Close()
@@ -83,8 +83,8 @@ type Client struct {
 	maxPacket int // max packet size read or written.
 	nextid    uint32
 
-	mu       sync.Mutex             // ensures only on request is in flight to the server at once
-	inflight map[uint32]chan result // outstanding requests
+	mu       sync.Mutex               // ensures only on request is in flight to the server at once
+	inflight map[uint32]chan<- result // outstanding requests
 }
 
 // Close closes the SFTP session.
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -127,6 +132,41 @@ func (c *Client) recvVersion() error {
 	return nil
 }
 
+func (c *Client) recv() {
```
davecheney: even though this is a private method, please add a comment as this is an important method

client.go
```diff
@@ -127,6 +132,41 @@ func (c *Client) recvVersion() error {
...
+	sendAll := func(res result) {
```
davecheney: please make this a method, and give it a comment. It probably shouldn't be called `sendAll`, maybe `demux` or `dispatch` would be better

client.go
```diff
@@ -127,6 +132,41 @@ func (c *Client) recvVersion() error {
...
+		c.mu.Lock()
+		listeners := make([]chan<- result, len(c.inflight))
+		for _, ch := range c.inflight {
+			listeners = append(listeners, ch)
+		}
+		c.mu.Unlock()
+		for _, ch := range listeners {
+			ch <- res
```
davecheney: does this need to have a timeout, or a be none initally non blocking ? If one listener is busy, the ones behind it will not receve this packet.

davecheney:
This is a really nice PR. Thank you for the effort and level of detail you put into this.

I started to make some minor comments on this PR, but I think it can probably land as is, and we can address them in followup PRs.

Do you want to tackle the small comments I made, or just land as is ?

davecheney:
Hmm. There is a data race in the close path, run `go test -race -integration` to see it
```
WARNING: DATA RACE
Write by goroutine 38:
  os.(*file).close()
      /home/dfc/go/src/os/file_unix.go:129 +0x1c1
  os.(*File).Close()
      /home/dfc/go/src/os/file_unix.go:118 +0x8b
  os/exec.(*Cmd).closeDescriptors()
      /home/dfc/go/src/os/exec/exec.go:232 +0xa2
  os/exec.(*Cmd).Wait()
      /home/dfc/go/src/os/exec/exec.go:377 +0x4c7
  github.com/pkg/sftp.TestClientReadLine()
      /home/dfc/src/github.com/pkg/sftp/client_integration_test.go:469 +0x5a6
  testing.tRunner()
      /home/dfc/go/src/testing/testing.go:454 +0xdf

Previous read by goroutine 39:
  os.(*File).read()
      /home/dfc/go/src/os/file_unix.go:211 +0x5a
  os.(*File).Read()
      /home/dfc/go/src/os/file.go:95 +0xbf
  io.ReadAtLeast()
      /home/dfc/go/src/io/io.go:298 +0x11b
  io.ReadFull()
      /home/dfc/go/src/io/io.go:316 +0x79
  github.com/pkg/sftp.recvPacket()
      /home/dfc/src/github.com/pkg/sftp/packet.go:85 +0x10d
  github.com/pkg/sftp.(*Client).recv()
      /home/dfc/src/github.com/pkg/sftp/client.go:148 +0xd1
```

kevinburke:
this is cool! just curious as someone who's unfamiliar with the code, what's the main difference that allows the speedup?

davecheney:
Overlapping requests.

ggriffiniii:
Thanks for the feedback. I'll address the comments and take a look at the data race (thanks for pointing that out) and get back to you.

davecheney:
The data race will be in close semantics shutting down the connectoin and
making sure that the recv() goroutine has quit before Close() returns.

We had this same problem in the ssh client, it's quite tricky to get these
semantics right.

client.go
ggriffiniii: Done

client.go
ggriffiniii: The inflight channels are all controlled privately and some care is taken to ensure that all are drained in a timely manner. I don't believe this needs to be non-blocking, but I could make an initial non-blocking pass through the listeners and followup with a blocking pass to catch all the remaining if you prefer.

ggriffiniii:
I've updated the PR addressing your comments and fixing the race on Close(). Let me know how it looks.

fd0:
Thanks @ggriffiniii for your work in getting this merged! I have a lot other things to do lately. :)

PR #46, opened 2015-09-07
sftp server implementation

marksheahan:
Mostly tested through integration tests, by using the client package and comparing behavior to the openssh sftp server.

Part of the server implementation of readdir involves implementing unix group lookup; gid -> groupname conversion for 'ls -l' style output. This is the user/ directory, which is os/user with this patchset applied: https://codereview.appspot.com/13454043


davecheney:
@marksheahan this looks awesome! Thank you for this.

I'll have a read through the diff later today, could you please fix the small build error.

Ta

attrs.go
```diff
@@ -106,6 +106,57 @@ func unmarshalAttrs(b []byte) (*FileStat, []byte) {
...
+
+	uid := uint32(0)
+	gid := uint32(0)
+	mtime := uint32(fi.ModTime().Unix())
+	atime := mtime
+
+	flags := ssh_FILEXFER_ATTR_SIZE |
+		ssh_FILEXFER_ATTR_PERMISSIONS |
+		ssh_FILEXFER_ATTR_ACMODTIME |
+		uint32(0)
```
davecheney: var flags uint32 = ...

so you don't need the uint32(0) at the end

client.go
```diff
@@ -16,6 +16,7 @@ import (
 	"github.com/kr/fs"
 
 	"golang.org/x/crypto/ssh"
+	//"github.com/ScriptRock/crypto/ssh"
```
davecheney: delete this please

client_integration_test.go
```diff
@@ -29,6 +32,9 @@ const (
 	debuglevel = "ERROR" // set to "DEBUG" for debugging
 )
 
+var spaceRegex = regexp.MustCompile(`\s+`)
```
davecheney: please move this closer to where it is used, it only looks like it is used in one test

example_test.go
```diff
@@ -6,9 +6,10 @@ import (
 	"os"
 	"os/exec"
 
-	"golang.org/x/crypto/ssh"
-
+	//"github.com/ScriptRock/crypto/ssh"
```
davecheney: please clean these up

packet.go
```diff
@@ -4,9 +4,18 @@ import (
 	"encoding"
 	"fmt"
 	"io"
+	"os"
 	"reflect"
 )
 
+var (
+	shortPacketError       = fmt.Errorf("packet too short")
+	debugDumpTxPacket      = false
```
davecheney: can these be made constants ?

packet.go
```diff
@@ -88,11 +141,36 @@ func recvPacket(r io.Reader) (uint8, []byte, error) {
...
 	}
+	if debugDumpRxPacketBytes {
+		debug("recv packet: %s %d bytes %x", fxp(b[0]).String(), l, b[1:])
+	} else if debugDumpRxPacket {
+		debug("recv packet: %s %d bytes", fxp(b[0]).String(), l)
+	}
 	return b[0], b[1:], nil
 }
 
+type ExtensionPair struct {
```
davecheney: Does this need to be exported ?

packet.go
```diff
@@ -132,27 +246,47 @@ func marshalIdString(packetType byte, id uint32, str string) ([]byte, error) {
 	return b, nil
 }
 
+func unmarshalIdString(b []byte, id *uint32, str *string) (err error) {
```
davecheney: can you return this string and the id, rather than updating the reference ?

please, no named return values unless absolutely necessary

server.go
```diff
@@ -0,0 +1,448 @@
...
+	"encoding"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"sync"
+	"syscall"
+)
+
+type Server struct {
```
davecheney: Please comment this public type

// Server is an ssh Server.
// etc.

server.go
```diff
@@ -0,0 +1,448 @@
...
+	debugLevel    int
+	readOnly      bool
+	rootDir       string
+	lastId        uint32
+	pktChan       chan rxPacket
+	openFiles     map[string]*os.File
+	openFilesLock *sync.RWMutex
+	handleCount   int
+	maxTxPacket   uint32
+	WorkerCount   int
```
davecheney: does this need to be public ?

server.go
```diff
@@ -0,0 +1,448 @@
...
+// A subsequent call to Run() is required.
+func NewServer(in io.Reader, out io.WriteCloser, debugStream io.Writer, debugLevel int, readOnly bool, rootDir string) (*Server, error) {
+	if rootDir == "" {
+		if wd, err := os.Getwd(); err != nil {
+			return nil, err
+		} else {
+			rootDir = wd
+		}
+	}
+	workerCount := 8
```
davecheney: make this a constant or assign it directly in the literal.

server.go
```diff
@@ -0,0 +1,448 @@
...
+			pkt.respond(svr)
+		}
+	}
+	doneChan <- nil
+}
+
+// Run this server until the streams stop or until the subsystem is stopped
+func (svr *Server) Run() error {
+	if svr.WorkerCount <= 0 {
+		return fmt.Errorf("sftp server requires > 0 workers")
```
davecheney: can this ever happen ?

server.go
```diff
@@ -0,0 +1,448 @@
...
+		} else {
+			//fmt.Fprintf(svr.debugStream, "pkt: %T %v\n", pkt, pkt)
+			pkt.respond(svr)
+		}
+	}
+	doneChan <- nil
+}
+
+// Run this server until the streams stop or until the subsystem is stopped
+func (svr *Server) Run() error {
```
davecheney: maybe call this Serve, like http.Serve

server.go
```diff
@@ -0,0 +1,448 @@
...
+	case ssh_FXP_RMDIR:
+	case ssh_FXP_REALPATH:
+		pkt = &sshFxpRealpathPacket{}
+	case ssh_FXP_STAT:
+		pkt = &sshFxpStatPacket{}
+	case ssh_FXP_RENAME:
+		pkt = &sshFxpRenamePacket{}
+	case ssh_FXP_READLINK:
+		pkt = &sshFxpReadlinkPacket{}
+	case ssh_FXP_SYMLINK:
```
davecheney: if these are cases that should be ignored, then do

case ssh_FXP_STATUS, ssh_FXP_HANDLE, ...:
   // continue

server.go
```diff
@@ -0,0 +1,448 @@
...
+	case ssh_FXP_FSETSTAT:
+	case ssh_FXP_OPENDIR:
+		pkt = &sshFxpOpendirPacket{}
+	case ssh_FXP_READDIR:
+		pkt = &sshFxpReaddirPacket{}
+	case ssh_FXP_REMOVE:
+		pkt = &sshFxpRemovePacket{}
+	case ssh_FXP_MKDIR:
+		pkt = &sshFxpMkdirPacket{}
+	case ssh_FXP_RMDIR:
```
davecheney: move this down

server.go
```diff
@@ -0,0 +1,448 @@
...
+		pkt = &sshFxpReadlinkPacket{}
+	case ssh_FXP_SYMLINK:
+	case ssh_FXP_STATUS:
+	case ssh_FXP_HANDLE:
+	case ssh_FXP_DATA:
+	case ssh_FXP_NAME:
+	case ssh_FXP_ATTRS:
+	case ssh_FXP_EXTENDED:
+	case ssh_FXP_EXTENDED_REPLY:
+	default:
```
davecheney: if these are ignored, then remove them and the default case will handle them

davecheney:
Very nice, thanks again. There are a few small nits to take care of but this is pretty close.

marksheahan:
Thanks, I'll address all the comments soon.

client.go
```diff
@@ -312,6 +336,25 @@ func (c *Client) ReadLink(p string) (string, error) {
 	}
 }
 
+// Symlink creates a symbolic link at 'newname', pointing at target 'oldname'
```
davecheney: This should be called Link, for compatibility with the os package, http://godoc.org/os#Link

You can do this in a followup if you like.

client_integration_darwin_test.go
```diff
@@ -0,0 +1,41 @@
+package sftp
```
davecheney: why does this test only run on darwin ?

can ou rename it to client_integration_test.go and use 

// +build !windows

to rule out that os ?

client_integration_test.go
```diff
@@ -80,12 +84,41 @@ func (w delayedWriter) Close() error {
...
+	rxPipeRd, rxPipeWr := io.Pipe()
+
+	server, err := NewServer(txPipeRd, rxPipeWr, os.Stderr, 0, readonly, ".")
+	if err != nil {
+		t.Fatal(err)
+	}
+	go server.Serve()
+
+	var ctx io.WriteCloser = txPipeWr
+	if delay > NO_DELAY {
```
davecheney: !delay.IsZero()

packet.go
```diff
@@ -52,26 +69,58 @@ func unmarshalUint32(b []byte) (uint32, []byte) {
 	return v, b[4:]
 }
 
+func unmarshalUint32Safe(b []byte) (uint32, []byte, error) {
```
davecheney: please make this wrap unmarshalUint32, do the same for the other *safe derivatives.

packet.go
```diff
@@ -88,11 +144,36 @@ func recvPacket(r io.Reader) (uint8, []byte, error) {
 	l, _ := unmarshalUint32(b)
 	b = make([]byte, l)
 	if _, err := io.ReadFull(r, b); err != nil {
+		debug("recv packet %d bytes: err %v", l, err)
 		return 0, nil, err
 	}
+	if debugDumpRxPacketBytes {
+		debug("recv packet: %s %d bytes %x", fxp(b[0]).String(), l, b[1:])
```
davecheney: you don't need the .String(), if fxp(b[0]) implements Stringer then it'll be called anyway

server.go
```diff
@@ -0,0 +1,565 @@
...
+	openFiles     map[string]serverOpenFile
+	openFilesLock *sync.RWMutex
+	handleCount   int
+	maxTxPacket   uint32
+	workerCount   int
+}
+
+type serverOpenFile struct {
+	*os.File
+	path string
```
davecheney: why is this needed, is os.File.Name() incorrect ?

server.go
```diff
@@ -0,0 +1,565 @@
...
+}
+
+func (svr *Server) nextHandle(path string, f *os.File) string {
+	svr.openFilesLock.Lock()
+	defer svr.openFilesLock.Unlock()
+	svr.handleCount++
+	handle := fmt.Sprintf("%d", svr.handleCount)
```
davecheney: Why stringify this, can you make svr.OpenFiles a map of int to *serverOpenFile ?

user/lookup.go
```diff
@@ -0,0 +1,49 @@
+// Copyright 2011 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package user
```
davecheney: Why do we need a new package for this ? The name is not good, pkg/sftp/user doesn't explain what this package provides. 

Can it be merged into the sftp package please.

user/lookup_unix.go
```diff
@@ -0,0 +1,286 @@
+// Copyright 2011 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// +build darwin dragonfly freebsd !android,linux netbsd openbsd solaris
+// +build cgo
+
+package user
```
davecheney: Why can't yuou use the functions in the os and os/user package for this ?

Is it possible to back this package out and propose it as a followup please.

davecheney:
Thanks. A bunch more comments and nits.

To get this landed as quickly as possible I want to encourage you to pull out anything that is controversial and handle it as a followup.

user/lookup_unix.go
marksheahan: os/user doesn't provide anything for looking up groups, and I couldn't find anything in the standard library that does group lookup, or even any actively maintained project either.

This whole section is there for username groupname lookup in the 'ls -l' style output returned by ReadDir. It's easy enough to remove this package, but when connecting to this sftp server via the 'sftp' program, output of ls -l will look strange, showing uid/gid instead of ownername groupname. Everything else should still work.

I don't believe group lookup should be part of the sftp package.

For the moment I'll remove this package entirely. Perhaps a group package, which has only the 'group' functions, and using os/user for user lookup is the best way to proceed.

client.go
marksheahan: os.Link() creates a hard link, not a soft link. This name was chosen for consistency with os.Symlink:
https://golang.org/pkg/os/#Symlink

davecheney: My apologies, this was my mistake

client_integration_darwin_test.go
marksheahan: I found that the field names available in syscall.Statfs_t were adequately different between kernels (darwin vs linux) that I could not re-use client_integration_linux_test.go when testing on darwin.

server.go
marksheahan: The handles generated here get returned as string type handles to the sftp client. Client requests on an already opened file provide a string typed handle in the request packets. With a map[string]*os.File, is makes slightly simpler requesting code in each of sftp.*File.Read(), sftp.*File.Write() etc; just lookup using the handle in the request. I can still change the code around, there's just less type conversion elsewhere if I do it here instead.

marksheahan:
Working on deleting the user/ dir, and the other items. Thanks for the quick review turnaround!

.gitignore
```diff
@@ -0,0 +1,6 @@
+.*.swo
+.*.swp
+
+testdata/
```
davecheney: the go tool ignores testdata, please remove this line

if something is writing into testdata/ let's fix that in a followup

client.go
```diff
@@ -259,6 +259,30 @@ func (c *Client) opendir(path string) (string, error) {
 	}
 }
 
+func (c *Client) Stat(p string) (os.FileInfo, error) {
```
davecheney: needs a public comment

davecheney:
LGTM. I've sent you an invitation to collaborate on the sftp repository, when you're happy, please submit this. Let's tackle the other changes that I asked to be deferred in follow up PRs.

Thanks again for this big improvement, this was a large hole in the package.

marksheahan:
Just got nervous before committing, wrote a few more tests and found a bug. I'll add a recursive put and get test; once those pass I'll be happy with it.

davecheney:
Cool. We can always land more pull requests after this one, you don't need
to eat the entire elephant in one sitting.

.gitignore
```diff
@@ -0,0 +1,5 @@
+.*.swo
+.*.swp
+
+server_standalone/server_standalone
```
davecheney: you can probably delete this as well, this would only happen if you did go build in the server_standalone directory

corydoras: Not that .gitignore files need to be filled with thousands of lines of junk. Assumedly including 'server_standalone/server_standalone' is useful during development?

marksheahan: Yes, I did my integration tests that way. I probably could have used go install but it's just habit...

davecheney: It's your call, doesn't really matter much each way, i'm just finding
things to nit pick while you land your change.

PR #48, opened 2015-09-20
fix broken windows build

### LINKED ISSUE #47: Latest commit broke the windows build
build on mac works nicely, but cross compile for windows and it blows up.   I also tried it on a windows box, and received similar error.  Roll back the latest PR, and all is well.

```
sftp $ go install
sftp $ env GOOS="windows" GOARCH="386" go install
# github.com/pkg/sftp
./attrs.go:134: undefined: syscall.Stat_t
./server_stubs.go:5: undefined: os in os.FileInfo
./server_stubs.go:6: undefined: path in path.Join
```

marksheahan:
Fixes #47

### DIFF
--- a/attrs.go
+++ b/attrs.go
@@ -71,6 +71,26 @@ func fileInfoFromStat(st *FileStat, name string) os.FileInfo {
 	return fs
 }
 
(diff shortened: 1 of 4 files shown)


jtwatson:
I tested with my project and all works now.  Thanks for the quick fix.

Issue #52, opened 2015-10-26
Lenience with non-compliant servers?


boomlinde:
I recently used this library for a client sftp setup. Their server software (unknown to us at the moment) disregards the relevant IETF document by not including the language tag in all its status messages. For this use case, I patched `unmarshalStatus` to use `unmarshalStringSafe` when reading the language tag from the status message which works fine.

The question now is whether this project is aiming for strict compliance, or if the patch I made is relevant to the project. `sftp` from OpenSSH seems to ignore that the language tag is missing, and I found that WinSCP had patched for similar issues with broken server software (http://winscp.net/tracker/show_bug.cgi?id=770).

davecheney:
> The question now is whether this project is aiming for strict compliance, or if the patch I made is relevant to the project.

Given that there is no spec (no really, there isn't) for sftp, and most impls just try to do what scp(1) does, I'd be happy to review any PR's you have.

PR #53, opened 2015-10-27
make the StatusError language tag optional

boomlinde:
Be more lenient with servers that don't implement the language tag part of status messages. Addresses issue #52


davecheney:
Thank you. I have no issue with this change, but this will need a test so it does not regress in the future.

boomlinde:
Thanks. I added a test to client_tests.go that would fail gracefully before the change in e6dc896 and passes after it.

davecheney:
Perfect, thanks.

PR #54, opened 2015-11-07
Created working example SFTP server

kothar:
Purely gluing together the standalone_server and the SSH server examples, since
it took me a while to work out how they fit together. Thought it may save the next
person a few minutes!

### DIFF
--- a/examples/sftp-server/main.go
+++ b/examples/sftp-server/main.go
@@ -1,0 +1,132 @@
+// An example SFTP server implementation using the golang SSH package.
+// Serves the whole filesystem visible to the user, and has a hard-coded username and password,
+// so not for real use!
+package main
+
+import (
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"net"
+	"os"
+
+	"github.com/pkg/sftp"
+	"golang.org/x/crypto/ssh"
+)
+
+// Based on example server code from golang.org/x/crypto/ssh and server_standalone
+func main() {
+
+	readOnly := false
+	debugLevelStr := "none"
+	debugLevel := 0
+	debugStderr := false
+	rootDir := ""
+
+	flag.BoolVar(&readOnly, "R", false, "read-only server")
+	flag.BoolVar(&debugStderr, "e", false, "debug to stderr")
+	flag.StringVar(&debugLevelStr, "l", "none", "debug level")
+	flag.StringVar(&debugLevelStr, "root", "", "root directory")
+	flag.Parse()
+
+	// term := terminal.NewTerminal(channel, "> ")
+	debugStream := ioutil.Discard
+	if debugStderr {
+		debugStream = os.Stderr
+		debugLevel = 1
+	}
+
+	// An SSH server is represented by a ServerConfig, which holds
+	// certificate details and handles authentication of ServerConns.
+	config := &ssh.ServerConfig{
+		PasswordCallback: func(c ssh.ConnMetadata, pass []byte) (*ssh.Permissions, error) {
+			// Should use constant-time compare (or better, salt+hash) in
+			// a production setting.
+			fmt.Fprintf(debugStream, "Login: %s\n", c.User())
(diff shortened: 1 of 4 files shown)


examples/sftp-server/main.go
```diff
@@ -0,0 +1,132 @@
...
+	"os"
+
+	"github.com/pkg/sftp"
+	"golang.org/x/crypto/ssh"
+)
+
+// Based on example server code from golang.org/x/crypto/ssh and server_standalone
+func main() {
+
+	readOnly := false
```
davecheney: try
 var (
   readOnly bool
   debufLevelStr string
   ...
)

These will then have their zero value, which will then be overwritten by flag.Parse

examples/sftp-server/main.go
```diff
@@ -0,0 +1,132 @@
...
+	debugStderr := false
+	rootDir := ""
+
+	flag.BoolVar(&readOnly, "R", false, "read-only server")
+	flag.BoolVar(&debugStderr, "e", false, "debug to stderr")
+	flag.StringVar(&debugLevelStr, "l", "none", "debug level")
+	flag.StringVar(&debugLevelStr, "root", "", "root directory")
+	flag.Parse()
+
+	// term := terminal.NewTerminal(channel, "> ")
```
davecheney: delete

examples/sftp-server/main.go
```diff
@@ -0,0 +1,132 @@
...
+			if c.User() == "testuser" && string(pass) == "tiger" {
+				return nil, nil
+			}
+			return nil, fmt.Errorf("password rejected for %q", c.User())
+		},
+	}
+
+	privateBytes, err := ioutil.ReadFile("id_rsa")
+	if err != nil {
+		panic("Failed to load private key")
```
davecheney: log.Fatal is better than panic in examples

examples/sftp-server/main.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	private, err := ssh.ParsePrivateKey(privateBytes)
+	if err != nil {
+		panic("Failed to parse private key")
```
davecheney: same

examples/sftp-server/main.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	config.AddHostKey(private)
+
+	// Once a ServerConfig has been configured, connections can be
+	// accepted.
+	listener, err := net.Listen("tcp", "0.0.0.0:2022")
+	if err != nil {
+		panic("failed to listen for connection")
```
davecheney: same

examples/sftp-server/main.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+	fmt.Printf("Listening on %v\n", listener.Addr())
+
+	nConn, err := listener.Accept()
+	if err != nil {
+		panic("failed to accept incoming connection")
```
davecheney: same,

davecheney:
Looks pretty good to me, thank you. If you can clean up the minor comments I'll merge this for you.

examples/sftp-server/main.go
kothar: This code is based on https://godoc.org/golang.org/x/crypto/ssh#example-NewServerConn - is it worth making these suggestions for inclusion there too?

I agree that an example should demonstrate best-practice, especially if it's likely people will use it as the base for future work.

davecheney: Yes please! Old examples age like fruit in the sun. If you could please fix
up those as well that would be great.

mdlayher:
Thanks for this example, it was very helpful!  I look forward to seeing it merged.

davecheney:
@mdlayher if you want to take over this PR that would be great.

mdlayher:
@davecheney Sure, would you just like the other examples updated as well?  I see your previous comments have been resolved.

davecheney:
Oh, I didn't realise the OP had updated it. I'll merge this now.

PR #58, opened 2015-12-21
Add Getwd to client

sykesm:
`Realpath` is useful when working with relative paths and it helps users of the client keep track of the current working directory.

### DIFF
--- a/client.go
+++ b/client.go
@@ -235,6 +235,7 @@ func (c *Client) ReadDir(p string) ([]os.FileInfo, error) {
 	}
 	return attrs, err
 }
+
 func (c *Client) opendir(path string) (string, error) {
 	id := c.nextId()
 	typ, data, err := c.sendRequest(sshFxpOpendirPacket{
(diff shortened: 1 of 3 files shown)


davecheney:
Thank you for this PR. Can you please raise a bug describing the problem you have, and link it to this PR as a solution. 

Thanks

Dave

sykesm:
@davecheney sure. Issue raised.

If you prefer to have issues raised before pull requests, you might want to add that information to the `contributing` section of the README.

Thanks.

davecheney:
Thanks. I'll make sure I update the contributing info in the readme. I've
been a bit slack about doing that for my projects. Sorry about that.

davecheney:
Thanks for raising #59. Can you please change this PR to add something like client.Getwd() (string, error) to match the `os` package function of the same name.

sykesm:
Please let me know if this is closer to what you were looking for. Thanks.

davecheney:
LGTM thanks.

PR #67, opened 2015-12-30
server: simplify code and improve readability

mdlayher:
- Eliminates unnecessary `else` blocks
- Breaks up long lines and struct literals
- Uses keyed struct literals
- Removes assignment from zero value variables
- Replaces repeated `if/else` with `switch`


davecheney:
LGTM, Thanks.

Issue #65, opened 2015-12-28
Root directory passed to NewServer is never used or acted on


sykesm:
The `NewServer` constructor accepts and propagates a `rootDir` argument to the `Server` but the location is never used after construction. What was the intent of the argument?  Could/should it be treated as the default working directory for the server instance?

sykesm:
Looks like #56 was related. I'd like to see the behavior [described] by @marksheahan implemented as we have linked the sftp server into an ssh daemon and would like to be able to set the default working directory of each instance to the home directory of the user. If this behavior is not implemented in the server, we will have to create a separate sftp server executable and exec it with the appropriate working directory.

If you're interested in a PR, I'd be happy to change the name of the argument to the constructor and make the necessary implementation changes.

Thanks.

[described]: https://github.com/pkg/sftp/pull/56#issuecomment-164325286

sykesm:
Given the confusion over the meaning of the `rootDir` on the constructor and the fact that it's not implemented, perhaps it makes sense to remove it completely?

davecheney:
Yes, please remove it, then we can start from a fresh slate. Thanks

mdlayher:
Sounds good to me as well.  I somehow didn't catch that it was totally unused when I was making my PRs.  Apologies for that.

sykesm:
Addressed with 8e47c759f51e7738c7cd4cf505f1ded28f79ddce.

PR #68, opened 2016-01-01
Implement server default working directory

sykesm:
Possible resolution to #65.

### DIFF
--- a/server.go
+++ b/server.go
@@ -29,7 +29,7 @@ type Server struct {
 	outMutex      *sync.Mutex
 	debugStream   io.Writer
 	readOnly      bool
-	rootDir       string
+	workingDir    string
 	lastID        uint32
 	pktChan       chan rxPacket
 	openFiles     map[string]*os.File
@@ -66,6 +66,13 @@ func (svr *Server) getHandle(handle string) (*os.File, bool) {
 	return f, ok
 }
 
+func (svr *Server) resolvePath(path string) string {
+	if filepath.IsAbs(path) {
+		return path
+	}
+	return filepath.Join(svr.workingDir, path)
+}
+
 type serverRespondablePacket interface {
 	encoding.BinaryUnmarshaler
 	id() uint32
@@ -78,14 +85,10 @@ type serverRespondablePacket interface {
 // functions may be specified to further configure the Server.
 //
 // A subsequent call to Serve() is required to begin serving files over SFTP.
-func NewServer(in io.Reader, out io.WriteCloser, rootDir string, options ...ServerOption) (*Server, error) {
-	if rootDir == "" {
-		wd, err := os.Getwd()
-		if err != nil {
-			return nil, err
-		}
-
-		rootDir = wd
+func NewServer(in io.Reader, out io.WriteCloser, workingDir string, options ...ServerOption) (*Server, error) {
+	workingDir, err := filepath.Abs(workingDir)
+	if err != nil {
+		return nil, err
 	}
 
 	s := &Server{
@@ -93,7 +96,7 @@ func NewServer(in io.Reader, out io.WriteCloser, rootDir string, options ...Serv
 		out:           out,
 		outMutex:      &sync.Mutex{},
 		debugStream:   ioutil.Discard,
-		rootDir:       rootDir,
+		workingDir:    workingDir,
 		pktChan:       make(chan rxPacket, sftpServerWorkerCount),
 		openFiles:     map[string]*os.File{},
 		openFilesLock: &sync.RWMutex{},
@@ -268,7 +271,7 @@ func (p sshFxpLstatPacket) readonly() bool { return true }
 
 func (p sshFxpLstatPacket) respond(svr *Server) error {
 	// stat the requested file
-	info, err := os.Lstat(p.Path)
+	info, err := os.Lstat(svr.resolvePath(p.Path))
 	if err != nil {
 		return svr.sendPacket(statusFromError(p.ID, err))
 	}
@@ -283,7 +286,7 @@ func (p sshFxpStatPacket) readonly() bool { return true }
 
 func (p sshFxpStatPacket) respond(svr *Server) error {
 	// stat the requested file
-	info, err := os.Stat(p.Path)
+	info, err := os.Stat(svr.resolvePath(p.Path))
 	if err != nil {
 		return svr.sendPacket(statusFromError(p.ID, err))
 	}
@@ -317,35 +320,35 @@ func (p sshFxpMkdirPacket) readonly() bool { return false }
 
 func (p sshFxpMkdirPacket) respond(svr *Server) error {
 	// TODO FIXME: ignore flags field
-	err := os.Mkdir(p.Path, 0755)
+	err := os.Mkdir(svr.resolvePath(p.Path), 0755)
 	return svr.sendPacket(statusFromError(p.ID, err))
 }
 
 func (p sshFxpRmdirPacket) readonly() bool { return false }
 
 func (p sshFxpRmdirPacket) respond(svr *Server) error {
-	err := os.Remove(p.Path)
+	err := os.Remove(svr.resolvePath(p.Path))
 	return svr.sendPacket(statusFromError(p.ID, err))
 }
 
 func (p sshFxpRemovePacket) readonly() bool { return false }
 
 func (p sshFxpRemovePacket) respond(svr *Server) error {
-	err := os.Remove(p.Filename)
+	err := os.Remove(svr.resolvePath(p.Filename))
 	return svr.sendPacket(statusFromError(p.ID, err))
 }
 
 func (p sshFxpRenamePacket) readonly() bool { return false }
 
 func (p sshFxpRenamePacket) respond(svr *Server) error {
-	err := os.Rename(p.Oldpath, p.Newpath)
+	err := os.Rename(svr.resolvePath(p.Oldpath), svr.resolvePath(p.Newpath))
 	return svr.sendPacket(statusFromError(p.ID, err))
 }
 
 func (p sshFxpSymlinkPacket) readonly() bool { return false }
 
 func (p sshFxpSymlinkPacket) respond(svr *Server) error {
-	err := os.Symlink(p.Targetpath, p.Linkpath)
+	err := os.Symlink(svr.resolvePath(p.Targetpath), svr.resolvePath(p.Linkpath))
 	return svr.sendPacket(statusFromError(p.ID, err))
 }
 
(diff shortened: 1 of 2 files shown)


server_integration_test.go
```diff
@@ -567,6 +604,41 @@ func TestServerGet(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.RemoveAll(tmpWorkingDir)
+
+	listenerGo, hostGo, portGo := testServer(t, GOLANG_SFTP, READONLY, tmpWorkingDir)
+	defer listenerGo.Close()
+
+	fn := randName()
+	tmpFileLocal := filepath.Join("/tmp", fn)
```
davecheney: $TMPDIR is not /tmp on osx ;(

server_integration_test.go
```diff
@@ -567,6 +604,41 @@ func TestServerGet(t *testing.T) {
...
+	defer os.RemoveAll(tmpFileLocal)
```
davecheney: Why not use ioutil.TempFile ?

server_integration_test.go
```diff
@@ -567,6 +604,41 @@ func TestServerGet(t *testing.T) {
...
+
+	t.Logf("get: local %v remote %v ", tmpFileLocal, fn)
+
+	// create a file with random contents. This will be the remote file pulled from the server
+	tmpFileRemoteData := randData(10 * 1024 * 1024)
```
davecheney: I don't see what this is testing ? Why do we need to write 10mb of zeros, sftp it from the sever to the same tmpdir locally. Isn't this about testing path names ?

davecheney:
Thanks for this PR. I have several concerns, the main one being that this PR appears to do the opposite of what I believe #56 #65 intended -- it lets the caller escape the rootDir/workingDir.

Here is what I think should be done to make #56 work.

1. retain rootDir, it's a fine name, if you want to rename it to workingDir, please do that in a followup, the name is not key.
2. create a function on sever, something like openFile(path, flags). openFile rejects paths that are not absolute. 
3. If the path is absolute then use filepath.join(srv.rootDir, path), and open that.

sykesm:
Thanks for the review.

As I [said](https://github.com/pkg/sftp/issues/65#issuecomment-167669611) in issue #65, I have no need for a `chroot` but I do need the behavior originally [described](https://github.com/pkg/sftp/pull/56#issuecomment-164325286) by @marksheahan in #56. These are two orthogonal use cases.

My goal is to create a single statically linked ssh daemon that includes an sftp server. When users start an sftp session, they should logically be placed into their own $HOME directory. The implemented behavior is consistent with sftp-server's concept of a [start directory](https://github.com/openssh/openssh-portable/blob/master/sftp-server.c#L1495).

If we're in agreement on this, I'd be happy to proceed with addressing concerns in the tests.

server_integration_test.go
sykesm: You're absolutely correct. This code was essentially copied from the 8 existing occurrences in the test. Would you like a PR, separate commit, or something else to correct this issue?

davecheney:
> My goal is to create a single statically linked ssh daemon that includes an sftp server. When users start an sftp session, they should logically be placed into their own $HOME directory. The implemented behavior is consistent with sftp-server's concept of a start directory.

My reading of start directory is it does a chdir when the client connects. However it sounds like you want `chroot`.

davecheney:
Hmm. Maybe I don't understand. Do you want a way to control the `pwd` of the sftp server ? Or do you want `chroot` to sandbox users ?

server_integration_test.go
sykesm: The test is about the behavior of a relative get where the base `fn` is used as the [source](https://github.com/pkg/sftp/pull/68/files#diff-3e4bb7147866350f6ffeea6564dfe0e9R630) of the transfer.

The generation of silly goo is, again, simply propagating the patterns that exist in the current tests.

sykesm:
> My reading of start directory is it does a chdir when the client connects.

Yes, for the sftp-server, it changes its working directory on start. This works there because it's running a separate process per user. I'd like to avoid that additional process overhead and the need for a separate binary. Ideally I can simply embed the sftp server in my daemon and let the sftp server apply a logical working directory.

> Do you want a way to control the pwd of the sftp server ? Or do you want chroot to sandbox users ?

I want to control the logical pwd of the sftp server. In particular, if clients use relative path names, they are treated as relative to the logical pwd and if a client sends a `realpath(".")`, they get an absolute path to the logical pwd.

davecheney:
Ahh, ok, I think I understand what you mean now. Thanks to baring with me.

I suggest that you start by teaching the server a per connection working
directory. I think the PR did some of this, but it didn't appear to track
this directory per connection.

sykesm:
> I suggest that you start by teaching the server a per connection working directory.

Can you elaborate on that? At first glance it would seem to change the current model quite a bit as a server instance is currently tied to a single ssh session channel.

sykesm:
@davecheney can you elaborate on what you mean by "teaching the server a per connection working directory?" The server implementation is already tied to an ssh session (which is a logical connection)?

The common use case is to create one sftp server instance per subsystem request.

Thanks.

marksheahan:
Looks good to me. The way I'm using the server is to call Serve() in the ssh server sftp subsystem handler. This could allow sftp servers with different directories per session, not per connection, but I can't imagine any real use case for that.

I occasionally worry about the number of worker pool goroutines spawned (8 per call to Serve()). Goroutines are cheap, but I'm not totally comfortable with it for some reason.

mdlayher:
Seems fine here as well.  Can you rebase on master and push again?

sykesm:
@mdlayher, @marksheahan I've rebased on top of master, replaced references to `/tmp/` with `os.TempDir()`, and used `ioutil.TempDir()` where I think it makes sense.

I think the only outstanding comments were related to the name of `rootDir` vs `workingDir` and whether or not the relative {get,put} tests make sense.

With respect to `rootDir` vs. `workingDir`. I don't know if there's a consensus on one vs. the other but I believe `workingDir` more accurately reflects the purpose. If this is something you want in a separate commit, I'm happy to do that.

For the relative path name tests, what's in the commit is probably overkill but it certainly validates the behavior. If you'd like any changes, please let me know and I'll rework as appropriate.

Thanks.

mdlayher:
Ping @davecheney since he had previous concerns.  This still looks fine to me.

server.go
```diff
@@ -66,6 +66,13 @@ func (svr *Server) getHandle(handle string) (*os.File, bool) {
 	return f, ok
 }
 
+func (svr *Server) resolvePath(path string) string {
+	if filepath.IsAbs(path) {
```
davecheney: I don't think this is right, I think 

return filepath.Rel(srv.workingDir, path) is all that is required.

server.go
```diff
@@ -78,14 +85,10 @@ type serverRespondablePacket interface {
...
-func NewServer(in io.Reader, out io.WriteCloser, rootDir string, options ...ServerOption) (*Server, error) {
-	if rootDir == "" {
-		wd, err := os.Getwd()
-		if err != nil {
-			return nil, err
-		}
-
-		rootDir = wd
+func NewServer(in io.Reader, out io.WriteCloser, workingDir string, options ...ServerOption) (*Server, error) {
+	workingDir, err := filepath.Abs(workingDir)
```
davecheney: abs is probably not necessary, if someone puts something dumb there, they get a dumb result.

server_integration_test.go
```diff
@@ -505,11 +506,11 @@ func TestServerSymlink(t *testing.T) {
 }
 
 func TestServerPut(t *testing.T) {
-	listenerGo, hostGo, portGo := testServer(t, GOLANG_SFTP, READONLY)
+	listenerGo, hostGo, portGo := testServer(t, GOLANG_SFTP, READONLY, ".")
 	defer listenerGo.Close()
 
-	tmpFileLocal := "/tmp/" + randName()
-	tmpFileRemote := "/tmp/" + randName()
+	tmpFileLocal := filepath.Join(os.TempDir(), randName())
```
davecheney: Can this please use ioutil.TempFile please

server_integration_test.go
```diff
@@ -534,12 +535,47 @@ func TestServerPut(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.RemoveAll(tmpWorkingDir)
+
+	listenerGo, hostGo, portGo := testServer(t, GOLANG_SFTP, READONLY, tmpWorkingDir)
+	defer listenerGo.Close()
+
+	fn := randName()
+	tmpFileLocal := filepath.Join(os.TempDir(), fn)
```
davecheney: same as above

server_integration_test.go
```diff
@@ -534,12 +535,47 @@ func TestServerPut(t *testing.T) {
...
+}
+
 func TestServerGet(t *testing.T) {
-	listenerGo, hostGo, portGo := testServer(t, GOLANG_SFTP, READONLY)
+	listenerGo, hostGo, portGo := testServer(t, GOLANG_SFTP, READONLY, ".")
 	defer listenerGo.Close()
 
-	tmpFileLocal := "/tmp/" + randName()
-	tmpFileRemote := "/tmp/" + randName()
+	tmpFileLocal := filepath.Join(os.TempDir(), randName())
```
davecheney: same as above

davecheney:
I'm sorry i've been so slow on reviewing this PR. This is for two reasons.

1. I don't understand the underlying motivation. I'm tempted to remove `rootDir` and start again from scratch.
2. I'm very very worried about this change introducing path traversal bugs, this is something that many open source projects have gotten wrong over the years, and I don't think I understand it well enough to not emulate their mistakes. I don't fully understand the difference between how this feature of changing the default working directory on connect interacts with the client, and issues like #56 which expect the client to be sandboxed to that directory or lower.

In short, I'm terrified of this change and I'd rather not make it. Is there any way we can solve #56 without introduction the requirement to compute every path relative

sykesm:
Hi @davecheney. Given you're not on board with the motivation, I don't think it matters what the code looks like - there's a disconnect. That's part of the reason why I didn't want to make any changes until I heard back on my [question](https://github.com/pkg/sftp/pull/68#issuecomment-168391167).

I'm going to close this PR and go back to the original issue where I hope we can come up with a possible means to get the behavior I desire that is not in conflict with where you want things to go.

davecheney:
@sykesm I'm sorry I cannot elaborate on my answer any more than I have already done. I don't understand the motivation for this change, so I was probably wrong in my response to you earlier. I am sorry for any confusion caused.

PR #72, opened 2016-01-03
client: make Client.{Lstat,Open,Stat} satisfy os.IsNotExist

### LINKED ISSUE #71: (*Client).Stat and (*Client).Open don't return os.ErrNotExist for nonexistent paths
Running the following test:

    go test --integration

    func TestNonexistentPath(t *testing.T) {
            sftp, cmd := testClient(t, READONLY, NO_DELAY)
            defer cmd.Wait()
            defer sftp.Close()
    
            if _, err := sftp.Stat("/doesnt/exist/"); !os.IsNotExist(err) {
                    t.Errorf("os.IsNotExist(%v) = false, want true", err)
            }
    
            if _, err := sftp.Open("/doesnt/exist/"); !os.IsNotExist(err) {
                    t.Errorf("os.IsNotExist(%v) = false, want true", err)
            }
    }

Results in 

    --- FAIL: TestNonexistentPath (0.01s)
            client_integration_test.go:171: os.IsNotExist(sftp: "No such file" (SSH_FX_NO_SUCH_FILE)) = false, want true
            client_integration_test.go:175: os.IsNotExist(sftp: "No such file" (SSH_FX_NO_SUCH_FILE)) = false, want true
    process_write: write failed
    FAIL
    exit status 1
    FAIL        github.com/pkg/sftp     2.149s

mdlayher:
Fixes #71.

### DIFF
--- a/client.go
+++ b/client.go
@@ -224,7 +224,7 @@ func (c *Client) ReadDir(p string) ([]os.FileInfo, error) {
 			}
 		case ssh_FXP_STATUS:
 			// TODO(dfc) scope warning!
-			err = eofOrErr(unmarshalStatus(id, data))
+			err = normaliseError(unmarshalStatus(id, data))
 			done = true
 		default:
 			return nil, unimplementedPacketErr(typ)
@@ -280,7 +280,7 @@ func (c *Client) Stat(p string) (os.FileInfo, error) {
 		attr, _ := unmarshalAttrs(data)
 		return fileInfoFromStat(attr, path.Base(p)), nil
 	case ssh_FXP_STATUS:
-		return nil, unmarshalStatus(id, data)
+		return nil, normaliseError(unmarshalStatus(id, data))
 	default:
 		return nil, unimplementedPacketErr(typ)
 	}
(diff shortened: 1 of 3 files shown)


client.go
```diff
@@ -1135,6 +1135,13 @@ func min(a, b int) int {
 	return a
 }
 
+func notExistOrErr(err error) error {
```
davecheney: what about somethign like 'normaliseError`

client.go
```diff
@@ -1135,6 +1135,13 @@ func min(a, b int) int {
...
+	if err, ok := err.(*StatusError); ok && err.Code == ssh_FX_NO_SUCH_FILE {
```
davecheney: switch err := err.(type) {
case *StatusError {
     switch err.Code {
           case ssh_FX_...:
                   return os.ErrNotExist
           default:
                   return err
           }
default:
          return err
}

davecheney:
lgtm with minor suggestions.

client.go
mdlayher: That's fine.  I just followed the convention that was already in place in this section of the code.  Perhaps they could all be combined into one function?

mdlayher: See `okOrErr` and `eofOrErr`.

davecheney: @mdlayher I have a feeling that there will be more error codes to translate, so one function for all of them sounds like a good plan.

mdlayher:
@davecheney Updated!  Please take a look.

davecheney:
Very nice. Thanks.

PR #73, opened 2016-01-04
*: golint (part 1)

mdlayher:
Fixes a large number of `golint` issues.  I tried to do fix a single type of issue in each commit.  Can squash on request.

The `StatVFS` and `FileStat` types had some of their exported struct fields renamed to fit with `golint`.  If this is unacceptable due to API breakage, we can leave them as-is, though I'd prefer to just try to fall in line with `golint` wherever possible.

Next up would be comments on exported types and possibly renaming some of the `ALL_CAPS` constants.


packet.go
```diff
@@ -612,31 +614,31 @@ func (s sshFxpWritePacket) MarshalBinary() ([]byte, error) {
 	return b, nil
 }
 
-func (p *sshFxpWritePacket) UnmarshalBinary(b []byte) (err error) {
-	if p.Id, b, err = unmarshalUint32Safe(b); err != nil {
+func (s *sshFxpWritePacket) UnmarshalBinary(b []byte) (err error) {
```
davecheney: these should all be `p` for packet, not `s`

packet.go
```diff
@@ -734,63 +736,63 @@ func (p *sshFxpFsetstatPacket) UnmarshalBinary(b []byte) error {
...
-	b = marshalUint32(b, p.Id)
+	b = marshalUint32(b, p.ID)
 	b = marshalUint32(b, p.Length)
 	b = append(b, p.Data[:p.Length]...)
 	return b, nil
 }
 
 func (p *sshFxpDataPacket) UnmarshalBinary(b []byte) (err error) {
-	if p.Id, b, err = unmarshalUint32Safe(b); err != nil {
+	if p.ID, b, err = unmarshalUint32Safe(b); err != nil {
```
davecheney: please delete the named return argument and add
```
var err error
```
at the top.

my opinion is named return arguments should only be used with naked returns, and if there are no naked returns, there should be no named return arguments.

server.go
```diff
@@ -552,10 +552,10 @@ func (p sshFxpFsetstatPacket) respond(svr *Server) error {
 		}
 	}
 
-	return svr.sendPacket(statusFromError(p.Id, err))
+	return svr.sendPacket(statusFromError(p.ID, err))
 }
 
-func errnoToSshErr(errno syscall.Errno) uint32 {
+func errnoToSSHErr(errno syscall.Errno) uint32 {
```
davecheney: I don't think renaming this has improved the name of this function, and it wasn't good to start with. If you want to come up with a new name, that would be great.

server_integration_test.go
```diff
@@ -275,39 +275,39 @@ func (chsvr *sshSessionChannelServer) handleSubsystem(req *ssh.Request) error {
...
+	exitStatus := uint32(1)
+	if runErr == nil {
+		exitStatus = uint32(0)
+	}
+
+	_, exitStatusErr := chsvr.ch.SendRequest("exit-status", false, ssh.Marshal(sshSubsystemExitStatus{exitStatus}))
+	return exitStatusErr
 }
 
 /***********************************************************************************************
```
davecheney: please kill the C style comment blocks while you're here. This one adds nothing, just delete it.

sftp.go
```diff
@@ -185,9 +185,9 @@ func unimplementedPacketErr(u uint8) error {
 	return fmt.Errorf("sftp: unimplemented packet type: got %v", fxp(u))
 }
 
-type unexpectedIdErr struct{ want, got uint32 }
+type unexpectedIDErr struct{ want, got uint32 }
```
davecheney: a horrible name, and more capitals did not help.

davecheney:
Thanks for this. Just a few little nits, as we're nit picking, then i'll merge this.

sftp.go
mdlayher: There are several other errors like this in the same file.  I can't think of a better name at the moment, but perhaps it would be best to address them all as a unit in a second pull request.

mdlayher:
Tried to squash this PR and rebase on master, but I still ended up with another commit in here somehow.  Any ideas how to separate it?

davecheney:
> Any ideas how to separate it?

gotta rebase against master I guess

mdlayher:
Fixed.  That was strange.  Had to `git fetch` to get my master to update.  Feel free to merge.

davecheney:
LGTM

I sent you an invite to join this repo. Feel free to commit this once the build passes.

Thanks again

mdlayher:
Thanks very much!  I will probably continue to use pull requests to ask for feedback, but I certainly appreciate it.

davecheney:
Yes, the general policy is no commits unless there is LGTM agreement.
Thanks.

PR #75, opened 2016-01-07
*: golint: add and fix up comments on exported types

mdlayher:
Added comments and rephrased a few others.

Unexported the `extensionPair` type, because I couldn't find any place where it was returned in the exposed API.

r: @davecheney


packet.go
```diff
@@ -207,7 +207,7 @@ func (p *sshFxInitPacket) UnmarshalBinary(b []byte) (err error) {
 		return err
 	}
 	for len(b) > 0 {
-		ep := ExtensionPair{}
+		ep := extensionPair{}
```
davecheney: var ep extensionPair

packet.go
```diff
@@ -207,7 +207,7 @@ func (p *sshFxInitPacket) UnmarshalBinary(b []byte) (err error) {
 		return err
```
davecheney: Please remove the named return value

mdlayher: Done.

mdlayher: I personally don't like named returns or naked returns at all, outside of for documentation purposes in `interface` declarations.

I'd be happy to remove them throughout in a different PR, if you agree.

davecheney: yes, please kill them with fire.

mdlayher: Opened #76 for further discussion.

mdlayher:
Any more thoughts before this one goes in?

davecheney:
LGTM.

PR #77, opened 2016-01-07
*: remove all named and naked returns

### LINKED ISSUE #76: Proposal: remove all named and naked returns
Named returns can be useful for documentation purposes, but are often unneeded.  Naked returns go hand-in-hand with named returns, and in my personal experience, only lead to confusing code.

I propose removing every named and naked return from this repository, with a few exceptions:

1) interface declarations, e.g.
```go
type Filesystem interface {
    Open(path string) (info *os.FileInfo, err error)
}
```

2) functions which return several values of the same type, hypothetical example:
```go
func Stat(path string) (freeSpace uint64, usedSpace uint64, totalSpace uint64, err error) { }
```

3) capturing panic recovery value into an error (haven't seen this at all in this repository, but it's occasionally useful)

Thoughts?

mdlayher:
Fixes #76 .

r: @davecheney

### DIFF
--- a/client_integration_test.go
+++ b/client_integration_test.go
@@ -1164,14 +1164,14 @@ func TestClientWalk(t *testing.T) {
 	makeTree(t)
 	errors := make([]error, 0, 10)
 	clear := true
-	markFn := func(walker *fs.Walker) (err error) {
+	markFn := func(walker *fs.Walker) error {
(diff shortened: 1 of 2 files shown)


davecheney:
LGTM. Thanks.

PR #78, opened 2016-01-07
packet, server: use a single read-only check

mdlayher:
Expands the `serverRespondablePacket` interface to remove the need for duplicated read-only checks later on.  Many of the packets already have an `id()` method.

Adds a couple methods to allow more readable read/write checking when opening files.

Also cut out an unneeded `fmt.Sprintf()` in favor of `strconv.Itoa()`.

r: @davecheney

### DIFF
--- a/packet_test.go
+++ b/packet_test.go
@@ -213,6 +213,90 @@ func TestRecvPacket(t *testing.T) {
 	}
 }
 
+func TestSSHFxpOpenPacketreadonly(t *testing.T) {
+	var tests = []struct {
+		pflags uint32
+		ok     bool
+	}{
+		{
+			pflags: ssh_FXF_READ,
+			ok:     true,
+		},
+		{
+			pflags: ssh_FXF_WRITE,
+			ok:     false,
+		},
+		{
+			pflags: ssh_FXF_READ | ssh_FXF_WRITE,
+			ok:     false,
+		},
+	}
+
+	for _, tt := range tests {
+		p := &sshFxpOpenPacket{
+			Pflags: tt.pflags,
+		}
+
+		if want, got := tt.ok, p.readonly(); want != got {
+			t.Errorf("unexpected value for p.readonly(): want: %v, got: %v",
+				want, got)
+		}
+	}
+}
+
+func TestSSHFxpOpenPackethasPflags(t *testing.T) {
+	var tests = []struct {
+		desc      string
+		haveFlags uint32
+		testFlags []uint32
+		ok        bool
+	}{
+		{
+			desc:      "have read, test against write",
+			haveFlags: ssh_FXF_READ,
+			testFlags: []uint32{ssh_FXF_WRITE},
+			ok:        false,
+		},
+		{
+			desc:      "have write, test against read",
+			haveFlags: ssh_FXF_WRITE,
+			testFlags: []uint32{ssh_FXF_READ},
+			ok:        false,
+		},
+		{
+			desc:      "have read+write, test against read",
+			haveFlags: ssh_FXF_READ | ssh_FXF_WRITE,
+			testFlags: []uint32{ssh_FXF_READ},
+			ok:        true,
+		},
+		{
+			desc:      "have read+write, test against write",
+			haveFlags: ssh_FXF_READ | ssh_FXF_WRITE,
+			testFlags: []uint32{ssh_FXF_WRITE},
+			ok:        true,
+		},
+		{
+			desc:      "have read+write, test against read+write",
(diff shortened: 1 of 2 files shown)


mdlayher:
I think I broke CI somehow.  Will check it out later.

server.go
```diff
@@ -131,8 +134,15 @@ func (svr *Server) sftpServerWorker(doneChan chan error) {
 			return
 		}
 
+		// If server is operating read-only and a write operation is requested,
+		// return permission denied
+		if dPkt.writing() && svr.readOnly {
+			_ = svr.sendPacket(statusFromError(dPkt.id(), syscall.EPERM))
```
davecheney: why assign to `_`, just skip the assignment

server.go
```diff
@@ -131,8 +134,15 @@ func (svr *Server) sftpServerWorker(doneChan chan error) {
...
+			return
+		}
+
 		//fmt.Fprintf(svr.debugStream, "pkt: %T %v\n", pkt, pkt)
-		dPkt.respond(svr)
+		_ = dPkt.respond(svr)
```
davecheney: same

server.go
```diff
@@ -131,8 +134,15 @@ func (svr *Server) sftpServerWorker(doneChan chan error) {
...
 		}
 
+		// If server is operating read-only and a write operation is requested,
+		// return permission denied
+		if dPkt.writing() && svr.readOnly {
+			_ = svr.sendPacket(statusFromError(dPkt.id(), syscall.EPERM))
+			return
+		}
+
 		//fmt.Fprintf(svr.debugStream, "pkt: %T %v\n", pkt, pkt)
```
davecheney: please delete this commented out code

server.go
```diff
@@ -67,7 +68,9 @@ func (svr *Server) getHandle(handle string) (*os.File, bool) {
 
 type serverRespondablePacket interface {
 	encoding.BinaryUnmarshaler
+	id() uint32
 	respond(svr *Server) error
+	writing() bool
```
davecheney: would `readonly` be a better name for this ? it does have a problem with the zero value making r/w the default, but writing isn't a very good name for a method IMO

server.go
mdlayher: This is a habit from the way I explicitly ignore errors in my personal and work code.  It also stops `errcheck` from generating an error, since you are explicitly stating that you are aware an error is returned, but do not care about its result.

If you'd prefer to remove these, I can do that.

server.go
mdlayher: Sure thing.

server.go
mdlayher: Yeah, my intent was to make the zero-value the "read-only" one, but I suppose I can invert it.

davecheney: Keeping the zero value is useful, but the name of the method isnt very
informative.

mdlayher: Updated to be `readonly`.

mdlayher:
@davecheney Figured out the CI issue.  I used `return` instead of `continue` in a processing loop, so the server didn't process the client's close packet, causing it to hang.

Any thoughts regarding this: https://github.com/pkg/sftp/pull/78#discussion_r49213051?

Once that is resolved, let me know what you think of the PR as a whole.

mdlayher:
@davecheney ping.  Going to start working on a separate PR, but would like to get this merged if possible to reduce later merge conflicts.  Let me know what you think.

packet_test.go
```diff
@@ -213,6 +213,90 @@ func TestRecvPacket(t *testing.T) {
...
+		},
+	}
+
+	for _, tt := range tests {
+		p := &sshFxpOpenPacket{
+			Pflags: tt.pflags,
+		}
+
+		if want, got := tt.ok, p.readonly(); want != got {
+			t.Errorf("unexpected value for p.readonly():\n- want: %v\n-  got: %v",
```
davecheney: one line please, we're only talking about `true` and `false`

packet_test.go
```diff
@@ -213,6 +213,90 @@ func TestRecvPacket(t *testing.T) {
...
+
+	for _, tt := range tests {
+		t.Log(tt.desc)
+
+		p := &sshFxpOpenPacket{
+			Pflags: tt.haveFlags,
+		}
+
+		if want, got := tt.ok, p.hasPflags(tt.testFlags...); want != got {
+			t.Errorf("unexpected value for p.hasPflags(%#v):\n- want: %v\n-  got: %v",
```
davecheney: one line please

davecheney:
LGTM with minor nits about the tests output.

mdlayher:
Updated.  Merging on green.

PR #81, opened 2016-01-11
server: return ServerOption for godoc documentation

mdlayher:
A usability improvement: this will make `WithDebug` and `ReadOnly` appear under `type ServerOption` in godoc.

r: @davecheney


davecheney:
LGTM. Normally I'm against aliasing types unless it is to give them methods, but this is an excellent exception to the rule.

mdlayher:
https://godoc.org/github.com/pkg/sftp#ServerOption

Much nicer.  I also like using type aliases to hint that a package constant should be used instead of a random `int`/`string`/etc., but we'll cross that bridge when we get there.

PR #85, opened 2016-01-16
server: remote rootDir parameter, as it does nothing

mdlayher:
Remove it until it is actually implemented.

r: @davecheney


davecheney:
LGTM, thanks.

PR #88, opened 2016-02-03
add func MkdirAll

lostz:
add func MkdirAll

### DIFF
--- a/client.go
+++ b/client.go
@@ -694,6 +694,37 @@ func (c *Client) Mkdir(path string) error {
 	}
 }
 
+func (c *Client) MkdirAll(path string) error {
+	dir, err := c.Stat(path)
+	if err == nil {
+		if dir.IsDir() {
+			return nil
+		}
+	}
+	i := len(path)
+	for i > 0 && os.IsPathSeparator(path[i-1]) { // Skip trailing path separator.
+		i--
+	}
+	j := i
+	for j > 0 && !os.IsPathSeparator(path[j-1]) { // Scan backward over element.
+		j--
+	}
+	if j > 1 {
+		// Create parent
+		err = c.MkdirAll(path[0 : j-1])
+		if err != nil {
+			return err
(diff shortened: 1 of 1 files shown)


mdlayher:
I like the idea, but this needs at least a test or two to verify its functionality.

client.go
```diff
@@ -694,6 +694,37 @@ func (c *Client) Mkdir(path string) error {
 	}
 }
 
+func (c *Client) MkdirAll(path string) error {
```
mdlayher: Needs a documentation comment.

client.go
```diff
@@ -694,6 +694,37 @@ func (c *Client) Mkdir(path string) error {
...
+			return err
+		}
+	}
+	err = c.Mkdir(path)
+	if err != nil {
+		return err
+	}
+
+	return nil
+
```
mdlayher: Remove this blank line.

client.go
```diff
@@ -694,6 +694,37 @@ func (c *Client) Mkdir(path string) error {
...
+		j--
+	}
+	if j > 1 {
+		// Create parent
+		err = c.MkdirAll(path[0 : j-1])
+		if err != nil {
+			return err
+		}
+	}
+	err = c.Mkdir(path)
```
davecheney: return c.Mkdir(path)

davecheney:
Why don't we just make c.Mkdir do this automatically ? Then we don't have to have two functions with subtly different semantics.

mdlayher:
I'd be okay with that as well. While I feel there could be value in matching the API of the `os` package, I would prefer recursive directory creation 9 times out of 10 in practice.

mordyovits:
Please match the os API.  Anything else would catch people (me) by surprise.

davecheney:
This PR appears to be stale. Closing

If someone wants to reactivate it later please _raise an issue first_.

PR #89, opened 2016-02-24
add serverside StatVFS function, implemented for darwin and linux

marksheahan:
Also have the client return errors if methods are invoked after the server recv loop dies; currently client functions hang if called after the server dies. This can cause problems in multithreaded clients.

### DIFF
--- a/packet.go
+++ b/packet.go
@@ -1,7 +1,9 @@
 package sftp
 
 import (
+	"bytes"
 	"encoding"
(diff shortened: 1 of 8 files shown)


davecheney:
@marksheahan can you please rebase this PR against master. Thank you.

davecheney:
@marksheahan I just rebased with 2570687, if this passes tests then I'll merge it.

PR #94, opened 2016-05-02
Abstract from os API

an2deg:
Moving all file operations in Server to StorageBackend interface. With those changes we can extend SFTP server by implementing custom StorageBackends. 

Example usage:

server, err := sftp.NewServer(
            channel,
            channel,
            sftp.WithStorage(custom_storage)
)


davecheney:
Please don't just sling code. Instead, would you please open an issue to discuss your oroblem and we can workshop solutions from there, one of which may be to introduce a vfs layer.

davecheney:
Closing this PR as it is stale and not the API as discussed.

PR #96, opened 2016-05-16
export everything needed for external packet handlers

eikenb:
Renamed as followed to export...
fxp                     -> Fxpkt
ssh_FX*                 -> SSH_FX*
sshFx*                  -> SshFx*
server.sendPacket       -> server.SendPacket
statusFromError()       -> StatusFromError
serverRespondablePacket -> ServerRespondablePacket
pkt.respond             -> pkt.Respond

Issue #95, opened 2016-05-02
Provide some FS API for Server


an2deg:
In my project I would like to use github.com/pkg/sftp as an embedded SFTP server.
Would be nice to make Server more extendable in terms of working with FS or having the following features:
1. Restrict user to use only specific directory (e.g. chroot)
2. Provide some kind of notifications when file upload is completed or user session is closed

In my pull request #94 I've moved all file operations into FileStorageBackend and made it replaceable.  Maybe this solution is little bit dirty so let's workshop here the better solution

davecheney:
Trying to scope the connection to a specific director is #92. I'm very cautious about doing anything here because I don't want to over promise. The second we say that the sftp server cannot break out of this chroot, we're writing a cheque that someone else is trying to cash.

With respect to the rest of this issue, a file upload notification should not require adding an entire virtual fs layer, and closing a user session has nothing to do with a virtual fs layer.

I'm really not keen on adding a virutal fs layer because there are always so many around for Go (there are two in the x/tools repo) and none have attracted any sort of following, adding yet another won't help the situation. Is there is another way to acomplish these goals without adding a vfs layer ?

an2deg:
What do you think about using callbacks?

For example let's use the signature for callback function:
    type SftpCallback(pktType, serverRespondablePacket) (serverRespondablePacket, error)

And Server structure will have a field with that type which by default could be null or a function with does nothing .

With those changes we can use it in sftpServerWorker() function right after decodePacket():

		dPkt, err := svr.decodePacket(pkt.pktType, pkt.pktBytes)
		if err != nil {
			fmt.Fprintf(svr.debugStream, "decodePacket error: %v\n", err)
			doneChan <- err
			return
		}
    
                dPkt, err = srv.callback(pkt.pktType, dPkt)
                if err != nil {
			fmt.Fprintf(svr.debugStream, "callback error: %v\n", err)
			doneChan <- err
			return
		}

It will allow a user to react to some types of packets. For example we can do some emulation of chroot here and send some notifications.

My usecase is relatively simple: I want to allow user to upload some files by SFTP and when upload finished - start processing those files. Maybe you could come up with better solution.

eikenb:
I am also interested in more flexibility for server side storage. In my case it is to implement an SFTP to S3 gateway service. I was considering a similar abstraction/interface for either the files or the filesystem as well. 

[...]

PR #97, opened 2016-05-16
Responder interface prototype

eikenb:
This pull request builds on pull request #96, first commit here is same as that. Second commit is the important part.

I left most of the filesystem handling code alone so that the changes required for the Responder interface are clear. In the final version it probably makes sense to migrate the filesystem handling into the filesystem_responders.go along with a struct for the openFiles, nextHandle, etc. Though it really doesn't affect the ability to use different backends, so it isn't necessary.

### DIFF
--- a/filesystem_responders.go
+++ b/filesystem_responders.go
@@ -1,0 +1,50 @@
+package sftp
+
+import "fmt"
+
+func FilesystemResponder(pktType Fxpkt) (ServerRespondablePacket, error) {
+	var pkt ServerRespondablePacket
+	switch pktType {
+	case SSH_FXP_INIT:
+		pkt = &SshFxInitPacket{}
+	case SSH_FXP_LSTAT:
+		pkt = &SshFxpLstatPacket{}
+	case SSH_FXP_OPEN:
+		pkt = &SshFxpOpenPacket{}
+	case SSH_FXP_CLOSE:
+		pkt = &SshFxpClosePacket{}
+	case SSH_FXP_READ:
+		pkt = &SshFxpReadPacket{}
+	case SSH_FXP_WRITE:
+		pkt = &SshFxpWritePacket{}
+	case SSH_FXP_FSTAT:
+		pkt = &SshFxpFstatPacket{}
+	case SSH_FXP_SETSTAT:
+		pkt = &SshFxpSetstatPacket{}
+	case SSH_FXP_FSETSTAT:
+		pkt = &SshFxpFsetstatPacket{}
+	case SSH_FXP_OPENDIR:
+		pkt = &SshFxpOpendirPacket{}
+	case SSH_FXP_READDIR:
+		pkt = &SshFxpReaddirPacket{}
+	case SSH_FXP_REMOVE:
+		pkt = &SshFxpRemovePacket{}
+	case SSH_FXP_MKDIR:
+		pkt = &SshFxpMkdirPacket{}
+	case SSH_FXP_RMDIR:
+		pkt = &SshFxpRmdirPacket{}
+	case SSH_FXP_REALPATH:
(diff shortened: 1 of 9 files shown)


eikenb:
The travis failure was only with go 1.4.3 which I don't have handy right off. I'll figure out the problem if it is decided that this idea is worth pursuing.

davecheney:
This diff is dirty, could you please rebase it on top of #96, so it's clear what has been added in this branch. Thanks.

filesystem_responders.go
```diff
@@ -0,0 +1,50 @@
+package sftp
+
+import "fmt"
+
+func FilesystemResponder(pktType Fxpkt) (ServerRespondablePacket, error) {
```
davecheney: please add a comment

server.go
```diff
@@ -65,6 +66,8 @@ func (svr *Server) getHandle(handle string) (*os.File, bool) {
 	return f, ok
 }
 
+type Responder func(Fxpkt) (ServerRespondablePacket, error)
```
davecheney: please add a comment or make this type private

davecheney:
What's the purpose of this PR ? It looks like you're making all the parsing and packet handling logic of this package public so something else can implement a very complicated API. I'm not in favor of this, it means the API of this package, from someone wanting to implement a virtual file system is basically the draft SFTP v2 RFC. I'm sorry if I have misunderstood this purpose of this change.

davecheney:
This PR does not list an issue, but i'm guessing this is #95 in which case this is not what I thought was going to happen. I thought that you were proposing something like the http api,

sftp.Serve(path, handlerFunc())

Where handler would get some kind of request structure that had the operation, READ, OPEN, READDIR, DELETE, etc as well as a few bits of metadata like the path to operate on. 

This is the API I was expecting to see, I'm sorry if I did not make myself clear.

eikenb:
Re: dirty diff..

The diff is as clean as it can get for a pull requests that depends on another pull request that hasn't been merged yet. The branch used was based off the branch used for #96 and is up to date. But the pull request is against your repo which doesn't have that branch.

I structured the commits such that all the code from the 2 pull requests are in separate commits. Just look at the 2nd commit to see the changes.

eikenb:
Example usage. To clarify how it would work in practice.

		
    server, err := sftp.NewServer(channel, channel,
			sftp.WithDebug(debugStream), sftp.ReadOnly(),
			sftp.UseResponder(MyResponder))
...

	func MyResponder(pktType sftp.Fxpkt) (sftp.ServerRespondablePacket, error) {
		var pkt sftp.ServerRespondablePacket
		var err error
		switch pktType {
		case sftp.SSH_FXP_OPEN:
			pkt = &MySshFxpOpenPacket{}
		default:
			pkt, err = sftp.FilesystemResponder(pktType)
		}
		return pkt, err
	}

	// always embed to get marshalling related methods
	type MySshFxpOpenPacket struct {
		sftp.SshFxpOpenPacket
	}

	// Simple example that could extend standard use
	func (p MySshFxpOpenPacket) Respond(srv *sftp.Server) error {
		fmt.Println("Open: ", p.Path)
		fmt.Println("Do more stuff here?")
		return p.SshFxpOpenPacket.Respond(srv)
	}

eikenb:
And yes, this is in relation to #95. Sorry, I'm used to referencing pull requests from issues, not other way around.

davecheney:
This change, as I understand it basically outsources the responding to every package to the caller of this package. I'm sorry but I don't want this API. It's not an API that I can support, the surface are is too large. 

Please see my comments on #95 for an alternative solution.

davecheney:
Closing because this is not the API that I want to see. Please understand my gratitude for the work you have put into this, but this is not something I want to see merged.

PR #108, opened 2016-05-31
make example server respect readOnly flag

rjocoleman:
Currently the example server has a flag `-R` that should toggle `sftp.ReadOnly`.

`sftp.ReadOnly` is hard-coded in `sftp.NewServer` so this flag is ineffective.

This PR changes that and provides an example of way to send an arbitrary number of `sftp.ServerOption`'s to `sftp.NewServer`


davecheney:
LGTM. Thanks.

PR #109, opened 2016-06-07
Make SSH_FXP_STATUS error message optional

akupas:


davecheney:
Can you please amend the commit message to include why you are making this
change.

akupas:
I've amended the commit message with the purpose of this change.

davecheney:
Can I please ask that you follow the contribution procedure documented in the readme; issue first, then code.

client.go
```diff
@@ -1159,13 +1159,16 @@ func normaliseError(err error) error {
...
 func unmarshalStatus(id uint32, data []byte) error {
+	var lang string
 	sid, data := unmarshalUint32(data)
 	if sid != id {
 		return &unexpectedIDErr{id, sid}
 	}
 	code, data := unmarshalUint32(data)
-	msg, data := unmarshalString(data)
-	lang, _, _ := unmarshalStringSafe(data)
+	msg, data, err := unmarshalStringSafe(data)
```
davecheney: Please change this to
```
        msg, data, err := unmarshalString(data)
        if err != nil {
                return err
        }
        lang, _, _ := unmarshalStringSafe(data)
```
My reading of https://tools.ietf.org/html/draft-ietf-secsh-filexfer-02 indicates that the message is not optional.

davecheney:
Thank you for creating #110 I now understand the problem. Please review my comments, thank you.

akupas:
Thanks for your comments. I saw this section at the end of the RFC document you linked:

> 10.1 Changes between versions 3 and 2
> 
>    o  The SSH_FXP_READLINK and SSH_FXP_SYMLINK messages were added.
> 
>    o  The SSH_FXP_EXTENDED and SSH_FXP_EXTENDED_REPLY messages were
>       added.
> 
>    o  The SSH_FXP_STATUS message was changed to include fields `error
>       message' and `language tag'.

It seems like the definition of SSH_FXP_STATUS changed in V3. What do you think about making the handling of the SSH_FXP_STATUS message backward compatible with version 2? If not, then the unmarshaling of the language tag should also return an error (as https://github.com/pkg/sftp/commit/e6dc896441ceed02c620b4309bd01de784e360f0 made it optional). 

If errors should be returned, I propose making a shortPacketErr error type, so one can handle it. (e.g. I would like to use sftp.Walk but ignore these kinds of errors while stepping through the file system)

Please let me know which way to go forward.

davecheney:
@akupas i'm honestly not sure which version of the _draft_ rfc this package implements. It is one of my frustrations that there are no RFCs for sftp, just expired drafts.

client_test.go
```diff
@@ -87,13 +88,67 @@ func TestFlags(t *testing.T) {
...
+			desc:   "request identifier mismatch",
+			reqID:  2,
+			status: idCodeMsgLang,
+			want:   &unexpectedIDErr{2, requestID},
+		},
+	}
+
+	for _, tt := range tests {
+		func() {
+			defer func() {
```
davecheney: You can drop this, the testing runner will catch the panic.

davecheney:
@akupas i'm sorry I don't know how to proceed. There is no official RFC for sftp. Please make the change I suggested and we can merge this PR as is. If the remote cisco client cannot cope with it, we can tackle that in a further PR.

Issue #110, opened 2016-06-08
Missing 'error message' in SSH_FXP_STATUS causing 'index out of range' panic


akupas:
The SSH-2.0-Cisco-2.0 server implementation does not return the SSH_FXP_STATUS 'error message' in response to a ReadDir packet causing an "index out of range" panic. The sftp unix utility gracefully handles this case. I propose that the 'error message' is made optional in unmarshalStatus to handle this case.

davecheney:
Can you please include the panic message.

akupas:
goroutine 1 [running]:
panic(0xc4b2e0, 0xc82000a140)
	go/gc/src/runtime/panic.go:478 +0x3cb
src/github.com/pkg/sftp/sftp.unmarshalStatus(0xc800000004, 0xc8204ce125, 0x0, 0x0, 0x0, 0x0)
	src/github.com/pkg/sftp/client.go:1167 +0x5d9
src/github.com/pkg/sftp/sftp.(*Client).close(0xc8204ba0f0, 0xc82040f444, 0x4, 0x0, 0x0)
	src/github.com/pkg/sftp/client.go:468 +0x169
panic(0xc4b2e0, 0xc82000a140)
	go/gc/src/runtime/panic.go:443 +0x4f2
src/github.com/pkg/sftp/sftp.unmarshalStatus(0xc800000003, 0xc82040fec5, 0x0, 0x0, 0x0, 0x0)
	src/github.com/pkg/sftp/client.go:1167 +0x5d9
src/github.com/pkg/sftp/sftp.(*Client).ReadDir(0xc8204ba0f0, 0xd9b1d0, 0x7, 0x0, 0x0, 0x0, 0x0, 0x0)
	src/github.com/pkg/sftp/client.go:226 +0x353
main.New(0xc820332000, 0xc8202a9a40, 0x0, 0x0)
	ssh.go:122 +0x8a3
main.main()
	ssh.go:217 +0x26b

panic: runtime error: index out of range
	panic: runtime error: index out of range

goroutine 1 [running]:
panic(0xc4b2e0, 0xc82000a140)
	go/gc/src/runtime/panic.go:483 +0x3ef
src/github.com/pkg/sftp/sftp.unmarshalStatus(0xc800000004, 0xc8204ce125, 0x0, 0x0, 0x0, 0x0)
	src/github.com/pkg/sftp/client.go:1167 +0x5d9
[...]

PR #124, opened 2016-06-23
Make SSH_FXP_STATUS "message" field optional.

akupas:
Follow-up to issue #110 and PR #109. I can read directories with the last commit, but reading files is now a problem...

Cisco SSH 2.0 server doesn't fill out the "message" field,
even though it returns valid status codes. This makes
File.Read method error out without actually reading the
file due to unmarshalStatus returning a "packet too short"
error.

### DIFF
--- a/client.go
+++ b/client.go
@@ -1085,10 +1085,7 @@ func unmarshalStatus(id uint32, data []byte) error {
 		return &unexpectedIDErr{id, sid}
 	}
 	code, data := unmarshalUint32(data)
-	msg, data, err := unmarshalStringSafe(data)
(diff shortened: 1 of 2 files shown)


akupas:
Close and re-opened to trigger new travis build since previous test failed to data race, which doesn't appear to be related to this PR.

eikenb:
This seems perfectly reasonable fix to deal with an out of spec sftp server. I can see no other effect that to make the message optional (just like it does for the lang value now). I'll let it sit for a day or so in case I think of something, but I'll probably merge this soon.

PR #127, opened 2016-07-25
Request

eikenb:
Related ticket: #95 

This PR contains a request based API for extending the SFTP server with arbitrary backends. It was inspired by a mix of the net/http package and the linux inotify system. No existing code is touched, this only adds new code. This required a bit of duplication in places, but I wanted to leave the existing code as is. I specifically tried to make the Handlers code not into a general filesystem abstraction. It is very specific to the SFTP protocol and tries to be lightweight (with 4, 1 method interfaces).

It is mostly done including at least some documentation, examples, and tests. The files are all named starting with "request" as well as ./examples/request-server/main.go. The public method to create the new server is in request-server.go as are the worker threads, so it is a good place to start working through the code. The request-readme.md contains the initial docs which hopefully introduces how it is supposed to work.

If there are any changes you'd like to see or questions you have please let me know. Thanks.

### DIFF
--- a/examples/request-server/main.go
+++ b/examples/request-server/main.go
@@ -28,7 +28,9 @@ func main() {
 	flag.Parse()
 
 	debugStream := ioutil.Discard
-	if debugStderr { debugStream = os.Stderr }
+	if debugStderr {
+		debugStream = os.Stderr
+	}
 
 	// An SSH server is represented by a ServerConfig, which holds
 	// certificate details and handles authentication of ServerConns.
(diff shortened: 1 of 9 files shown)


eikenb:
I've fixed all the linter changes except a few of the underscore ones that I disagree with. If you feel strongly about them I can fix them as well, I just think they are appropriate in some cases to help differentiate between variables and method names.

eikenb:
Now that everything seems OK with this PR I'm going to leave it alone so you can look it over without anything changing. Thanks.

eikenb:
I was adding some additional tests and found a bug in out of order packets when uploading larger files. I'm going to close this pull request until I get that fixed then create a new pull request. Sorry for the spam.

The code will still be available in my repo, in the request branch (github.com/eikenb/sftp/tree/request).

PR #129, opened 2016-08-02
Request based server

eikenb:
Related ticket: #95 

I've fixed the issues I found since PR #127 and added tests to cover them. This is the same code otherwise and below is the same note from it. If 

This PR contains a request based API for extending the SFTP server with arbitrary backends. It was inspired by a mix of the net/http package and the linux inotify system. No existing code is touched, this only adds new code. This required a bit of duplication in places, but I wanted to leave the existing code as is. I specifically tried to make the Handlers code not into a general filesystem abstraction. It is very specific to the SFTP protocol and tries to be lightweight (with 4, 1 method interfaces).

It is mostly done including at least some documentation, examples, and tests. The files are all named starting with "request" as well as ./examples/request-server/main.go. The public method to create the new server is in request-server.go as are the worker threads, so it is a good place to start working through the code. The request-readme.md contains the initial docs which hopefully introduces how it is supposed to work.

If there are any changes you'd like to see or questions you have please let me know. Thanks.

### DIFF
--- a/examples/request-server/main.go
+++ b/examples/request-server/main.go
@@ -28,7 +28,9 @@ func main() {
 	flag.Parse()
 
 	debugStream := ioutil.Discard
-	if debugStderr { debugStream = os.Stderr }
+	if debugStderr {
+		debugStream = os.Stderr
+	}
 
 	// An SSH server is represented by a ServerConfig, which holds
 	// certificate details and handles authentication of ServerConns.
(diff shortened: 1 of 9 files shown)


eikenb:
Got so happy I got the test I wanted working I forgot to re-run the race detector.

eikenb:
This is embarrassing, but I'm going to close another PR. That data race is being a pain to nail down and I hate leaving a unready PR open. I'm sorry for the red-herring. I'm sure this will help me remember to run the data race detector more regularly. I'll hopefully be submitting again soon.

PR #136, opened 2016-09-08
client: satisfy go vet

zenazn:
The specific issue was:
```
client.go:506: unreachable code
```

There's one additional `vet` issue:
```
example_test.go:13: Example should be niladic
```
but I'm not sure how that one ought to be fixed


davecheney:
Thank you

PR #139, opened 2016-09-22
misc test fixes recommended by `go vet`

gofish:
**Before**

```
client_integration_linux_test.go:32: possible formatting directive in Fatal call
client_integration_linux_test.go:36: possible formatting directive in Fatal call
client_integration_linux_test.go:40: possible formatting directive in Fatal call
exit status 1
example_test.go:13: Example should be niladic
exit status 1
```

The `Fatal` vs. `Fatalf` calls are trivial fixes. The Example function change is needed because godoc will ignore parameters to Example functions (see https://godoc.org/github.com/pkg/sftp#ex-package)


eikenb:
Looked good. Thanks.

Issue #146, opened 2016-10-20
Some SFTP appliances return file does not exist for directories during a Remove


crazed:
Similar to #62,  non-compliant servers return unexpected results when you try to call Remove on a directory. Maybe it is best to expose RemoveDirectory, as I bet there's endless servers out there doing all kinds of weird things. The specific server I'm dealing with winds up with error:

"file does not exist"

I did not have time to track down which specific server this is, or what packet that error correlates with, but I do know with the standard openssh client an rmdir works fine, but an rm fails with the same response.

eikenb:
After reading through #62 and #35 I agree that the best solution would probably be to make RemoveDirectory public. Parity with go's os package is nice, but given the non-compliance of some servers out there having a way to work around it when they aren't in compliance would be better than trying to fix each one. At least at this stage in the project.

I'll leave the comment open for a time to allow for others to chime in before moving forward with the change.

eikenb:
I see you already did a PR for this, #147. Thanks.

eikenb:
Merged #147, so this is fixed.

PR #147, opened 2016-10-21
Make remove directory public

crazed:
This addresses #146, after further investigation whatever server I'm dealing with (I think it's a GlobalScape something or other), I do not get an sftp.StatusError back, just a string error stating "file does not exist". 

In my use case, I've already done a file stat so I just check IsDir() and use RemoveDirectory directly.


eikenb:
As I said in #146, I believe this is the best option for dealing with non-compliant servers at the moment. So I'm merging.

PR #152, opened 2017-01-11
Fix memory leak and hangs in WriteTo.

### LINKED ISSUE #150: Hanging occurs when opening multi-connection download
I create more than one sftp.NewClient to do the download. One or two seconds later all connections hang up. And no data to be downloaded. 

If there is one connection, all is ok.

The 'WriteTo' or 'Copy' has no response forever:
nSize, err := rFile.WriteTo(wFile)
//nSize, err := io.Copy(wFile, rFile)

theraphim:
Leak is when requests are arriving out of order - they are never cleaned from the map. Hang is when Writer returns an error - because of the way inFlight is used, the function is stuck forever on receiving from channel.

This may fix #150. I don't have tests sadly, to properly test this and other corner cases I'll need to do a larger refactoring. I tested this in production instead.

### DIFF
--- a/client.go
+++ b/client.go
@@ -778,16 +778,21 @@ func (f *File) WriteTo(w io.Writer) (int64, error) {
 
 	var copied int64
 	for firstErr.err == nil || inFlight > 0 {
-		for inFlight < desiredInFlight && firstErr.err == nil {
-			b := make([]byte, f.c.maxPacket)
-			sendReq(b, offset)
-			offset += uint64(f.c.maxPacket)
-			if offset > fileSize {
-				desiredInFlight = 1
+		if firstErr.err == nil {
+			for inFlight+len(pendingWrites) < desiredInFlight {
+				b := make([]byte, f.c.maxPacket)
+				sendReq(b, offset)
+				offset += uint64(f.c.maxPacket)
+				if offset > fileSize {
+					desiredInFlight = 1
+				}
 			}
 		}
 
 		if inFlight == 0 {
+			if firstErr.err == nil && len(pendingWrites) > 0 {
+				return copied, errors.New("internal inconsistency")
+			}
 			break
 		}
 		select {
@@ -816,6 +821,8 @@ func (f *File) WriteTo(w io.Writer) (int64, error) {
 					nbytes, err := w.Write(data)
 					copied += int64(nbytes)
 					if err != nil {
+						// We will never receive another DATA with offset==writeOffset, so
+						// the loop will drain inFlight and then exit.
 						firstErr = offsetErr{offset: req.offset + uint64(nbytes), err: err}
 						break
 					}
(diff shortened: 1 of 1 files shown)


theraphim:
Maybe it makes sense to drop 1.4.3 from the CI?

eikenb:
I'd prefer it if the 2 issues were split out, 1 PR for the leak and 1 PR for the hang. It'd also be great if there were a basic test for each, as I'd really like to reproduce the problem and then see the fix in action.

For the leak, I assume the map you are referring to is pendingWrites. From my reading it would get freed/gc'd after the function returned, but it would keep the all incoming data in memory until that point. So not so much a leak as just inefficient memory use. Or am I missing something?

eikenb:
Also, I've removed 1.4.3 from travis testing.

theraphim:
Greetings.

So, in reverse order.
You may call this inefficient memory use but to me if I use it to stream 16 gigabytes and it'll eat 16 gigabytes I'll call this a leak.

As for testing - yes it would be nice to have tests, unfortunately as is the code isn't very testable, and I don't want to refactor incorrect code into another incorrect code.

eikenb:
I agree that saving all the data in memory even after being sent to the writer is bad behavior and should be addressed. In other words, putting semantics of "leaks" aside, I agree.

Would you consider reworking the PR to better split up the changes for the 2 fixes. Either as 2 PRs or as 2 commits in one? I am a strong believer in 1 commit/changeset per issue.

Thanks.

theraphim:
2 commits now.

eikenb:
Thank you for splitting up the commits. I'll look at getting it merged soon. I'm first trying to see what I can do about all the races in the integration tests that cause the intermittent failures of the automated/travis testing. Your's fails every time on travis, but usually passes when I run the tests locally. I can get it to race occasionally locally and am trying to track those down.

theraphim:
Any luck merging this?

eikenb:
I will get to it soon. I just pushed a fix to a race and found that the upstream ssh lib package made a change that breaks one of our tests. I've located the commit that breaks things and am currently working to figure out where the bug is. Once I have the tests running reliably I'll work on getting your PR merged first.

eikenb:
Filed a ticket on the problem (#155). Tracked it down to an upstream bug introduced a couple weeks ago in the x/crypto/ssh lib. There is already a change out for review that fixes it. Our travis tests will continue to fail until it gets reviewed and applied. I have it applied locally and all tests are passing here.

AlexLi1001:
@theraphim have you email address? I have some technical problem about sftp client response to discuss . My email address: liji@huawei.com. Thanks!

eikenb:
FYI, I've reviewed this and am happy with it. I'll merge it as soon as the bug breaking the testing is resolved as it looks like they intend to fix that one very soon.

eikenb:
Just added a hack/workaround to the travis config so it would pass (pulls in the fix). So I'll be merging this...

PR #159, opened 2017-02-23
sftp: get windows compiling and allow tests to run that can pass

### LINKED ISSUE #157: request-example prevents code from compiling on windows
Updated the package with glide and now compiling throws an error:

`vendor\github.com\pkg\sftp\request-example.go:190: undefined: syscall.Stat_t`

kardianos:
Separate out windows specific calls to separate build files.
Unix specific calls, such as unix sockets or unix syscalls
give windows equivalents. Do not yet provide a windows
equivalent to createRequestServerPair as doing so exposes
other windows breakages. For now simply get it to compile on
windows. Passing the client-server tests will be handled in a follow
up PR.

Fixes #157

### DIFF
--- a/os_unix.go
+++ b/os_unix.go
@@ -1,0 +1,11 @@
+//+build !windows
+
+package sftp
+
+import "syscall"
+
+// fakeFileInfoSys returns a platform dependent value normally
+// returned from the os#FileInfo.Sys method.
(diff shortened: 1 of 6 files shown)


eikenb:
#160 supersedes this.

PR #162, opened 2017-02-25
client,packet: remove redundant select & return

unclejack:
This PR removes a few redundant single case selects and one redundant return.


eikenb:
Less indenting is always welcome. Thanks.

PR #164, opened 2017-02-28
Pass actual errors from FileReader/FileWriter implementations

### LINKED ISSUE #163: Every non-nil error from FileReader is replaced with syscall.EBADF
Currently [every non-nil error from FileReader is replaced with syscall.EBADF](https://github.com/pkg/sftp/blob/94df0a367c78a139a5351d2a3ec204b61170ded9/request.go#L144-L146). Which I believe is a bug, as implementation of sftp.FileReader has no way or properly signaling clients about file absence/permissions issues using predefined errors from os or syscall packages.

Clients based on sftp package log errors like this:

    sftp: "bad file descriptor" (SSH_FX_FAILURE)

Default sftp client used in macOS reports issue like this:

    Couldn't read from remote file "non-existen-file" : Failure

If `syscall.EBADF` value is replaced by actual error returned from FileReader, it's becomes possible to properly handle file absence/permission errors, i.e. if handler returns os.ErrNotExist or syscall.ENOENT, sftp client sees falure like this:

    Couldn't read from remote file "non-existen-file" : No such file or directory

This change also breaks TestRequestReadFail test which may probably need to be updated; I'll look into this.

artyom:
Fixes #163

### DIFF
--- a/request-server_test.go
+++ b/request-server_test.go
@@ -165,7 +165,7 @@ func TestRequestReadFail(t *testing.T) {
 	contents := make([]byte, 5)
 	n, err := rf.Read(contents)
 	assert.Equal(t, n, 0)
-	assert.IsType(t, &StatusError{}, err)
+	assert.Exactly(t, os.ErrNotExist, err)
 }
 
 func TestRequestOpen(t *testing.T) {
--- a/request.go
+++ b/request.go
@@ -143,7 +143,7 @@ func fileget(h FileReader, r Request) (responsePacket, error) {
 	if reader == nil {
 		reader, err = h.Fileread(r)
 		if err != nil {
-			return nil, syscall.EBADF
+			return nil, err
 		}
 		r.setState(reader)
 	}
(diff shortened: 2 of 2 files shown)


unclejack:
LGTM

ping @eikenb

eikenb:
I'll take a look at this soon. Been busy of late and am dedicating today to see if I can make any headway on another ticket. This will probably be fine, my original code was very conservative about making sure errors returned from the request setup were compatible with the existing sftp error handling code. Now that it is an option to make sure it all works together doing more to conserve the real errors sounds like a good idea.

eikenb:
I didn't return the outer error in an attempt to make sure the error returned would be understood by the code that convert errors to status packets. Passing the actual error back means the handler implementer will need to make sure the error makes sense to the packet/error conversion code. But I generally prefer assuming the dev knows what they are doing so I was probably being overly paranoid  here.

TLDR; accepting

PR #171, opened 2017-03-15
server_standalone: handle read only properly

unclejack:
This PR fixes the read only flag for the standalone server. It was always read only before this change.


eikenb:
Thanks

PR #174, opened 2017-03-27
Calm down go race detector and fix livelock with concurrent clients

pborzenkov:
The first patch in the pull request fixes go race detector errors by defining Server methods on pointer receiver instead of on the value one to avoid copying.

The second patch fixes livelock when the server is accessed by a concurrent client (for example, sftp.Client used from multiple goroutines simultaneously).

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -102,11 +102,13 @@ func (c *clientConn) sendPacket(p idmarshaler) (byte, []byte, error) {
 func (c *clientConn) dispatchRequest(ch chan<- result, p idmarshaler) {
 	c.Lock()
 	c.inflight[p.id()] = ch
+	c.Unlock()
 	if err := c.conn.sendPacket(p); err != nil {
+		c.Lock()
 		delete(c.inflight, p.id())
+		c.Unlock()
 		ch <- result{err: err}
 	}
-	c.Unlock()
 }
 
 // broadcastErr sends an error to all goroutines waiting for a response.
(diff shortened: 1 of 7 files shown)


pborzenkov:
The single Travis fail on osx/tip is because the modification time on /dev/null changed between system sftp and go sftp clients invocations:


```
--- FAIL: TestServerCompareSubsystems (1.48s)
	server_integration_test.go:439: outputs differ, go:
		crw-rw-rw-    1 0        0               0 Mar 27 15:56 null
		openssh:
		crw-rw-rw-    1 root     wheel           0 Mar 27 15:57 null

```

eikenb:
I originally had the sorting on the incoming queue as well but took it out thinking TCP would ensure ordering if the client sent them in order (was deep into fixing winscp issues at the time, which sends in order). Thinking about it I guess I see that if the client code uses a pool of client connections it can send the packets out of order just like the server can handle them out of order. So patch 2 makes sense to me.

For patch 1, where do you see a race. Does it happen using existing tests? If so, which one? I'd like to figure out a way to reproduce the issue if possible.

Thanks.

pborzenkov:
@eikenb The race is triggered by the tests in my project, which uses sftp package.
I will write the test for sftp that triggers this race and update the pull request with it.

eikenb:
That is great! Thanks.

pborzenkov:
@eikenb I've added the test that triggers the race detector (patch 3). While doing this, I've also found a deadlock (only when Client and Server are connected via io.Pipe) and fixed it in patch 5.

eikenb:
Thank you.  I'll try to get this reviewed today.

eikenb:
Everything looked good. Thanks again.

PR #175, opened 2017-04-07
Fix race between client's close and write

pborzenkov:
After commit 506f3a7 which removed clientConn mutex around
conn.sendPacket, it is now possible for conn.Close and conn.Write to
race (https://travis-ci.org/pkg/sftp/jobs/219048782#L1838). The problem
is really hard to reproduce, but I believe this patch fixes it.

Now clientConn.Mutex protects 'inflight' map only, and conn.Mutex
serializes connection writes/close.

Signed-off-by: Pavel Borzenkov <pavel.borzenkov@gmail.com>

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -56,9 +56,9 @@ func (c *clientConn) loop() {
 // appropriate channel.
 func (c *clientConn) recv() error {
 	defer func() {
-		c.Lock()
+		c.conn.Lock()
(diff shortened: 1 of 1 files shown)


eikenb:
It makes sense to lock the wrapped conn object anyways, as that is what you are closing.
+1

Issue #172, opened 2017-03-15
Section-7 spec issue; ensure incoming packets are processed in order


eikenb:
The problem is that the spec dictates in [section-7](https://tools.ietf.org/html/draft-ietf-secsh-filexfer-02#section-7) that the packets don't have to be sent to the client in order but they have to be processed as if they were handled in order (ie. they don't have to be handled exactly in-order as long as the result is the same as if they were). The worker pool doesn't guarantee that the packets will be processed as-if they were in order. This was done to get throughput on transferring large files (packet order doesn't matter for read/write) at the cost of possible issues with non-transferring commands.

The fix I was considering was to change the server to use a simple state machine about whether to use a pool of workers or a single worker. It would use a single worker for all non-transferring commands and pool for uploads/downloads. It would need to be a state machine to enforce things like waiting for the upload/download pool to finish before processing the Close command.

Note this arose while looking into #132.

PR #176, opened 2017-04-09
Packet process order

eikenb:
This is a fix for issue #172. It defaults to a single worker for everything _except_ all packets between an file open and close packet. For the packets working on the open file it uses the pool of 8 as it was before for everything. After receiving the close packet, it makes sure all read/write packets have been handled before handling the close.

This should provide the provide the guaranteed in-order processing required by the spec while allowing higher throughput on file uploads/downloads.

I'll wait a few days before merging in case anyone has a comment.

### DIFF
--- a/packet-manager.go
+++ b/packet-manager.go
@@ -1,6 +1,9 @@
 package sftp
 
-import "encoding"
+import (
+	"encoding"
+	"sync"
+)
 
 // The goal of the packetManager is to keep the outgoing packets in the same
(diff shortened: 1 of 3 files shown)


eikenb:
@unclejack If you have the time and the interest I'd be curious if this PR fixing any of the issues you were still having with Winscp and Cyberduck. It should fix the last issues with Winscp and ordering.

eikenb:
Fixes #172

PR #191, opened 2017-08-10
various improvements / potential fix for issue #8

sandreas:
On windows systems the directory separator is a backslash (\). This leads to misbehavior of the library in some cases. I tried to fix this issues without changing to much code.

According to RFC https://tools.ietf.org/html/draft-ietf-secsh-filexfer-02, SFTP always uses slashes. 

Other improvements:
- added elementary unit-test for runLs (server.go)
- changed stub version of "runLs" to provide an as valid as possible longname attribute (i tried to follow the windows 10 ubuntu layer ls command, which is an improvement, because widely used open source sftp client FileZilla can now be used connecting windows sftp servers)

### DIFF
--- a/request-server.go
+++ b/request-server.go
@@ -131,7 +131,7 @@ func (rs *RequestServer) packetWorker(pktChan chan requestPacket) error {
 			rs.closeRequest(handle)
 			rpkt = statusFromError(pkt, nil)
 		case *sshFxpRealpathPacket:
-			rpkt = cleanPath(pkt)
+			rpkt = cleanPacketPath(pkt)
 		case isOpener:
 			handle := rs.nextRequest(requestFromPacket(pkt))
 			rpkt = sshFxpHandlePacket{pkt.id(), handle}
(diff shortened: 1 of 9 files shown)


eikenb:
Reviewing...

request-server.go
```diff
@@ -183,10 +184,12 @@ func (rs *RequestServer) packetWorker(pktChan chan requestPacket) error {
 func cleanPath(pkt *sshFxpRealpathPacket) responsePacket {
 	path := pkt.getPath()
 	if !filepath.IsAbs(path) {
-		path = "/" + path
+		// prevent double slash (e.g. on windows, / paths are not absolute)
+		path = "/" + strings.TrimPrefix(path, "/")
```
davecheney: filepath.Clean may be more appropriate here

request-server.go
```diff
@@ -183,10 +184,12 @@ func (rs *RequestServer) packetWorker(pktChan chan requestPacket) error {
...
 	} // all paths are absolute
 
-	cleaned_path := filepath.Clean(path)
+	cleaned_path := filepath.ToSlash(filepath.Clean(path))
```
davecheney: don't create a new variable, overwrite path so that the unclean version cannot be used by mistake.

request.go
```diff
@@ -52,16 +52,16 @@ func requestFromPacket(pkt hasPath) Request {
...
-		request.Target = filepath.Clean(p.Linkpath)
+		request.Target = filepath.ToSlash(filepath.Clean(p.Linkpath))
 	}
 	return request
 }
 
 // NewRequest creates a new Request object.
 func NewRequest(method, path string) Request {
-	request := Request{Method: method, Filepath: filepath.Clean(path)}
+	request := Request{Method: method, Filepath: filepath.ToSlash(filepath.Clean(path))}
```
davecheney: This filepath.ToSlash(filepath.Clean() logic appears too many times.

Please break it into a helper function and add a unit test for all the appropriate cases; "/", "//", "\\", "/a/", "\a/", etc

request.go
```diff
@@ -52,16 +52,16 @@ func requestFromPacket(pkt hasPath) Request {
...
 	request.packets = make(chan packet_data, sftpServerWorkerCount)
```
eikenb: This is the line that is in conflict below. The conflict is due to my making sftpServerWorkerCount public (capitalizing it). It is easy to resolve.

server_test.go
```diff
@@ -1,11 +1,156 @@
...
+
+	// spacer
+	got = result[42:43]
+	if ok, err := regexp.MatchString("^\\s$", got); !ok {
+		t.Errorf("runLs(%#v, *FileInfo): spacer 5 mismatch, expected whitespace, got: %#v, err: %#v", path,  got, err)
+	}
+
+	// mod time (len 12, e.g. Aug  9 19:46)
+	got = result[43:55]
+	layout := "Jan  2 15:04"
```
eikenb: Date stat display flips over to the year at a certain point (6 months I think). The LICENSE file on my system did this and this test failed. The format can be "Jan  2 15:04" or "Jan  2 2006" depending on this. One fix would be to check both and only error if neither parses.

sandreas:
All mentioned things should be fixed now...

eikenb:
@davecheney What do you think? There are a few extraneous things but they are all improvements, so I'm OK with them. I'm ready to merge it unless you have more you'd like to see.

PR #193, opened 2017-08-13
Move all methods to *Request

### LINKED ISSUE #192: Methods on sftp.Request mix pointer and value receivers
gopl recommends that all methods be on the same type of receiver; either pointer or value, and give the gynmastics that newRequest goes through to avoid copying a mutex and the state object each time Request is called as a value receiver or passed by value, it's clear the type of Requets' receiver should be *Request.

davecheney:
Fixes #192

### DIFF
--- a/request-example.go
+++ b/request-example.go
@@ -25,7 +25,7 @@ func InMemHandler() Handlers {
 }
 
 // Handlers
-func (fs *root) Fileread(r Request) (io.ReaderAt, error) {
+func (fs *root) Fileread(r *Request) (io.ReaderAt, error) {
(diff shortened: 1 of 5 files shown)


davecheney:
While trying to fix this bug I found this behaviour in 

https://github.com/pkg/sftp/blob/master/request.go#L308

update is altering the Method field of a _copy_ of Request (because the method is on a value receiver), this has no impact on the caller's value of Request, so the operation is a noop.

davecheney:
Oops, this is my mistake, update is one of the few methods on Request that mutates it's receiver.

PR #197, opened 2017-08-16
Fix #196; Increase offset correctly

### LINKED ISSUE #196: sftp.ListerAt processing seems broken
When calling ListAt(), the returned count is not used to increment the value for the next offset. Instead MaxFilelist seems to always be used.  This breaks ListerAt implementations that chose not to (or are unable to) return MaxFilelist entries each time.

This seems to be broken since 733115c415560ff2068431f5e9a540e387694fc8

mjgarton:
If a ListerAt implementation chooses to return fewer than MaxFileList
entries, the next offset it increased by MaxFileList in any case and so
we get unwanted results.  Instead, increase the offset by the correct
number as returned by ListAt.

### DIFF
--- a/request.go
+++ b/request.go
@@ -72,13 +72,18 @@ func NewRequest(method, path string) Request {
 	return request
 }
 
-// Returns current offset for file list, and sets next offset
-func (r Request) lsNext(offset int64) (current int64) {
+// Returns current offset for file list
+func (r Request) lsNext() int64 {
 	r.stateLock.RLock()
 	defer r.stateLock.RUnlock()
-	current = r.state.lsoffset
+	return r.state.lsoffset
+}
+
+// Increases next offset
+func (r Request) lsInc(offset int64) {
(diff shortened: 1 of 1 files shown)


eikenb:
I modeled ListAt() after ReadAt() is supposed to return an error if it returns fewer entries than the length of the slice passed in. As such, if this was followed, there would be no next call after it returns less than the set offset as it doesn't continue after an error. The documentation of the ListerAt interface closely follows the ReaderAt documentation in laying out this expectation.

The code doesn't actually check that the returned length (n) is the same the length of the slice, so what you say is possible. I looked at uses of ReadAt() in the standard lib to see what they did and I couldn't find one that checked this. They mostly just ignore the returned int (n).

Implementing your patch would make for a more forgiving API.. but at what cost. I'm not sure. Can anyone come up with a downside? I'm going to sleep on it and see if I can think of any reason not to merge this patch.

PR #198, opened 2017-08-18
Fixes to mkdir example

holtwilkins:
* Adjusting name of `ssh_fx_failure` variable to match go formatting
* Fixing logic error for absolute paths


eikenb:
Thanks for the contribution!

PR #200, opened 2017-09-05
Add PosixRename method which uses the posix-rename@openssh.com extension

bilbofishy:
to support actual rename() operations rather than the link() call
effectively mandated by the sftp v3/draft-2 requirement that targets
not be overwritten.

this is my first git/github operation so please be kind if I've prepared
it incorrectly or am not following the right procedure :)

thanks!

### DIFF
--- a/client.go
+++ b/client.go
@@ -575,6 +575,26 @@ func (c *Client) Rename(oldname, newname string) error {
 	}
 }
 
+// PosixRename renames a file using the posix-rename@openssh.com extension
+// which will replace newname if it already exists.
+func (c *Client) PosixRename(oldname, newname string) error {
+	id := c.nextID()
+	typ, data, err := c.sendPacket(sshFxpPosixRenamePacket{
+		ID:      id,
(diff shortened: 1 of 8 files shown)


eikenb:
Hey @bilbofishy, thanks for the PR. I've been super busy as of late but will try to get a chance to review this as soon as I can.

eikenb:
One thing of note. The built-in filesystem based SFTP server implementation uses os.Rename(), which overwrites even without the extension.

eikenb:
I'm not sure I like the server side extension support. My goal for the request based server was an interface based design that favored simplicity over completeness. If you are implementing the server you and you want to default to overwriting you are free to do so. You'd be out of spec, but it would work. 

Supporting the extension on the client side is a completely different matter and I'm fine with that addition. It would allow you to use the extension with servers that supported it.

I'd suggest splitting the client implementation out. I'll likely merge it right off as it looks good. We can then talk about the best way to handle the server side of things.

PR #206, opened 2017-12-01
Call stat before opening file for reading

wardn:
Calls .Stat() during .open() rather than during .WriteTo() and .Seek(). This is to address the an issue present in IBM instances with Extractability <= 1. The server doesn't allow Stat() to be called after a file descriptor has been opened.

For reference:
paramiko/paramiko#194
paramiko/paramiko#562
paramiko/paramiko#579

### DIFF
--- a/client.go
+++ b/client.go
@@ -399,6 +399,10 @@ func (c *Client) OpenFile(path string, f int) (*File, error) {
 }
 
 func (c *Client) open(path string, pflags uint32) (*File, error) {
+	fi, err := c.Stat(path)
+	if err != nil {
+		return nil, err
+	}
 	id := c.nextID()
 	typ, data, err := c.sendPacket(sshFxpOpenPacket{
 		ID:     id,
@@ -415,7 +419,7 @@ func (c *Client) open(path string, pflags uint32) (*File, error) {
 			return nil, &unexpectedIDErr{id, sid}
 		}
 		handle, _ := unmarshalString(data)
-		return &File{c: c, path: path, handle: handle}, nil
+		return &File{c: c, path: path, handle: handle, size: fi.Size()}, nil
 	case ssh_FXP_STATUS:
 		return nil, normaliseError(unmarshalStatus(id, data))
 	default:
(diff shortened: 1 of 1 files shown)


client.go
```diff
@@ -399,6 +399,10 @@ func (c *Client) OpenFile(path string, f int) (*File, error) {
 }
 
 func (c *Client) open(path string, pflags uint32) (*File, error) {
+	fi, err := c.Stat(path)
```
puellanivis: When calling `Create()` this function is called. Except it could be the case that the path does not already exist (we are after all, expecting to create a new file), which means that this c.Stat returns a “file does not exist” error, and so this function now returns nil, “does not exist” rather than actually creating a new file.

client.go
```diff
@@ -763,15 +768,11 @@ func (f *File) Read(b []byte) (int, error) {
...
-	fi, err := f.Stat()
-	if err != nil {
-		return 0, err
-	}
 	inFlight := 0
 	desiredInFlight := 1
 	offset := f.offset
 	writeOffset := offset
-	fileSize := uint64(fi.Size())
+	fileSize := uint64(f.size)
```
puellanivis: Using this cached size, we run into a problem. Follow these steps, create a new file (`f.size == 0`), now write out 512 bytes to that file. Seek to the begining, and now, make a call to WriteTo. This will assume the file is zero-length.

client.go
```diff
@@ -1046,11 +1047,7 @@ func (f *File) Seek(offset int64, whence int) (int64, error) {
 	case os.SEEK_CUR:
 		f.offset = uint64(int64(f.offset) + offset)
 	case os.SEEK_END:
-		fi, err := f.Stat()
-		if err != nil {
-			return int64(f.offset), err
-		}
-		f.offset = uint64(fi.Size() + offset)
+		f.offset = uint64(f.size + offset)
```
puellanivis: Using this cached size, we run into a problem. Follow these steps, create a new file (`f.size == 0`), now write out 512 bytes to that file. Seek to the end of the file. You are now at the beginning of the file, `(f.size + 0)` not the end `(f.Stat().Size() + 0)`.

Issue #207, opened 2017-12-06
REALPATH which returns /. causes Cyberduck to error


LordRalex:
We have noticed that Cyberduck versions 4-6 (as far as we've tested) do not properly handle the following case when the library returns the following path:

'/.'

Our use-case is in a jailing SFTP, which we use the request-server to handle processing the requests by the clients differently.

We found that when the client requests '.', the library returns "/." (which is not a true absolute path) as the return, then Cyberduck pukes, and we have no way to recover.

Naturally, this used to work before with this library, so something changed within the SFTP library to cause this.

This is the log we got from Cyberduck, pointing to the use of the Realpath being the cause: https://gist.github.com/LordRalex/61b71bfac43232be8ef89068d2495e00

I have found the change which introduced this bug:
https://github.com/pkg/sftp/commit/4d7bb970c49d8a86d12e2cd34952dc80dbc7cb0d#diff-412115c53c6c5fea203e8253f32d2645

In particular, this line of code was removed:
https://github.com/pkg/sftp/commit/4d7bb970c49d8a86d12e2cd34952dc80dbc7cb0d#diff-412115c53c6c5fea203e8253f32d2645L186-189

Before: /. is transformed to /
After: /. is returned to the client

This was a breaking change to the Cyberduck client, and so causes our application to therefore break.

Since /. is not a true absolute path, this code should be re-introduced.

puellanivis:
Yeah, it looks like a number of unexpected behavior changes were introduced in that commit. :(

A number of calls that were not wrangling relative paths into absolute paths ended up gaining this behavior.

LordRalex:
I know that at least, we had found that Cyberduck breaks because of it, and it took a while to figure out where it came from. 

Since for our mac users we recommend Cyberduck to, this is a critical issue for us. 

It seems like there might need a secondary "cleanToAbs" function which handles cases when it needs to be an Abs, for things like REALPATH. It can branch through cleanPath to clean the path, then resolve it to an Abs.

puellanivis:
Indeed, and _super_ helpful how much detail you gave in the issue to begin with. I was able to very quickly zoom in on the code changes and tease out the details.

Yes, I think a cleanToAbs function would be better, but it should resolve the path to an absolute path _first_, then it can branch through cleanPath. This is how the server.go FXP_REALPATH works: (with some significant paraphrasing) `path = filepath.Abs(path); path = filepath.Clean(path); path = filepath.ToSlash(path)`

[...]

PR #209, opened 2017-12-10
Fix parsing of '.' for Cyberduck client

LordRalex:
This PR resolves the issue identified here:
https://github.com/pkg/sftp/issues/207

Cyberduck requests the REALPATH of ".", which after filepath.Clean, returns as ".", which is invalid.

This fixes such logic to ensure that if the clean returns ".", it changes it to "/", following the old, correct logic.

### DIFF
--- a/request-server.go
+++ b/request-server.go
@@ -199,6 +199,12 @@ func cleanPacketPath(pkt *sshFxpRealpathPacket) responsePacket {
 
 func cleanPath(path string) string {
 	cleanSlashPath := filepath.ToSlash(filepath.Clean(path))
+	if cleanSlashPath == "." || cleanSlashPath == ".." {
+		//At this stage, the path is not valid
+		//This occurs when there is root .. or . in the path, which clean did not remove. This will point to root.
+		return "/"
+	}
+
 	if !strings.HasPrefix(cleanSlashPath, "/") {
 		return "/" + cleanSlashPath
 	}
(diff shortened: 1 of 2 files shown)


request-server.go
```diff
@@ -199,6 +199,12 @@ func cleanPacketPath(pkt *sshFxpRealpathPacket) responsePacket {
 
 func cleanPath(path string) string {
```
puellanivis: `cleanPath` is called for more than just SSH_FXP_REALPATH. In fact, it looks like commit 4d7bb970c49d8a86d12e2cd34952dc80dbc7cb0d made a logical mistake when they removed the duplicate codes of `filepath.ToSlash(filepath.Clean(path))` in favor of `cleanPath(path)`, as the behavior of those calls was explicitly _not_ supposed to append a '/' to the front of the paths. (Only RequestServer.SSH_FXP_REALPATH was supposed to do this.)

I am concerned that this code would just further exacerbate the problem rather than provide a root-cause fix. When `cleanPath` was only working on an FXP_REALPATH call, its assumptions on constructing an absolute path were “acceptable”, but then the code got pushed around and used elsewhere, where the expanded behavior of turning relative paths into absolute paths did not exist before.

Properly, all of the FXP_REALPATH code paths should be calling cleanPath only _after_ it turns it from relative to absolute, and this functionality should not be done itself in the cleanPath function at all. (Even if it looks like in the RequestServer this amounts simply to “prepend a slash to the path".)

LordRalex:
Closing PR as the solution proposed will not solve all issues introduced by the commit, and more work is needed to devise a good solution.

PR #213, opened 2017-12-20
Add PosixRename method which uses the posix-rename@openssh.com extension (without server support)

vansante:
Hi,

As I kind of need support for this extension in the sftp client, I took the liberty to create a new PR without the server side code. Old PR can be found here: https://github.com/pkg/sftp/pull/200

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,11 @@
 language: go
 go_import_path: github.com/pkg/sftp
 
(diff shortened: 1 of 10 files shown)


eikenb:
Looks good to me. Just the client code from #200.

eikenb:
I just pushed a new tag/release which contains this code: 1.1.0

PR #216, opened 2017-12-24
respect create flag in open packet

### LINKED ISSUE #214: request-server doesn't handle uploads of empty files
Hi!

I figured out that request-server has a problem with uploading empty files. Uploading of an empty file finishes without errors but no file has created.
It happens because the file supposed to be created after the first SSH2_FXP_WRITE packet but SFTP client just sends SSH2_FXP_OPEN and than SSH2_FXP_CLOSE. 
Possible solution:

1. Extend isOpener interface by methods isOpenedForWrite() bool and isOpenedForRead()  bool.
2. In packetWorker() in case of isOpener packet if the file is opening for write, create empty sshFxpWritePacket structure and process it.  Dirty and hacky example:

		case isOpener:
			handle := rs.nextRequest(requestFromPacket(pkt))
			if pkt.isOpenedForWrite() {
				request, _ := rs.getRequest(handle)
				wpkt := &sshFxpWritePacket{}
				uerr := request.updateMethod(wpkt)
				if uerr != nil {
					rpkt = statusFromError(pkt, syscall.EBADF)
				} else {
					rpkt = request.call(rs.Handlers, wpkt)
				}
			}
			rpkt = sshFxpHandlePacket{pkt.id(), handle}

In this case will be invoked a write of empty buffer and the file will be created. In case of already existent file the file content will no be overwritten because we use empty byte buffer.

It will also cover scenario when some SFTP client creates a  file using SSH2_FXP_OPEN and then sends  immediately SSH2_FXP_STAT packet just to check file existents. 

P.S. Adding methods isOpenedForWrite() and isOpenedForRead() to the isOpener interface doesn't cover situation when SFTP client opens a file  with O_TRUNC flag.

eikenb:
When the SSH_FXP_OPEN packet has the SSH_FXF_CREAT pflag set the file
should get created even if no data is sent. We do this by having the
Open method call through to the FilePut Handler with empty packet data.

Fixes #214

### DIFF
--- a/request-server.go
+++ b/request-server.go
@@ -152,7 +152,12 @@ func (rs *RequestServer) packetWorker(pktChan chan requestPacket) error {
 		case *sshFxpRealpathPacket:
 			rpkt = cleanPacketPath(pkt)
 		case isOpener:
-			handle := rs.nextRequest(requestFromPacket(pkt))
+			request := requestFromPacket(pkt)
+			handle := rs.nextRequest(request)
+			p, ok := pkt.(*sshFxpOpenPacket)
+			if ok && p.hasPflags(ssh_FXF_CREAT) {
+				request.call(rs.Handlers, pkt)
+			}
 			rpkt = sshFxpHandlePacket{pkt.id(), handle}
 		case *sshFxpFstatPacket:
 			handle := pkt.getHandle()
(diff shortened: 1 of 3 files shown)


eikenb:
@puellanivis Thanks for giving it a look. I need to add a test and once that is done I'll merge it.

Issue #212, opened 2017-12-15
"File does not exist" every time when using OpenFile


ericcancil:
As i said in the title, I get "File does not exist" every time when using OpenFile.  I can successfully call Open on the same file without any error, and am wondering what is causing this issue.

Thank you

eikenb:
Both OpenFile() and Open() use the exact same underlying call. The only difference is that Open() has the readonly flag hardcoded in where OpenFile() requires you pass the flags. So my best guess with the information available is that you are calling it with a bad flag. You should be passing in flags from the os package. Eg. `client.OpenFile("/file/path",  os.O_RDONLY)` should behave the same thing as `client.Open("/file/path")`.

eikenb:
@ericcancil Just pinging you in case you didn't see my previous post. I'm not sure if you get emailed notifications of my reply w/ vs w/o mentioning you.

ericcancil:
@eikenb 
I'm sorry, these were going to spam for some reason.  Thanks for your reply.

I'm writing the file as follows

```
func checkForLogFile(client *sftp.Client) {
	if _, err := client.Stat(logFile); os.IsNotExist(err) {
		var file, createErr = client.Create(logFile)
		if createErr != nil {
			panic(createErr)
		} else {
			log.Println("log file created!")
			defer file.Close()
		}
	} else {
		if err != nil {
			panic(err)
		} else {
			log.Println("Log file already exists")
		}
	}
}
```

Then in the future I am trying to write to it as such 

```
func writeLogFile(fileInfo FileAndPath, log FileLog) error {
	if f, err := fileInfo.Client.OpenFile(logFile, os.O_RDWR|os.O_CREATE|os.O_TRUNC); err != nil {
		return err
	} else {
		defer f.Close()

		if _, err := f.Write([]byte(log.String())); err != nil {
			return err
		}
	}
	return nil
}
```

I'm using ```os.O_RDWR|os.O_CREATE|os.O_TRUNC``` as that is what the create call uses.  Yet I am still getting file does not exist every time.

Any ideas?

eikenb:
Which SFTP server is your client connecting to? Testing your code with my local OpenSSH/SFTP server and it works.

ericcancil:
When I test locally with docker it works as well.  When i connect to any remote sftp server 
(which i have no insight into the setup) I am getting these issues.

You don't see anything funny as far as my code?  Thanks for answering back so quickly.  Really appreciate it.

eikenb:
[...]

Issue #185, opened 2017-07-13
failed to send packet header: EOF on file with size >= 300 KB


vvsh:
Hi All,

I am trying to upload local files to SFTP.

     
`   client = openSSHConnection(config)
    ftpClient, err := sftp.NewClient(client, sftp.MaxPacket(10240000))
    if err != nil {
        log.Fatal(err)
    }
    defer ftpClient.Close()

    for _, jar := range newJars {
        w, err := ftpClient.Create(config["directory"] + "/" + jar.Name)
        if err != nil {
            log.Println("create error: ", jar.Name)
            log.Fatal(err)
        }
        defer w.Close()

        f, err := os.Open(jar.Path)
        if err != nil {
            log.Fatal(err)
        }
        defer f.Close()

        _, uploadError := io.Copy(w, f)
        if uploadError != nil {
            log.Println("copy error: ", jar.Name)
            log.Fatal(uploadError)
        }

        log.Print(jar.Name + " uploaded")
    }

    func openSSHConnection(config map[string]string) *ssh.Client {
    hostname := config["host"]
    port := config["port"]

    sshConfig := &ssh.ClientConfig{
        User: config["user"],
        Auth: []ssh.AuthMethod{
            ssh.Password(config["password"]),
        },
        HostKeyCallback: func(hostname string, remote net.Addr, key ssh.PublicKey) error {
            return nil
        },
        Timeout: time.Duration(90) * time.Second,
    }

    connection, err := ssh.Dial("tcp", hostname+":"+port, sshConfig)
    if err != nil {
        log.Fatal("Failed to dial: ", err.Error())
    }
    return connection
}`

But if a file is larger than ~ 300 KB it does not upload.

I receive the following in the log:

2017/07/13 18:46:07 create error 123.jar
2017/07/13 18:46:07 failed to send packet header: EOF

and file size is 0 on the server.

What am I doing wrong?

Thanks,
Uladzimir

eikenb:
What server are you uploading to? Openssh's sftp server?

eikenb:
@vvsh  What server are you uploading to? Openssh's sftp server?

vvsh:
@eikenb, OpenSSH_6.6.1p1, OpenSSL 1.0.1e-fips 11 Feb 2013

eikenb:
I cannot reproduce any issue like you describe. I can upload anywhere from empty files to multi-gigabyte files.

[...]

PR #220, opened 2018-01-15
Expanded limits to the MaxPacket option.

CodyDWJones:
According to the RFC servers must support packets 32 768 bytes or
smaller.  This might be necessary to prevent "failed to send packet
header: EOF" errors.

The artificial maximum was removed because it's possible the library
will interact with a server that has a higher limit than OpenSSH's
SFTP server.

### DIFF
--- a/client.go
+++ b/client.go
@@ -24,27 +24,55 @@ var InternalInconsistency = errors.New("internal inconsistency")
 // A ClientOption is a function which applies configuration to a Client.
 type ClientOption func(*Client) error
 
-// This is based on Openssh's max accepted size of 1<<18 - overhead
-const maxMaxPacket = (1 << 18) - 1024
-
-// MaxPacket sets the maximum size of the payload. The size param must be
-// between 32768 (1<<15) and 261120 ((1 << 18) - 1024). The minimum size is
-// given by the RFC, while the maximum size is a de-facto standard based on
-// Openssh's SFTP server which won't accept packets much larger than that.
+// MaxPacketChecked sets the maximum size of the payload, measured in bytes.
+// This option only accepts sizes servers should support, ie. <= 32768 bytes.
 //
-// Note if you aren't using Openssh's sftp server and get the error "failed to
-// send packet header: EOF" when copying a large file try lowering this number.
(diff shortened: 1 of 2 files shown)


eikenb:
I'm going to have to think about this though. That code was added due to multiple people trying to use larger packet sizes with OpenSSH and having issues (eg. #185). Why you want this beyond the fact that the spec allows for it? Does using a packet size larger than the current (261120) improve throughput?

If we do change it I will want to come up with a more explicit warning about the OpenSSH issue. It was pretty obscure and took me a quite a bit of time to track down. I'd hate to inflict that on a future maintainer.

Thanks.

CodyDWJones:
I'm a proponent of letting developers choose any value so long as it satisfies the RFC. That said, I took care to read the MaxPacket godoc before using it, because I am trying to diagnose an intermittent case of #212. It's possible someone else won't pay attention, use a value larger than 32KB, and after it starts sporadically failing in the field, waste time repeating the investigation you did for #185.

The language in the godoc could be changed to draw more attention to this problem:

"WARNING: values larger than 32768 might not be supported by all servers. Each server can enforce its own limitation. For example, OpenSSH allows payloads up to 261120 bytes ((1 << 18) - 1024). Only use a payload larger than 32768 if your application always connects to the same server or after sufficiently broad testing."

Another option is to use type names as documentation: remove MaxPacket and replace it with two mutually exclusive ClientOptions called, say, `ShortPayload ` and `ServerKnownToSupportPayload`. One would accept any value < 32768 and the other any value > 32768. Not entirely happy with the names, but you get the idea.

Or, I'm happy to restore the OpenSSH maximum check, if you like.

eikenb:
I like the idea of breaking this up into a couple options. One option with the checks and one without. Naming them in such a way as to be more self documenting. Maybe something like... MaxPacketSafe/MaxPacketUnsafe or MaxPacketChecked/MaxPacketUnchecked.

MaxPacket would have to stay for compatibility reasons, but it could default to either one.

CodyDWJones:
OK, great. I'll see about making this change and updating my PR.

CodyDWJones:
PR updated.

PR #224, opened 2018-01-22
Add context to request lifecycle

### LINKED ISSUE #222: context.Context for request lifecycle?
I have some async calls that are made during various handlers, and at the moment I just have to throw a timeout context around them. It would be nice if there was a cancellable context attached to the handlers that is cancelled once the request closes. Looking through the code I think it could be done, but would need to add handlers that accept a context as the first argument. Is this something that makes sense for SFTP requests?

crazed:
Fixes #222

### DIFF
--- a/request-server.go
+++ b/request-server.go
@@ -59,8 +59,8 @@ func NewRequestServer(rwc io.ReadWriteCloser, h Handlers) *RequestServer {
 
 // New Open packet/Request
 func (rs *RequestServer) nextRequest(r *Request) string {
-	rs.openRequestLock.RLock()
-	defer rs.openRequestLock.RUnlock()
+	rs.openRequestLock.Lock()
+	defer rs.openRequestLock.Unlock()
 	rs.handleCount++
 	handle := strconv.Itoa(rs.handleCount)
 
--- a/request-server.go
+++ b/request-server.go
@@ -59,8 +59,8 @@ func NewRequestServer(rwc io.ReadWriteCloser, h Handlers) *RequestServer {
 
 // New Open packet/Request
 func (rs *RequestServer) nextRequest(r *Request) string {
-	rs.openRequestLock.Lock()
-	defer rs.openRequestLock.Unlock()
+	rs.openRequestLock.RLock()
+	defer rs.openRequestLock.RUnlock()
 	rs.handleCount++
 	handle := strconv.Itoa(rs.handleCount)
 
@@ -73,7 +73,7 @@ func (rs *RequestServer) nextRequest(r *Request) string {
 	return handle
 }
 
-// Returns openRequest from openRequests, bool is false if it is missing
+// Returns Request from openRequests, bool is false if it is missing
 // If the method is different, save/return a new Request w/ that Method.
 //
 // The Requests in openRequests work essentially as open file descriptors that
@@ -86,7 +86,8 @@ func (rs *RequestServer) getRequest(handle, method string) (*Request, bool) {
 	rs.openRequestLock.RUnlock()
 	if !ok {
 		return nil, ok
-	} else if or.r.Method == method {
+	}
+	if or.r.Method == method {
 		return or.r, ok
 	}
 	// if we make it here we need to replace the request
@@ -95,7 +96,8 @@ func (rs *RequestServer) getRequest(handle, method string) (*Request, bool) {
 	or, ok = rs.openRequests[handle]
 	if !ok {
 		return nil, ok
-	} else if or.r.Method == method { // re-check needed b/c lock race
+	}
+	if or.r.Method == method { // re-check needed b/c lock race
 		return or.r, ok
 	}
 	or = &openRequest{
--- a/request-server.go
+++ b/request-server.go
@@ -1,6 +1,7 @@
 package sftp
 
 import (
+	"context"
 	"encoding"
 	"io"
 	"path"
@@ -24,12 +25,17 @@ type Handlers struct {
 	FileList FileLister
 }
 
+type openRequest struct {
+	r      *Request
+	cancel context.CancelFunc
+}
+
 // RequestServer abstracts the sftp protocol with an http request-like protocol
 type RequestServer struct {
 	*serverConn
 	Handlers        Handlers
 	pktMgr          *packetManager
-	openRequests    map[string]*Request
+	openRequests    map[string]*openRequest
 	openRequestLock sync.RWMutex
 	handleCount     int
 }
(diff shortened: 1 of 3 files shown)


request-server.go
```diff
@@ -24,12 +25,17 @@ type Handlers struct {
 	FileList FileLister
 }
 
+type openRequest struct {
```
puellanivis: Why a new type and not just use Request?

request-server.go
```diff
@@ -57,11 +63,17 @@ func (rs *RequestServer) nextRequest(r *Request) string {
...
+	r = r.WithContext(ctx)
+	rs.openRequests[handle] = &openRequest{
+		r:      r,
+		cancel: cancel,
+	}
 	return handle
 }
 
-// Returns Request from openRequests, bool is false if it is missing
+// Returns openRequest from openRequests, bool is false if it is missing
```
puellanivis: The code is still returning the Request, not an openRequest.

request-server.go
```diff
@@ -70,29 +82,37 @@ func (rs *RequestServer) nextRequest(r *Request) string {
...
 func (rs *RequestServer) getRequest(handle, method string) (*Request, bool) {
 	rs.openRequestLock.RLock()
-	r, ok := rs.openRequests[handle]
+	or, ok := rs.openRequests[handle]
 	rs.openRequestLock.RUnlock()
-	if !ok || r.Method == method {
-		return r, ok
+	if !ok {
+		return nil, ok
+	} else if or.r.Method == method {
```
puellanivis: https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

request-server.go
```diff
@@ -70,29 +82,37 @@ func (rs *RequestServer) nextRequest(r *Request) string {
...
 	// if we make it here we need to replace the request
 	rs.openRequestLock.Lock()
 	defer rs.openRequestLock.Unlock()
-	r, ok = rs.openRequests[handle]
-	if !ok || r.Method == method { // re-check needed b/c lock race
-		return r, ok
+	or, ok = rs.openRequests[handle]
+	if !ok {
+		return nil, ok
+	} else if or.r.Method == method { // re-check needed b/c lock race
```
puellanivis: https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow (same as above)

request.go
```diff
@@ -58,6 +60,41 @@ func NewRequest(method, path string) *Request {
...
+// The returned context is always non-nil; it defaults to the
+// background context.
+//
+// For incoming server requests, the context is canceled when the
+// client's connection closes.
+func (r *Request) Context() context.Context {
+	if r.ctx != nil {
+		return r.ctx
+	}
+	return context.Background()
```
puellanivis: In general, one should never be returning the background context.

If no context has been set, then this should return nil, and this should indicate that they should not attempt to use ctx.Done().

Imagine here a program failed to set a context, but then used `<-ctx.Done()` to wait for the context to end…

request.go
```diff
@@ -58,6 +60,41 @@ func NewRequest(method, path string) *Request {
...
+}
+
+// WithContext returns a copy of r with its context changed to ctx.
+// The provided ctx must be non-nil.
+func (r *Request) WithContext(ctx context.Context) *Request {
+	if ctx == nil {
+		panic("nil context")
+	}
+	r.stateLock.Lock()
```
puellanivis: You are not mutating state here, so better would be to use RLock()

request-server.go
```diff
@@ -57,11 +63,17 @@ func (rs *RequestServer) nextRequest(r *Request) string {
 	defer rs.openRequestLock.Unlock()
 	rs.handleCount++
 	handle := strconv.Itoa(rs.handleCount)
-	rs.openRequests[handle] = r
+
+	ctx, cancel := context.WithCancel(context.Background())
+	r = r.WithContext(ctx)
```
puellanivis: Shouldn’t this be the responsibility of the person making the request?

What if we don’t want every request to have a context?

Default behavior should be that contexts are not added in, if they’re not explicitly given by the user.

request.go
crazed: Hmm, I was just copying from net/http/request.go:

https://golang.org/src/net/http/request.go?s=11390:11433#L306

I think in general if you have a nil context it's the same as having context.Background() since it will never be cancelled. I'm willing to change this but figured we were mimicking http.Request so might want to keep the same properties around?

request-server.go
crazed: I don't have a problem adding the cancel function to the Request itself but it seemed a little out of place since the sftp request handler code is keeping track of the open requests. I figured it makes sense to separate tracking of requests from the details about a specific request.

request-server.go
crazed: This is where I somewhat disagree, I'd say it makes sense to attach a cancellable context at the layer that is creating these requests and knows when the requests are done and thus can cancel the attached context. For example in the net/http Server implementation:

https://github.com/golang/go/blob/release-branch.go1.9/src/net/http/server.go#L1730

They setup a cancelFunc and context at the time of accepting the TCP connection and cancel it once that connection has completed.

request-server.go
puellanivis: Hm… I suppose my “better” idea would be that somehow the cancel function should then be contained through a closure or something… but with the design as is, I am not sure how that would work, so… 🤷‍♀️

request-server.go
puellanivis: Hm… but you notice the design there is to build a context, with a cancelFunc, then defer the cancel… here, it seems that we stuff the context, and cancel into a structure, and who knows where it gets taken care of.

request.go
puellanivis: Ah, didn’t know about that code. Looks good in that case, but… eh… just hangups from my days back in Google where the guidelines basically said, “never use context.Background(), if you’re using context.Background, you’re doing something wrong. (some exceptions apply)”

But then, this is building things from the bottom up, where the exceptions almost certainly apply…

request-server.go
crazed: That's a good point, but I'm not sure if we can set up a similar flow given the way packetWorker switches off specific sftp packet types. 

Here I'm sort of relying on nextRequest getting called first and eventually a closeRequest, but I'm unsure if the sshFxpClosePacket is guaranteed to be received even if there's a non-clean ending to the underlying connection.

request-server.go
eikenb: I'm also not sure about the separate struct just for the request + cancel func, but I get your point. I'll leave it for now so I can see it in context and work with it some.

eikenb:
Thanks for the PR. I'm currently reading/thinking through it.

eikenb:
One of my initial concerns is that you are limiting the Requests that get a Context to those that are guarded by Open/Close packets. This leaves the single packet Requests (eg. Stat, Readlink, Rmdir, etc) without a Context and I think they should all have a Context for consistency.

I'm going to think about this for a bit. I'm not 100% on what I think would be the best approach yet.

eikenb:
Closed in favor or #225

PR #225, opened 2018-01-31
Add context to request lifecycle (revised)

eikenb:
New PR fixing #222, based on the other by @crazed but reworked a bit.

### DIFF
--- a/request.go
+++ b/request.go
@@ -40,9 +40,11 @@ type state struct {
 }
 
 // New Request initialized based on packet data
-func requestFromPacket(pkt hasPath) *Request {
+func requestFromPacket(ctx context.Context, pkt hasPath) *Request {
 	method := requestMethod(pkt)
 	request := NewRequest(method, pkt.getPath())
+	request.ctx, request.cancelCtx = context.WithCancel(ctx)
+
 	switch p := pkt.(type) {
 	case *sshFxpOpenPacket:
 		request.Flags = p.Pflags
@@ -59,9 +61,7 @@ func requestFromPacket(pkt hasPath) *Request {
 
 // NewRequest creates a new Request object.
 func NewRequest(method, path string) *Request {
-	ctx, cancel := context.WithCancel(context.Background())
-	return &Request{Method: method, Filepath: cleanPath(path),
-		ctx: ctx, cancelCtx: cancel}
+	return &Request{Method: method, Filepath: cleanPath(path)}
 }
 
 // Context returns the request's context. To change the context,
--- a/request.go
+++ b/request.go
@@ -1,6 +1,7 @@
 package sftp
 
 import (
+	"context"
 	"io"
 	"os"
 	"path"
@@ -26,6 +27,9 @@ type Request struct {
 	// reader/writer/readdir from handlers
 	stateLock sync.RWMutex
 	state     state
+	// context lasts duration of request
+	ctx       context.Context
+	cancelCtx context.CancelFunc
 }
 
 type state struct {
@@ -55,7 +59,44 @@ func requestFromPacket(pkt hasPath) *Request {
 
 // NewRequest creates a new Request object.
 func NewRequest(method, path string) *Request {
-	return &Request{Method: method, Filepath: cleanPath(path)}
+	ctx, cancel := context.WithCancel(context.Background())
+	return &Request{Method: method, Filepath: cleanPath(path),
+		ctx: ctx, cancelCtx: cancel}
+}
+
+// Context returns the request's context. To change the context,
+// use WithContext.
+//
+// The returned context is always non-nil; it defaults to the
+// background context.
+//
+// For incoming server requests, the context is canceled when the
+// request is complete or the client's connection closes.
+func (r *Request) Context() context.Context {
+	if r.ctx != nil {
+		return r.ctx
+	}
+	return context.Background()
+}
+
+// WithContext returns a copy of r with its context changed to ctx.
+// The provided ctx must be non-nil.
+func (r *Request) WithContext(ctx context.Context) *Request {
+	if ctx == nil {
+		panic("nil context")
+	}
+	r.stateLock.Lock()
+	defer r.stateLock.Unlock()
+	r2 := &Request{
+		Method:   r.Method,
+		Filepath: r.Filepath,
+		Target:   r.Target,
+		Flags:    r.Flags,
+		Attrs:    r.Attrs,
+		state:    r.state,
+		ctx:      ctx,
+	}
+	return r2
 }
 
 // Returns current offset for file list
@@ -109,6 +150,9 @@ func (r *Request) getLister() ListerAt {
 
 // Close reader/writer if possible
 func (r *Request) close() error {
+	if r.cancelCtx != nil {
+		r.cancelCtx()
+	}
 	rd := r.getReader()
 	if c, ok := rd.(io.Closer); ok {
 		return c.Close()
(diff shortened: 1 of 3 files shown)


request.go
```diff
@@ -55,7 +59,44 @@ func requestFromPacket(pkt hasPath) *Request {
...
+
+// WithContext returns a copy of r with its context changed to ctx.
+// The provided ctx must be non-nil.
+func (r *Request) WithContext(ctx context.Context) *Request {
+	if ctx == nil {
+		panic("nil context")
+	}
+	r.stateLock.Lock()
+	defer r.stateLock.Unlock()
+	r2 := &Request{
```
puellanivis: This sort of explicit copying of each field is error-prone, since if one adds a new field it would need to be added here. This is why the http.Request.WithContext method uses `*r2 = *r` shallow copy. But in our case, we have both: a Mutex that needs to not be copied, and a cancelFunc that we would need to clear (with the complications of “how to ensure it still gets called if this context is replaced”.)

request.go
```diff
@@ -55,7 +59,44 @@ func requestFromPacket(pkt hasPath) *Request {
 
 // NewRequest creates a new Request object.
 func NewRequest(method, path string) *Request {
-	return &Request{Method: method, Filepath: cleanPath(path)}
+	ctx, cancel := context.WithCancel(context.Background())
```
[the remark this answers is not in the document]
eikenb: Maybe just a private newRequestWithContext() and NewRequest reverts back to its normal state. So if you call NewRequest, you would need to use WithContext if you wanted it with a context.

This way the initial request gets the context and can handle cleaning it up. If you re-wrap it with a new WithContext you'll get a copy and the original is still there and closed.

request.go
eikenb: I have this changed already locally. After reading the http request/context code I thought that method was much cleaner. I have a couple queued up commits that clean up the lock usage a bit, convert the lock to a pointer (which you can copy then replace)  and uses the *r2 = *r for copying.

eikenb: Oh. And the cancel would be set to nil post copy. The original request would still be tracked and still be closed in the server code. Calling WithContext doesn't replace the one in memory (scoped in packetworker) nor does it replace the one tracked in RequestServer.openRequests. Those would still be around and get closed properly.

request.go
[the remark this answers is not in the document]
eikenb: > I think the http.Request design gets around all of this by not automatically setting up a context that will be closed.

They do auto-create/set one in the server code where they need it. I think they did it this way due to the 1.0 guarantee, which would prevent breaking external servers that use http.Request. We don't have the 1.0 guarantee and the request server framework doesn't try to provide standalone components with guaranteed stability, only the guarantee's of semantic versioning.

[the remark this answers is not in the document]
eikenb: Regarding the example code snippets, assuming the newRequestsWithContext change, they would all be handled fine. The server coded would track and close the original request, thus calling its cancelFunc.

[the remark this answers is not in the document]
puellanivis: I think the interest here is that a client calling NewRequest will not be queuing up any way to ensure the CancelFunc is called.

The http.Request code ignores making contexts until WithContext, leaving it a matter for the client to arrange and ensure that a CancelFunc is called if created.

request.go
puellanivis: 👍

request.go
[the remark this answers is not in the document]
eikenb: The http.Request code doesn't make a context, but the http.server code does. The sftp lib isn't trying to provide a stand alone sftp.Request object, it is using it in concert with the server code.

I'll push up my changes that move the context creation out of NewRequest up as the new baseline for the discussion.

request.go
```diff
@@ -59,9 +59,14 @@ func requestFromPacket(pkt hasPath) *Request {
...
 // NewRequest creates a new Request object.
 func NewRequest(method, path string) *Request {
-	ctx, cancel := context.WithCancel(context.Background())
-	return &Request{Method: method, Filepath: cleanPath(path),
-		ctx: ctx, cancelCtx: cancel}
+	return &Request{Method: method, Filepath: cleanPath(path)}
+}
+
+// Used internally to create request with default context
+func newRequestWithContext(method, path string) *Request {
```
puellanivis: A naming point here. I would expect a `newRequestWithContext` function to have a context.Context as a parameter, which gets folded into the `&Request{}`. In this case, I think `newContextRequest` would make more sense.

But then, the English-usage here is ambiguous anyways, so…

puellanivis:
I think additionally, if we’re going to make the auto-Context creation for the server in the unexposed newRequest function, let’s add a caller to WithContext doesn’t try and replace the Context if there is a CancelFunc?

I think it would be a fundamental error in usage of the library for a request handler to try and manipulate the context of a Request anyways…

request.go
eikenb: I can see how the WithContext part of the newRequestWithContext does seem to imply passing in a context as that is what WithContext does... So sure, this seems like a reasonable change.

eikenb:
> let’s add a caller to WithContext doesn’t try and replace the Context if there is a CancelFunc?

Are you saying it should not set the passed Context if the CancelFunc was set? If so, why? WithContext returns a copy of the Request, so the original is left alone and would always have its CancelFunc called.

puellanivis:
More just a thought… “hey, there is no real reason why server code should have to modify the context in the request anyways…”

More just a thought than anything, rather than any sort of “recommendation”.

request.go
```diff
@@ -38,7 +42,7 @@ type state struct {
 // New Request initialized based on packet data
 func requestFromPacket(pkt hasPath) *Request {
 	method := requestMethod(pkt)
-	request := NewRequest(method, pkt.getPath())
+	request := newRequestWithContext(method, pkt.getPath())
```
ernesto-jimenez: Why not leverage the cancelation of a context tree doing the following:

```go
func requestFromPacket(ctx context.Context, pkt hasPath) *Request {
  	method := requestMethod(pkt)
	request := NewRequest(method, pkt.getPath())
	request.ctx, request.cancelCtx = context.WithCancel(ctx)
	// ...
```

Then, change the signature of `packetWorker`

```go
func (rs *RequestServer) packetWorker(ctx context.Context, pktChan chan requestPacket) error {
	// Use ctx when calling requestFromPacket
	// ...
}
```

And finally, `Serve`:

```go
 func (rs *RequestServer) Serve() error {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 	var wg sync.WaitGroup
 	runWorker := func(ch requestChan) {
 		wg.Add(1)
 		go func() {
 			defer wg.Done()
 			if err := rs.packetWorker(ctx, ch); err != nil {
 				rs.conn.Close() // shuts down recvPacket
 			}
 		}()
 	}
 	// ...
}
```

eikenb: @ernesto-jimenez Thanks. That sounds like a great idea. Let me make those changes and see them in context, if they look as good as they sound I'll be pushing up an amended PR soon.

eikenb:
Just pushed the great changes suggested by @ernesto-jimenez (thanks!). I think this is the one to merge. If interested please have a look and let me know if you see any issues.

Thanks again everyone for helping out with this ticket/feature.

eikenb:
Note for @puellanivis... once this is merged I'll be reworking things a bit to clean up the state/lock/copy issue. I have that code written, it just needs to be merged with the final version of this PR.

PR #226, opened 2018-02-02
Remove gccgo special cases

tamird:
This was fixed in a different way in 4d0e916.

@davecheney

### DIFF
--- a/server_statvfs_impl.go
+++ b/server_statvfs_impl.go
@@ -1,4 +1,4 @@
-// +build darwin linux,!gccgo
+// +build darwin linux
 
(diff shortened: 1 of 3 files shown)


eikenb:
Isn't server_statvfs_stubs.go still needed for builds not on Darwin or Linux (eg. Windows)?

tamird:
Good point! Fixed.

eikenb:
I'm going to let this sit for a little bit before merging in case @davecheney wants to chime in.

eikenb:
Long enough. Merging. Thanks @tamird.

Issue #230, opened 2018-02-22
RequestServer should allow for externalization of realpath method, allowing for better current working directory support.


robermorales:
There is a minimal issue in the current implementation of `cleanPath` in `request-server.go@223`:

```golang
// Makes sure we have a clean POSIX (/) absolute path to work with
func cleanPath(p string) string {
	p = filepath.ToSlash(p)
	if !filepath.IsAbs(p) {
		p = "/" + p
	}
	return path.Clean(p)
}
```

It does the assumption of current working directory to be "/", since "/" got prepended when path is relative.

SFTP RFC supports the idea of remote working directory, and most clients send a request to clarify which it is, just after connecting. Actually, the client included in this package also provides the functionality of `Getwd` and implement it sending a request of `SSH_FXP_REALPATH` with the path `"."`

The assumption made in `cleanPath` is clearly not what most of servers implement it, so we are forcing every RequestServer to start at "/", without providing a way to change it.

Here, the same file `request-server.go`, line `165`:

```golang
		case *sshFxpRealpathPacket:
			rpkt = cleanPacketPath(pkt)
		case *sshFxpOpendirPacket:
			request := requestFromPacket(ctx, pkt)
			handle := rs.nextRequest(request)
			rpkt = sshFxpHandlePacket{pkt.id(), handle}
```

We can see that the request for the realpath is not proxied to the RequestServer `rs`.

I propose redirect this request to the `rs`, to the `Filelist` handler, and allow the creator of every RequestServer to specify how to manage this.

If the issue got some attention and seems reasonable to be accepted, I can implement it and send a PR, with tests, modifying the documentation present at `request-readme.md` and the example at `request-example.go`.

Current implementation proposal will not break current 3rd party code, since we are not adding more methods to the interface FileLister.

Related issues refer to the ability to expose this option in the file based server, but the current issue is related to the RequestServer, to allow the functionality to be implemented outside, in the handlers.

https://github.com/pkg/sftp/issues/82
https://github.com/pkg/sftp/pull/92

So, should I implement the behaviour and send the PR?

eikenb:
Thanks for contributing!

The current code does not allow for an starting remote working directory as a way simplify things. Now that the code has mostly stabilized I am more than willing to entertain adding it. I'd need to see the code before saying for sure it'd be accepted, but I'm willing to review and (possibly) merge a PR.

TLDR; Thanks and please submit the PR.

PR #231, opened 2018-03-04
forward realpath requests to request handlers, fixes pkg/sftp#230

robermorales:


robermorales:
No idea about the error that travis reported:

```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
```

but seems nothing to do with the new code. Any idea?

eikenb:
That was just travis having issues. Re-ran the test and it passed.

request-server.go
```diff
@@ -161,7 +161,9 @@ func (rs *RequestServer) packetWorker(
 			handle := pkt.getHandle()
 			rpkt = statusFromError(pkt, rs.closeRequest(handle))
 		case *sshFxpRealpathPacket:
```
eikenb: This case is not correct and ultimately not needed. The sshFxpRealpathPacket was special cased as it was only cleaning the path and returning it, without calling the request handler code. Your change makes it so that it wants to be passed to the request handler code, to do that you'd just delete this case. As is it returns the sshFxpHandlePacket instead of the sshFxpNamePacket, which is wrong. Anyways, just deleting this case fixes that.

request-example.go
```diff
@@ -151,6 +151,16 @@ func (fs *root) Filelist(r *Request) (ListerAt, error) {
 			return nil, err
 		}
 		return listerat([]os.FileInfo{file}), nil
+	case "Realpath":
+		fp := r.Filepath
+		if !filepath.IsAbs(fp){
+			fp = "/" + fp
+		}
```
eikenb: This needs a cleanPath() call here to canonicalize the path. Otherwise you'll end up with things like "/." as the path, which would fail to be found.

eikenb:
I've pulled the branch down locally and fixed the couple issues I mentioned w/ the inline comments. With it working I wanted to see how it affected my current backend and it breaks it. If it breaks mine I'm sure it will break others and that I've committed to the semantic versioning guarantee not to break backward compatibility with minor version bumps. I'm really not ready at the moment to do a major version bump.

So I'm trying to think if it is possible to enable this without breaking the current backends that rely on receiving a  canonicalized path. I'm thinking something more along the lines of keeping the current path information as is, but somehow allowing access to the path as it was originally sent. It would also require code to handle the case where people didn't handle the Realpath request method, defaulting to the old behavior.

I appreciate the work you put into this and the feature does seem to have some general interest, so I'd like to implement it if I can, just not in its current form. I'll look into helping with this as I have time, but if you want to take a stab at it please do.

request-server.go
robermorales: You are right. I will review it. Thanks.

request-example.go
robermorales: It is only an example, but I agree that can be improved with `filepath.Clean`.

puellanivis:
Should this be closed in favor of PR-238?

drakkan:
This is now implemented, see #437

PR #235, opened 2018-03-19
Send unsupported error on extended packets.

### LINKED ISSUE #223: Returning ssh_FX_OP_UNSUPPORTED from sftp.Handlers
From what I can tell following the code, it does not seem possible to send the unsupported error since `statusFromError` does not include a way to set it. I'd propose that the sftp package define an error such as `ErrOperationUnsupported` and the `statusFromError` be updated to check for that error and set the return `StatusError.Code` appropriately.

### LINKED ISSUE #233: Unknown extended packet handling
I was testing out integration with a few different clients and came across the fact that WS_FTP (a terrible windows client) sends an extended packet for something called "xquota". I wound up having to throw a `fmt.Println` in [packet.go#L900](https://github.com/pkg/sftp/blob/master/packet.go#L900) to pull this out.

It seems to me there should be some way for the server to tell the client this is unsupported and continue on just fine, but am unclear on the right approach there. 

I was looking over [this rfc](https://tools.ietf.org/html/draft-ietf-secsh-filexfer-extensions-00), which seems to indicate an SSH_FXP_STATUS packet can be returned to indicate an error. I wonder if it makes sense to add a struct that implements `sshFxpExtendedPacket.SpecificPacket` which writes an appropriate `SSH_FXP_STATUS` packet by default rather than returning an error here?

crazed:
Following the rules outlined here:

https://tools.ietf.org/html/draft-ietf-secsh-filexfer-extensions-00

Return an SSH_FXP_STATUS with appropriate status error for extended
packets that we do not support.

Fixes #233

### DIFF
--- a/packet-typing.go
+++ b/packet-typing.go
@@ -127,7 +127,9 @@ func makePacket(p rxPacket) (requestPacket, error) {
 		return nil, errors.Errorf("unhandled packet type: %s", p.pktType)
 	}
 	if err := pkt.UnmarshalBinary(p.pktBytes); err != nil {
-		return nil, err
+		// Return partially unpacked packet to allow callers to return
+		// error messages appropriately with necessary id() method.
+		return pkt, err
 	}
 	return pkt, nil
 }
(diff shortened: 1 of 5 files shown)


eikenb:
It looked good and has been merged. Thanks!

PR #237, opened 2018-04-10
Add ability to set restricted root directory

TheMasterGabriel:
I would like to see the ability to set the root directory of the SSH server, so that users cannot access files and directories outside that new root. I'm not sure whether or not it is possible to limit the filesystem served, so as an alternative this PR denies clients permission when that is attempted.

This was an experiment, and there might be a much better way of doing so. I figured I should PR it anyway in case it sparks someone to think of something brilliant, and to hopefully get this functionality implemented.

### DIFF
--- a/server.go
+++ b/server.go
@@ -138,7 +138,7 @@ func (svr *Server) sftpServerWorker(pktChan chan requestPacket) error {
 
 		// permission checks
 		permiss := true
-		if svr.systemRoot != "" {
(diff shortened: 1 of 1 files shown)


TheMasterGabriel:
Not really user friendly, actually. Other PRs seem to do a better job at accomplishing this.

TheMasterGabriel:
Closed to switch source to alternate branch.

PR #92, opened 2016-04-13
Support setting working directory as ServerOption and cleanup

tv42:

PR #238, opened 2018-04-11
Add ability to set restricted root directory

thearchitector:
I would like to see the ability to set the root directory of the SSH server, so that users cannot access files and directories outside that new root. I'm not sure whether or not it is possible to limit the filesystem served, so as an alternative this PR denies clients permission when that is attempted.

This was an experiment, and there might be a much better way of doing so. I figured I should PR it anyway in case it sparks someone to think of something brilliant, and to hopefully get this functionality implemented.

This was originally in a different PR, but I switch the source branches for my own sanity.


eikenb:
I'd consider merging this but I'm planning on merging the filesystem based server implementation into the request based server as a provided backend. And as the request server doesn't have a system for doing this yet it would just end up going away at that point.

That merge is probably a ways off, but I'm hesitant to merge this until there is at least an idea of how it might be supported in the request server. See PR #92 and PR #231, and issue #230.

AkihiroSuda:
What's current status?

drakkan:
> What's current status?

If you use the request server you can easily restrict the root directory yourself. For example I do this in [SFTPGo](https://github.com/drakkan/sftpgo). 

This change only apply to the server implementation, it only allow to serve local filesystem. I think the server implementation is not future proof and should be removed, so I'm against merging this change

AkihiroSuda:
Unlike github.com/pkg/sftp, SFTPGo is GPLv3-licensed and hard to combine with other libraries 😞

drakkan:
> Unlike github.com/pkg/sftp, SFTPGo is GPLv3-licensed and hard to combine with other libraries

SFTPGo is a standalone application, not a library. It provides a full SFTP server (and it supports FTP/s and WevDAV too).

If you need to customize something, several hooks are available. You can customize SFTPGo by writing your own hooks, it is not meant to be combined with other libraries.

Regarding this ticket, this is only my opinion, this patch could be merged anyway if other contributors agree

puellanivis:
The implementation as proposed is complicated by a number of factors. First, `strings.Contains(filename, "..")` will match `"not/a/bad/path/filename..ex"`.

I can probably take a look at this and think about how we might be able to resolve the issue properly. Really, we should never be opening up access to arbitrary files above some directory point anyways.

thearchitector:
It's been a while, but looking at this again that `strings.Contains` check was a hacky way of determining if a file exists as a child of the parent path. It would be much better to convert the paths to absolute paths and check if the system root prepends the path in question. I can edit this PR later today when I get the chance.

drakkan:
I didn't tested your code but is seems quite weak, please add some tests to check that the root dir cannot be escaped in this cases:

1) the root dir is `/tmp/home` and you have a symlink like this one `/tmp/home/dir -> /tmp/home1/dir`
2) if you rename a file/dir the target could not exist, check what happen if a parent path for the non-existing target dir is a symlink outside the home dir

d4z3x:
I added the code by hand, and it seems to work fine, but perhaps a os.Chdir is also needed, or it runs from wherever the binary is called from. Is this the proper way of doing this?

d4z3x:
> > What's current status?
> 
> If you use the request server you can easily restrict the root directory yourself. For example I do this in [SFTPGo](https://github.com/drakkan/sftpgo).
> 
> This change only apply to the server implementation, it only allow to serve local filesystem. I think the server implementation is not future proof and should be removed, so I'm against merging this change

I really like SFTPGo, but I have no idea how to create hooks, let's say with Portable mode running as a subsystem. Is there any documentation I can look at or I missed? My use case is to jail a user, and ensure files are written with as a specific user and permissions.

Is what I'm suggesting possible? And where can I see in sftp-go how you restrict a user to a root dir?

drakkan:
Hi @d4z3x 

the SFTPGo code to check the restricted root for local fs is [here](https://github.com/drakkan/sftpgo/blob/master/vfs/osfs.go#L271). This code should avoid to escape the restricted root via symlinks.

If you have questions about SFTPGo please open issues on its GitHub page. 

In general SFTPGo isn't a library but a system command/service. An hook is a simple external executable (for example a bash script) or an HTTP server that accepts the SFTPGo HTTP notifications.

For example if you write a simple hook like this one in the file `/tmp/logger.sh`:

```
#!/bin/bash

echo `date` $@ >> /tmp/command_sftp.log
echo `date` "operation: $1 username: $2 path: $3 target: $4 ssh cmd: $5" >> /tmp/command_sftp.log
echo `date` "env, action: $SFTPGO_ACTION username: $SFTPGO_ACTION_USERNAME path: $SFTPGO_ACTION_PATH target: $SFTPGO_ACTION_TARGET ssh cmd: $SFTPGO_ACTION_SSH_CMD file size: $SFTPGO_ACTION_FILE_SIZE fs provider: $SFTPGO_ACTION_FS_PROVIDER bucket: $SFTPGO_ACTION_BUCKET endpoint: $SFTPGO_ACTION_ENDPOINT" status: $SFTPGO_ACTION_STATUS >> /tmp/command_sftp.log

exit 0
```

and in `sftpgo.json` you set:

```
"actions": {
      "execute_on": ["upload","download","delete","rename","ssh_cmd"],
      "hook": "/tmp/logger.sh"
    },
```

you will have a log such as this one in `/tmp/commands.log` after uploading a file

```
dom 25 ott 2020, 09:16:28, CET upload a /tmp/test/file.txt
dom 25 ott 2020, 09:16:28, CET operation: upload username: a path: /tmp/test/file.txt target:  ssh cmd: 
dom 25 ott 2020, 09:16:28, CET env, action: upload username: a path: /tmp/test/file.txt target:  ssh cmd:  file size: 4034 fs provider: 0 bucket:  endpoint:  status: 1
```

For some other hooks examples take a look here:

https://github.com/drakkan/sftpgo/tree/master/examples/OTP/authy
https://github.com/drakkan/sftpgo/tree/master/examples/ldapauthserver
https://github.com/drakkan/sftpgo/blob/master/docs/external-auth.md

In SFTPGo you can map all the users to the system account that runs SFTPGo (all platforms are supported) or you can run SFTPGo as root user and map each user to a different system account (*NIX only).

Portable mode is for simple file sharing on demand, restricted root works but for the advanced features (such as hooks etc) you have to run SFTPGo as service. If you want a stateless service you can use the `memory` provider and load the initial users from a file.

farhany:
Hi drakkan,

> the SFTPGo code to check the restricted root for local fs is [here](https://github.com/drakkan/sftpgo/blob/master/vfs/osfs.go#L271). This code should avoid to escape the restricted root via symlinks.

This looks really interesting. Thank you for such a detailed message. Those of us not used to this protocol, it's a big of an undertaking to grasp. 

> If you have questions about SFTPGo please open issues on its GitHub page.

Investigating and will ask in that repo. But this gives me enough of a blue print to move forward a little bit. I'm afraid, I'd have to extend this current library myself, as the changes of interest are not being merged/looked at. I also find that Transmit (on the Mac) often hangs with this lib/server.

drakkan:
> Hi drakkan,
> 
> > the SFTPGo code to check the restricted root for local fs is [here](https://github.com/drakkan/sftpgo/blob/master/vfs/osfs.go#L271). This code should avoid to escape the restricted root via symlinks.
> 
> This looks really interesting. Thank you for such a detailed message. Those of us not used to this protocol, it's a big of an undertaking to grasp.
> 

No problem, you are welcome

> > If you have questions about SFTPGo please open issues on its GitHub page.
> 
> Investigating and will ask in that repo. But this gives me enough of a blue print to move forward a little bit. I'm afraid, I'd have to extend this current library myself, as the changes of interest are not being merged/looked at. I also find that Transmit (on the Mac) often hangs with this lib/server.

If you are able to reproduce these hangs against sftpgo please open an issue [here](https://github.com/drakkan/sftpgo) providing the logs, I'll try to take a look, thank you

d4z3x:
> > Hi drakkan,
> > > the SFTPGo code to check the restricted root for local fs is [here](https://github.com/drakkan/sftpgo/blob/master/vfs/osfs.go#L271). This code should avoid to escape the restricted root via symlinks.
> > 
> > 
> > This looks really interesting. Thank you for such a detailed message. Those of us not used to this protocol, it's a big of an undertaking to grasp.
> 
> No problem, you are welcome
> 
> > > If you have questions about SFTPGo please open issues on its GitHub page.
> > 
> > 
> > Investigating and will ask in that repo. But this gives me enough of a blue print to move forward a little bit. I'm afraid, I'd have to extend this current library myself, as the changes of interest are not being merged/looked at. I also find that Transmit (on the Mac) often hangs with this lib/server.
> 
> If you are able to reproduce these hangs against sftpgo please open an issue [here](https://github.com/drakkan/sftpgo) providing the logs, I'll try to take a look, thank you

Will do. 

> If you are able to reproduce these hangs against sftpgo please open an issue [here](https://github.com/drakkan/sftpgo) providing the logs, I'll try to take a look, thank you

For sftp, it's a bit difficult at the moment until I re-factor the example to keep running. The example in the lib is easy to make into a subsystem, but the rest of it is difficult to follow for me at this time. I suspect the patch to chroot a user, while working does cause Transmit errors if you try to move outside of the dir with an error. I'll post that once I get to my setup.

farhany:
Hi drakkan,

> the SFTPGo code to check the restricted root for local fs is [here](https://github.com/drakkan/sftpgo/blob/master/vfs/osfs.go#L271). This code should avoid to escape the restricted root via symlinks.

This looks really interesting. Thank you for such a detailed message. Those of us not used to this protocol, it's a big of an undertaking to grasp. 

> If you have questions about SFTPGo please open issues on its GitHub page.

Investigating and will ask in that repo. But this gives me enough of a blue print to move forward a little bit. I'm afraid, I'd have to extend this current library myself, as the changes of interest are not being merged/looked at. I also find that Transmit (on the Mac) often hangs with this lib/server.

drakkan:
> Hi drakkan,
> 
> > the SFTPGo code to check the restricted root for local fs is [here](https://github.com/drakkan/sftpgo/blob/master/vfs/osfs.go#L271). This code should avoid to escape the restricted root via symlinks.
> 
> This looks really interesting. Thank you for such a detailed message. Those of us not used to this protocol, it's a big of an undertaking to grasp.
> 
> > If you have questions about SFTPGo please open issues on its GitHub page.
> 
> Investigating and will ask in that repo. But this gives me enough of a blue print to move forward a little bit. I'm afraid, I'd have to extend this current library myself, as the changes of interest are not being merged/looked at. I also find that Transmit (on the Mac) often hangs with this lib/server.

The request server requires you write some code to implement the required interfaces. As explained above I have no time/interest to maintain the old server implementation that only work with local filesystem.

If you want a ready to go solution you can use SFTPGo, or copy its code if you can comply with AGPLv3 license. 

As for the hangs, if I remember correctly you have modified the example to use the library as an SSH subsystem. If so please open a new issue sharing a reproducer. Before doing that, you could check if the hangs also happen using SFTPGo in SSH subsystem mode.

Thank you

PR #239, opened 2018-04-25
Adds a MkdirAll implementation for the client to bring it closer to os Filesystem interface

urjitbhatia:
Test results at: https://travis-ci.org/urjitbhatia/sftp/builds/370926952

(includes a client `MkdirAll` test)

### DIFF
--- a/client.go
+++ b/client.go
@@ -3,12 +3,11 @@ package sftp
 import (
 	"bytes"
 	"encoding/binary"
-	"fmt"
 	"io"
 	"os"
 	"path"
-	"strings"
 	"sync/atomic"
+	"syscall"
 	"time"
 
 	"github.com/kr/fs"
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -680,6 +682,32 @@ func (c *Client) Mkdir(path string) error {
 	}
 }
 
+// MkdirAll creates a directory named path, along with any necessary parents,
+// and returns nil, or else returns an error.
+// If path is already a directory, MkdirAll does nothing and returns nil.
+// If path contains a regular file, an error is returned
+func (c *Client) MkdirAll(path string) error {
+	parts := ""
+	for _, p := range strings.Split(path, "/") {
```
puellanivis: Path here may not be clean, and may not be using "/" as a directory separator.

client.go
```diff
@@ -680,6 +682,32 @@ func (c *Client) Mkdir(path string) error {
...
+		if p == "" {
+			continue
+		}
+		parts += "/" + p
```
puellanivis: String building with concatenation is O(n²) time and space. It should be avoided in tight loops.

client.go
```diff
@@ -680,6 +682,32 @@ func (c *Client) Mkdir(path string) error {
...
+		dir, err := c.Stat(parts)
+		if err == nil {
+			if !dir.IsDir() {
+				return fmt.Errorf("Found a non-directory file on path: %s", parts)
```
puellanivis: One should return the same error as os.MkdirAll here, `&PathError{"mkdir", parts, syscall.ENOTDIR}`

urjitbhatia:
Thanks for taking a look @puellanivis - I'll update the PR with your suggestions.

urjitbhatia:
Ah, Travis seems to have had some connectivity issues:
```
Fetching https://golang.org/x/crypto/ssh?go-get=1
https fetch failed: Get https://golang.org/x/crypto/ssh?go-get=1: net/http: TLS handshake timeout
```
But tests passed here: https://travis-ci.org/urjitbhatia/sftp

@puellanivis can you retrigger the build on this PR? or any other suggestions to retrigger the build?

urjitbhatia:
Hmm, actually, I realize that the `IsPathSeparator` part won't work on the remote server if it is windows...

urjitbhatia:
Nevermind prev comment, I re-read the FTP spec and it does say separator is `/` so this should be good to go

https://tools.ietf.org/html/draft-ietf-secsh-filexfer-13#section-9.1
```
6.  File Names

   This protocol represents file names as strings.  File names are
   assumed to use the slash ('/') character as a directory separator.
```

puellanivis:
What I had referred to in my initial review was that we have various “clean path” sort of functions here and there, which are intended to normalize paths before use.

urjitbhatia:
Won't `mkdir` have the same problem currently then? It doesn't seem to be calling `cleanpath` from the client anywhere. The server does though, on the receiving side. So this should work?

eikenb:
I don't think having a clean path is as important on the Client as on the Server. On the client you have control of the path input or, in other words, you can make sure it is clean before sending it. The server has to deal with what the client sends, so it making sure it has a clean path is important.

urjitbhatia:
Apologies for my confusion...
@eikenb So this `mkdirall` implementation on the Client should be good to be merged?

Also happy to add the `cleanPath` checks for this (Client side) if that's what you guys recommend.

urjitbhatia:
@puellanivis hey folks, any pointers about next steps here?

puellanivis:
In my case, I think any feedback I could offer would not actually be helpful, or useful. So… I’ve just been keeping my mouth shut.

eikenb:
Finally had a chance to review locally and it looks fine. I find it is bit weird that os.MkdirAll doesn't use the library path primitives (Dir), but there is probably a reason. This code follows its example exactly.

puellanivis:
Since the filepath imports os, os cannot use the filepath library. Meanwhile, since the path library assumes a "/" path delimiter, os cannot use the path library.

I had originally recommended the os.MkdirAll as a guideline, because it avoids a lot of pitfalls that had been in the first implementation proposed. I had not initially really intended for it to be so closely copied.

eikenb:
@puellanivis Thanks for the clarification. That makes total sense.

PR #244, opened 2018-05-10
 set flags and attrs for fsetstat packet type in filecmd

sigmonsays:


sigmonsays:
sounds good, i'll make the change shortly.

sigmonsays:
So on second thought.. there is an issue with this approach because *sshFxpFsetstatPacket does not have a path, and that's how pkt is typed. I wasn't sure changing the signature of that function (requestFromPacket) was correct. Should I make a new type which combines both hasPath and hasHandle ?

eikenb:
Hold on.. let me take another look at that code. I thought I had Fsetstat working like Setstat, but I may be mis-remembering.

eikenb:
Sorry, you were correct. The requestFromPacket() method doesn't even get called for an sshFxpFsetstatPacket packet. The Request object gets created when the initial sshFxpOpenPacket is processed, so the incoming sshFxpFsetstatPacket would find the existing handle/Request and use it.

Unfortunately there doesn't seem like there is a better place for it. I had hoped to have the request object all situated by that point in the code, but there is just no place in the code obviously better than in your original fix. I'm thinking I should just merge it for now and maybe something will come to me during a later refactor.

sigmonsays:
Thanks. As I spend more time in the code perhaps i'll be able to help come up with a better solution as well.

eikenb:
Sigh. I pulled it down and rebased/merged it locally and push. I thought that would close this out but it seems my rebase screwed that up. But it is merged and you got credit.

eikenb:
Forgot to say thanks. I appreciate your looking into and fixing this... addressed not only your issue but one other. Been super lacking on time recently and it is a big help.

PR #248, opened 2018-05-12
Write packet length header & payload together on the wire.

bradfitz:


packet.go
```diff
@@ -125,15 +125,14 @@ func sendPacket(w io.Writer, m encoding.BinaryMarshaler) error {
...
 	}
-	l := uint32(len(bb))
-	hdr := []byte{byte(l >> 24), byte(l >> 16), byte(l >> 8), byte(l)}
-	_, err = w.Write(hdr)
-	if err != nil {
-		return errors.Errorf("failed to send packet header: %v", err)
-	}
-	_, err = w.Write(bb)
+	// Slide packet down 4 bytes to make room for length header.
+	packet := append(bb, make([]byte, 4)...) // optimistically assume bb has capacity
```
bradfitz: Btw, this is recognized and optimized by the compiler in Go 1.11, but is valid code in prior versions too.

But sending more TCP packets and SSH packets dwarfs any such optimizations. Just an FYI. The point of this PR is to reduce Write calls, not to optimize allocations or reduce memory copies.

eikenb: I was curious about what optimization he meant here and just in case anyone else is...

https://github.com/golang/go/issues/21266

eikenb:
I noticed a slight downgrade in throughput resulting from this so am waiting for the 1.11 release w/ its optimization to merge this.

PR #254, opened 2018-05-25
Allow to configure maxConcurrentRequests

gm42:
Change the constant `maxConcurrentRequests` into a configurable option (per-file).

Admittedly here the most useful option would be `MaxConcurrentRequestsPerClient` but a per-file limit is easier to implement as it follows 1:1 the current functionality.

### DIFF
--- a/client.go
+++ b/client.go
@@ -76,6 +76,19 @@ func MaxPacket(size int) ClientOption {
 	return MaxPacketChecked(size)
 }
 
+// MaxConcurrentRequestsPerFile sets the maximum concurrent requests allowed for a single file.
+//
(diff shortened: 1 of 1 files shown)


eikenb:
You don't really need to add the maxConcurrentRequests to the File object, as it stores a reference to the Client and the client already has it as a field. So instead of f.maxConcurrentRequests you'd have f.c.maxConcurrentRequests. Or you could pull that out to a local variable if you like that better.

gm42:
@eikenb my bad, I did not see the `c` field; I have updated it to use that.

eikenb:
Thanks for the contribution.

PR #255, opened 2018-05-26
Opendir return an error status when not found

### LINKED ISSUE #253: OPENDIR is not processed correctly
As part of the SFTP Version 3 spec related to OPENDIR:

```
   The SSH_FXP_OPENDIR opens a directory for reading.  It has the
   following format:

        uint32     id
        string     path

   where `id' is the request identifier and `path' is the path name of
   the directory to be listed (without any trailing slash).  See Section
   ``File Names'' for more information on file names.  This will return
   an error if the path does not specify a directory or if the directory
   is not readable.  The server will respond to this request with either
   a SSH_FXP_HANDLE or a SSH_FXP_STATUS message.
```

In the case of both the request-server and server implementation, it will never return a SSH_FXP_STATUS  because this call is validated to ensure the directory exists. In the case of the server, it should check if the path requested exists on the server. In the case of the request-server, it should ask the handler if the directory exists.

In the case where the directory does not exist, then this call should be failing. However, what is happening is this call successes, so the SFTP client assumes the directory exists. A READDIR then is attempted, which errors.

Filezilla in this case burns because of it, as it followed the spec and assumed it was there. It doesn't react too well with this violation.

eikenb:
The initial Opendir packet is supposed to repond with an error status if
the directory wasn't found. It was just returning a handle without
checking, now it does a Stat on the path and only returns the handle if
the Stat is successful, otherwise it returns an error.

Fixes #253

### DIFF
--- a/request-server.go
+++ b/request-server.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding"
 	"io"
+	"os"
 	"path"
 	"path/filepath"
 	"strconv"
@@ -173,8 +174,18 @@ func (rs *RequestServer) packetWorker(
 			rpkt = cleanPacketPath(pkt)
 		case *sshFxpOpendirPacket:
 			request := requestFromPacket(ctx, pkt)
-			handle := rs.nextRequest(request)
-			rpkt = sshFxpHandlePacket{pkt.id(), handle}
+			rpkt = request.call(rs.Handlers, pkt)
+			request.close()
+			if stat, ok := rpkt.(*sshFxpStatResponse); ok {
+				if stat.info.IsDir() {
+					request := requestFromPacket(ctx, pkt)
+					handle := rs.nextRequest(request)
+					rpkt = sshFxpHandlePacket{pkt.id(), handle}
+				} else {
+					rpkt = statusFromError(pkt, &os.PathError{
+						Path: request.Filepath, Err: syscall.ENOTDIR})
+				}
+			}
 		case *sshFxpOpenPacket:
 			request := requestFromPacket(ctx, pkt)
 			handle := rs.nextRequest(request)
--- a/server.go
+++ b/server.go
@@ -240,6 +240,12 @@ func handlePacket(s *Server, p interface{}) error {
 			}},
 		})
 	case *sshFxpOpendirPacket:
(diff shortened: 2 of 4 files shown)


server.go
```diff
@@ -240,6 +240,9 @@ func handlePacket(s *Server, p interface{}) error {
 			}},
 		})
 	case *sshFxpOpendirPacket:
+		if _, err := os.Stat(p.Path); err != nil {
```
LordRalex: Technically, this would not error if the path requested is not a folder. The spec indicates it must be a directory, so a simple stat isn't complying here. You'd have to make sure the FileInfo.IsDir() is true also.

eikenb: That was a dumb miss on my part. Good catch, thanks.

puellanivis: It’s not dumb. It’s a brain fart, and is what code reviews are for! ;)

eikenb:
Updated the PR with IsDir checks.

request-server.go
```diff
@@ -173,8 +174,18 @@ func (rs *RequestServer) packetWorker(
 			rpkt = cleanPacketPath(pkt)
 		case *sshFxpOpendirPacket:
 			request := requestFromPacket(ctx, pkt)
-			handle := rs.nextRequest(request)
-			rpkt = sshFxpHandlePacket{pkt.id(), handle}
+			rpkt = request.call(rs.Handlers, pkt)
+			request.close()
```
LordRalex: Just more curious, is there a reason to close the request here, and then make a new one to use for the handle? Seems like the original could be used instead of explicitly closing it to just make it again.

eikenb: It is the quickest/cleanest way to reset the Request back to 0 values. The Stat call for the Opendir and the (usually) following Readdir would both use the same Request.state and if it isn't reset they conflict.

Really Stat and Readdir shouldn't conflict this way as it is legal for a client to do this. I'll need to test if it is an actual bug though, as it might not be in practice (when a client is calling vs internal use). I'll make a note to come up with a test for this. I don't have time ATM.

Thanks.

eikenb:
I'm going to go ahead and merge this though I might rework the request replacement part once I've addressed #256.

PR #262, opened 2018-07-26
added build tags required for gcc-go on AIX

ghafoors:
while building on AIX Unix, as os name is "aix" these files are not included in the build hence the failure.
with the added build tag it works fine and build is successful.


eikenb:
Thanks for the contribution.

ghafoors:
Happy to help

PR #271, opened 2018-09-11
Fix race condition when getting a file reader.

marshallbrekka:
This fixes a race condition where multiple successive read calls could result in multiple readers being created, and even worse because they were all assigned to the same state var, many of the opened ones would not get closed when the client closed the handler.

### DIFF
--- a/request.go
+++ b/request.go
@@ -204,13 +204,17 @@ func packetData(p requestPacket) (data []byte, offset int64, length uint32) {
 // wrap FileReader handler
 func fileget(h FileReader, r *Request, pkt requestPacket) responsePacket {
 	var err error
-	reader := r.getReader()
+	r.state.Lock()
+	reader := r.state.readerAt
 	if reader == nil {
 		reader, err = h.Fileread(r)
-		if err != nil {
-			return statusFromError(pkt, err)
+		if err == nil {
+			r.state.readerAt = reader
 		}
-		r.setReaderState(reader)
+	}
+	r.state.Unlock()
+	if err != nil {
+		return statusFromError(pkt, err)
 	}
 
 	_, offset, length := packetData(pkt)


eikenb:
Thank you very much for the submission. I'll review it soon (not today, I'm spent).

puellanivis:
TL;DR: good code, but it probably should have all been in `getReader` function from the start.

So, this is a good catch. There absolutely is a relocking problem here that results in a race-condition.

However, I think a better solution to this would be to change the `getReader` function signature to take the `*Request`, and then you can put the whole `getReader; if readerInvalid { makeReader; setReaderForFutureCalls }` functionality into that one `getReader` function.

This also then lets you use the `mu.Lock(); defer mu.Unlock()` pattern and avoiding putting off checking the `err != nil` condition from `h.Fileread(r)` until after you unlock the Mutex so as to ensure that the mutex is properly unlocked, which is clever, but looks odd and confusing.

eikenb:
I want to run a few tests on the impact of changing the getReader() lock from a read lock to a rw lock. See if it should be optimized to something more like getRequest() with the initial RLock and the duplicate test.

eikenb:
Re-structuring the read/write/open packet handling isn't a bad idea but I'll have to think about how and use might look. I think we should probably fix the race first, then look into refactoring it to make more sense. So focusing on an ugly-but-straightforward locking fix for now with a refactor to clean it up.

The get/set reader/writer methods are both localized to that file and only the gets are used more than once (twice). So the initial ugly fix shouldn't be to hard.

And yeah.. I knew fileput has the same issue as they are nearly identical in that part.

eikenb:
I just pushed PR #272 as an alternative fix for this issue. It only fixes the locking issue and doesn't attempt the refactor @marshallbrekka suggested. I'm going to look into that more later.

PR #272, opened 2018-09-14
fix race condition in fileget/fileput

eikenb:
This fixes the concurrent access/locking race in both fileget and
fileput when accessing the saved readerAt/writerAt. It eliminates the
dedicated functions and moves the access/locking inline where they were
called to make it clearer.

This is an alternative to the fix in #271.


request.go
```diff
@@ -230,13 +219,20 @@ func fileget(h FileReader, r *Request, pkt requestPacket) responsePacket {
...
+	r.state.RUnlock()
 	if writer == nil {
-		writer, err = h.Filewrite(r)
-		if err != nil {
-			return statusFromError(pkt, err)
+		r.state.Lock()
+		if r.state.writerAt == nil {
+			r.state.writerAt, err = h.Filewrite(r)
+			if err != nil {
+				r.state.Unlock()
```
puellanivis: I had recommended on the other PR moving this code into the `getReader` / `getWriter` function to be able to use the `defer r.state.Unlock()` so that one can avoid having to ensure that one puts unlocks in every code path.

It’s not that this code is wrong, it’s rather just a defensive coding tactic that I like to follow.

eikenb: Normally I agree, but I wanted to group the code together like that as I thought it'd help make the refactoring simpler by eliminating the extra indirection. I also noticed in my testing that the defer still adds about a 50% overhead to the calls and thought I'd go with the faster method until I had a chance to test how it might affect performance/throughput more thoroughly.

puellanivis: :+1: sounds good.

PR #279, opened 2018-12-03
Add Wait method to detect underlying SFTP connection closed

### LINKED ISSUE #278: Client need a way to detect underyling SFTP connection closed
When sftp-server have been killed or crashed, there need a way to detect SFTP connection closed and reconnect.

Like ssh which provides a Wait method, sftp can also provides a Wait method in https://github.com/pkg/sftp/blob/08de04f133f27844173471167014e1a753655ac8/conn.go#L34

wutzx:
Closes #278

### DIFF
--- a/client.go
+++ b/client.go
@@ -6,7 +6,6 @@ import (
 	"io"
 	"os"
 	"path"
-	"sync"
 	"sync/atomic"
 	"syscall"
 	"time"
@@ -123,7 +122,7 @@ func NewClientPipe(rd io.Reader, wr io.WriteCloser, opts ...ClientOption) (*Clie
 				WriteCloser: wr,
 			},
 			inflight: make(map[uint32]chan<- result),
-			errCond:  sync.NewCond(new(sync.Mutex)),
+			closed:   make(chan struct{}),
 		},
 		maxPacket:             1 << 15,
 		maxConcurrentRequests: 64,
--- a/client.go
+++ b/client.go
@@ -122,6 +122,7 @@ func NewClientPipe(rd io.Reader, wr io.WriteCloser, opts ...ClientOption) (*Clie
 				WriteCloser: wr,
 			},
 			inflight: make(map[uint32]chan<- result),
+			errCh:    make(chan error, 1),
 		},
 		maxPacket:             1 << 15,
 		maxConcurrentRequests: 64,
--- a/client.go
+++ b/client.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"os"
 	"path"
+	"sync"
 	"sync/atomic"
 	"syscall"
 	"time"
@@ -122,7 +123,7 @@ func NewClientPipe(rd io.Reader, wr io.WriteCloser, opts ...ClientOption) (*Clie
 				WriteCloser: wr,
 			},
 			inflight: make(map[uint32]chan<- result),
-			errCh:    make(chan error, 1),
+			errCond:  sync.NewCond(new(sync.Mutex)),
 		},
 		maxPacket:             1 << 15,
 		maxConcurrentRequests: 64,
--- a/conn.go
+++ b/conn.go
@@ -36,13 +36,21 @@ type clientConn struct {
 	wg         sync.WaitGroup
 	sync.Mutex                          // protects inflight
 	inflight   map[uint32]chan<- result // outstanding requests
-	errCh      chan error
+
+	errCond *sync.Cond
+	err     error
 }
 
 // Wait blocks until the conn has shut down, and return the error
-// causing the shutdown.
+// causing the shutdown. It can be called concurrently from multiple
+// goroutines.
 func (c *clientConn) Wait() error {
-	return <-c.errCh
+	c.errCond.L.Lock()
+	defer c.errCond.L.Unlock()
+	for c.err == nil {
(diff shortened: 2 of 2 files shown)


conn.go
```diff
@@ -36,6 +36,13 @@ type clientConn struct {
 	wg         sync.WaitGroup
 	sync.Mutex                          // protects inflight
 	inflight   map[uint32]chan<- result // outstanding requests
+	errCh      chan error
+}
+
+// Wait blocks until the conn has shut down, and return the error
+// causing the shutdown.
+func (c *clientConn) Wait() error {
```
puellanivis: What happens if two goroutines are calling `Wait()` at the same time?

wutaizeng: The Wait method should be called once. The one will be exited, the another one will be block forever when SFTP connection closed.

wutaizeng: If we need Wait method to support called multiple times, may be implement it like ssh Wait method.

https://github.com/golang/crypto/blob/master/ssh/mux.go#L106-L113

puellanivis: My intent was to provoke you to think about how this will be used, and what contracts need to be documented and what behavior people would expect from the library.

If we write this such that it may be called once and only once, we need to resolve how we would handle it being called more than once:
1. Document that it cannot be used this way, and then point people to the code comment documenting the behavior when it comes in as a support ticket.
2. Return an error in the case where it is called more than once, document that it returns an error in this case, and then wait for the PR to come in, which adds the functionality of handling an arbitrary number of calls.

Meanwhile, as you note, `ssh.Wait` can handle being called multiple times, and if we're clearly mimicking that library, it would be reasonable for them to expect the same behavior here.

eikenb: The ssh.Wait() method does seem to behave more like @wutaizeng's version. If you call it twice the second call will block forever, even if an error was returned by the call. The documentation isn't clear about this behavior, but here's a quicky example that shows it.

https://gist.github.com/eikenb/d3a92a399cd20beb54f1173246f32e2f

eikenb: Sorry, it seems we were talking 2 different Wait calls. I was looking at the public API Session.Wait, while @wutaizeng was pointing to an internal one on the mux struct.

puellanivis: I think it’s still basically relevant that `ssh.Session.Wait` blocks forever if called more than once.

But 😕 I find that confusing behavior. There are a lot of cases where one might want to wait on a connection and do something after there is a disconnect, and at least some of those I could imagine someone wanting to do in separate goroutines.

But then as well, `This package does not fall under the stability promise of the Go language itself, so its API may be changed when pressing needs arise.`

client.go
```diff
@@ -122,7 +123,7 @@ func NewClientPipe(rd io.Reader, wr io.WriteCloser, opts ...ClientOption) (*Clie
 				WriteCloser: wr,
 			},
 			inflight: make(map[uint32]chan<- result),
-			errCh:    make(chan error, 1),
+			errCond:  sync.NewCond(new(sync.Mutex)),
```
puellanivis: Since this triggers at most only once, rather than a `sync.Cond` you can also use the feature that a receive on a closed channel immediately sends the zero value.

So, here you would have a `closed: make(chan struct{})`, where you have the whole `c.errCond.Wait()` block, all of that can be just `<-c.closed; return c.err`. Then instead of `Broadcast` you just have `close(c.closed)`.

This tends to be a much more simple way to get a single-use state-change trigger of a broadcast unlock. While `sync.Cond` is better if it needs to broadcast repeated state-changes.

conn.go
```diff
@@ -36,13 +36,21 @@ type clientConn struct {
...
 
 // Wait blocks until the conn has shut down, and return the error
-// causing the shutdown.
+// causing the shutdown. It can be called concurrently from multiple
+// goroutines.
 func (c *clientConn) Wait() error {
-	return <-c.errCh
+	c.errCond.L.Lock()
+	defer c.errCond.L.Unlock()
+	for c.err == nil {
```
puellanivis: Now that I think about it, this `Wait` should probably return if the connection is shut down properly with a `Close` as well, right? In that case, we would want `c.err == nil`.

conn.go
```diff
@@ -56,7 +64,10 @@ func (c *clientConn) loop() {
 	err := c.recv()
 	if err != nil {
 		c.broadcastErr(err)
-		c.errCh <- err
+		c.errCond.L.Lock()
+		c.err = err
+		c.errCond.Broadcast()
```
puellanivis: It would also probably make sense to have the `c.err` and boadcasting of a signal that it is now set in the `broadcastErr` method?

puellanivis:
Looks good on my side.

wutaizeng:
@eikenb Any plans to merge this PR ?

eikenb:
Sorry for the delay. Personal life went sideways and I'm only now finding time to get back to this. I'll try to review (and hopefully merge) this soon.

conn.go
eikenb: I prefer the current behavior to ssh.Wait's block forever on additional read behavior and I don't see any value from modeling ssh's behavior here.

eikenb:
I'd like to see a test for Wait but I think I'll go ahead and merge this as it's been so long and it is such a simple feature. We can write up a test later.

PR #282, opened 2019-01-19
sftp: update tests to finish and not panic on windows

kardianos:
Many Unix assuptions have been made when testing.
Previously running tests on windows threw a panic
part way through the tests. After these changes
many more tests pass when the fix was isolated
to the test itself, and none panic.

Some tests are skipped because they do make sense
on windows (chmod, chown), while others are skipped
just to how the test was implemented.

Lastly, some of the external executables were hard coded.
Change these to look paths in TestMain first. This is done
to better support windows, where openssh may be installed and
sftp and sftp-server may both be in the PATH.


kardianos:
FYI: I see this as the first step to adding windows to the travis-ci build script, probably similar to how osx is added (exuding old version and tip).

eikenb:
I'll try to review these soon. Been underwater for the last couple months but hope to spend some time on this going forward.

PR #283, opened 2019-01-29
Initialize opening of files/directories upon receiving open packets

### LINKED ISSUE #280: Request.WithContext() deadlock when called from handler
Calling `Request.WithContext()` in our handlers (middleware) causes a deadlock. This only happens after the race condition fix in v1.8.3 (which likely solves another issue we've experienced).

What happens:

1. `fileget()` locks the given `Request` and calls the provided `FileReader` with said request
2. The `FileReader`, which is our handler (middleware) in this case, calls `Request.WithContext()`
3. `Request.WithContext()` calls `Request.copy()`, causing a deadlock

I assume this is unintended given the obvious influence from the `http`-package, where this pattern (enriching the context via middleware) is very common.

eikenb:
This change reworks read/write/readdir to 'open' or initialize the ReaderAt/WriterAt/ListerAt objects upon receiving the Open and Opendir packets, instead of waiting until receiving the Read/Write/Readdir packets as it was doing. This eliminates multiple uses of the RWMutex on  request.state and eliminates the deadlock caused caused by trying to use WithContext().

Fixes #280 

Thanks to @marshallbrekka for the inspiration (https://github.com/pkg/sftp/pull/271#issuecomment-421157790).


eikenb:
There was one change in this that I want to call out. In the existing code when you open a directory it makes a Stat call and uses that to send the appropriate errors. Now that it is calling the Filelist handler, initializing ListerAt, it can get the errors from that and return them. But this shifts the need to check that it is a directory to the Filelist handler.

I implemented it this way as I felt checking that you are working on a directory and returning an error if not (or if it doesn't exist) is something you would already be doing in the Filelist() method and so the additional Stat call would/should be redundant.

Please let me know if you have any thoughts on this. Thanks.

request-example.go
```diff
@@ -209,6 +214,7 @@ func newMemFile(name string, isdir bool) *memFile {
 		name:    name,
 		modtime: time.Now(),
 		isdir:   isdir,
+		content: []byte(""),
```
puellanivis: Is it important that this be set to an empty `[]byte{}` and not use a `nil`?

https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices

request-server_test.go
```diff
@@ -147,7 +147,7 @@ func TestRequestWriteEmpty(t *testing.T) {
 	f, err := r.fetch("/foo")
 	if assert.Nil(t, err) {
 		assert.False(t, f.isdir)
-		assert.Equal(t, f.content, []byte(""))
+		assert.Equal(t, []byte(""), f.content)
```
puellanivis: Probably one should prefer `assert.Equal(t, len(f.content), 0)` to avoid making a possible distinction between a `nil` and an empty byte-slice.

Also, go prefers the order `got COMPARATOR want`  https://github.com/golang/go/wiki/CodeReviewComments#useful-test-failures

Some languages can easily confuse an assignment with a comparator, and so it became stylish to put your test value on the left because it was not a valid assignment L-value. But since go does not allow assignments in conditions, that risk is entirely removed, and the unconventional ordering just to make sure the L-value is unassignable is unnecessary.

request-server_test.go
```diff
@@ -354,7 +351,9 @@ func TestRequestReaddir(t *testing.T) {
 	for i := 0; i < 100; i++ {
 		fname := fmt.Sprintf("/foo_%02d", i)
 		_, err := putTestFile(p.cli, fname, fname)
-		assert.Nil(t, err)
+		if !assert.NoError(t, err) {
```
puellanivis: Since you‘re calling `t.Fatal` below which will log the error anyways, this would probably be clearer as:
```
if err != nil {
  t.Fatal("expected no error, got:", err)
}
```

request-example.go
```diff
@@ -26,6 +27,7 @@ func InMemHandler() Handlers {
 
 // Example Handlers
 func (fs *root) Fileread(r *Request) (io.ReaderAt, error) {
+	_ = r.WithContext(r.Context())
```
puellanivis: I‘m not sure I see the purpose of all of these `_ = r.WithContext(r.Context())` calls. Were they there just for debugging?

eikenb: Issue #280 was totally missed in testing because the Context was never initialized in any test. So I added these to trigger that issue and I decided to leave it in to be sure I wouldn't re-introduce the deadlock by mistake. I tried a couple times to come up with a single/standalone test and I didn't like any of them. So I just kept these as being better than nothing.

Earlier in my career I suffered from being a bit to much of a perfectionist and would over-engineer and take forever to get things out. Now as long as I think it isn't terrible I'll go ahead and get it released knowing it can always be reworked.

Maybe I should add a comment to make sure the intention is clear?

request-example.go
eikenb: That was a quick fix I made early in testing to move on and is now unnecessary. I'll remove it.

request-server_test.go
eikenb: I seem to have an occasional bout of comparator dyslexia as I just am sure the reverse order is correct sometimes. I'll replace it with `assert.Len(t, f.content, 0)` that way I don't need to worry about the nil comparison.

request-server_test.go
eikenb: Another good suggestion.

request-example.go
eikenb: I'm just thinking... `_ = r.WithContext(r.Context()) // initialize context for testing` for now.

eikenb:
@puellanivis Just wanted to add a big THANK YOU for reviewing my code. I really appreciate it.

request-example.go
eikenb: Slight tweak to before... `_ = r.WithContext(r.Context()) // initialize context for deadlock testing` makes it clearer.

eikenb:
I meant to split up the context comments from the other fixes. Reset and forced push the branch and I'm going to re-push the proper commits in a minute.

PR #285, opened 2019-04-11
Add io.ReaderAt interface compatibility

qrpike:
Read is now a proxy for ReadAt. Read uses the returned bytes count to set the new offset for the file, since ReadAt does not alter the offset.

### DIFF
--- a/client.go
+++ b/client.go
@@ -790,8 +790,6 @@ func (f *File) Read(b []byte) (int, error) {
 	return r, err
 }
 
-
-
 // ReadAt reads up to len(b) byte from the File at a given offset `off`. It returns 
 // the number of bytes read and an error, if any. ReadAt follows io.ReaderAt semantics, 
 // so the file offset is not altered during the read.
@@ -886,9 +884,6 @@ func (f *File) ReadAt(b []byte, off int64) (n int, err error) {
 	}
 	return read, firstErr.err
 }
-
-
-
 
 // WriteTo writes the file to w. The return value is the number of bytes
 // written. Any error encountered during the write is also returned.
--- a/client.go
+++ b/client.go
@@ -785,13 +785,24 @@ func (f *File) Name() string {
 // than calling Read multiple times. io.Copy will do this
 // automatically.
 func (f *File) Read(b []byte) (int, error) {
+	r, err := f.ReadAt(b, int64( f.offset ))
+	f.offset = uint64(r)
+	return r, err
+}
+
+
+
+// ReadAt reads up to len(b) byte from the File at a given offset `off`. It returns 
+// the number of bytes read and an error, if any. ReadAt follows io.ReaderAt semantics, 
+// so the file offset is not altered during the read.
+func (f *File) ReadAt(b []byte, off int64) (n int, err error) {
 	// Split the read into multiple maxPacket sized concurrent reads
 	// bounded by maxConcurrentRequests. This allows reads with a suitably
 	// large buffer to transfer data at a much faster rate due to
 	// overlapping round trip times.
 	inFlight := 0
 	desiredInFlight := 1
-	offset := f.offset
+	offset := uint64( off )
 	// maxConcurrentRequests buffer to deal with broadcastErr() floods
 	// also must have a buffer of max value of (desiredInFlight - inFlight)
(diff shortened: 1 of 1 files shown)


client.go
```diff
@@ -785,13 +785,22 @@ func (f *File) Name() string {
...
 func (f *File) Read(b []byte) (int, error) {
+	r, err := f.ReadAt(b, int64( f.offset ))
+	f.offset += uint64(r)
+	return r, err
+}
+
+// ReadAt reads up to len(b) byte from the File at a given offset `off`. It returns 
+// the number of bytes read and an error, if any. ReadAt follows io.ReaderAt semantics, 
+// so the file offset is not altered during the read.
+func (f *File) ReadAt(b []byte, off int64) (n int, err error) {
```
puellanivis: From the documentation about the `io.ReaderAt` semantics:
```
When ReadAt returns n < len(p), it returns a non-nil error explaining why more bytes were not returned. In this respect, ReadAt is stricter than Read.

If some data is available but not len(p) bytes, ReadAt blocks until either all the data is available or an error occurs. In this respect ReadAt is different from Read.
```

It looks like unfortunately, the proper semantics between `io.ReaderAt` means that it cannot be immediately used with an underlying tracking offset as an `io.Reader` like you are doing. 🙁 It’s just close enough to be different.

eikenb: I agree that the io.ReaderAt has different requirements than io.Reader, reading the io.Reader seems to indicate that io.ReaderAt semantics are still valid io.Reader semantics.

So io.ReaderAt specifies...

> If some data is available but not len(p) bytes, ReadAt blocks until either all the data is available or an error occurs. In this respect ReadAt is different from Read.

While io.Reader says...

> If some data is available but not len(p) bytes, Read conventionally returns what is available instead of waiting for more.

I read 'conventionally' here to mean that it can, but doesn't have to, return instead of waiting. That is it is still valid Reader semantics to wait for more data instead of returning thus the ReaderAt behavior is still valid Reader behavior.

The other differing semantics of ReaderAt...

>When ReadAt returns n < len(p), it returns a non-nil error explaining why more bytes were not returned. In this respect, ReadAt is stricter than Read.

If it does wait for more data, then it seems to me that it would fulfill this as well as it would either eventually return the full data or partial data and an error.

If correct then io.ReaderAt semantics is valid io.Reader semantics. So this would all be OK except that we don't want to change the existing behavior. But after a quick reading of the client Read() method, it looks to me like it actually has the more restrictive, io.ReaderAt like semantics. It doesn't stop sending SFTP requests and reading in the returned data until it has filled up the passed in slice (b) or hit an error.

@puellanivis, @qrpike if either/both of you would read through the Read() methods and confirm my reading I'd appreciate it. If it is correct, then this might work.

puellanivis: So, including having read your assessment here, my understanding is that `ReaderAt` is stricter than `Reader` so it _ought_ to work for all uses of `Reader`. So, all we should need to do is ensure the new `ReadAt` method fits the requirements of `ReaderAt` and then it should be good.

Slightly varying specifications are the most enjoyable part of the work, no? :wink:

eikenb: > my understanding is that ReaderAt is stricter than Reader so it ought to work for all uses of Reader

Not only should it work but Reader, as currently implemented, already has the semantics of io.ReaderAt except for the missing offset parameter. I do need to read through it one more time to double check that the semantics do line up.

eikenb:
I still think this might be possible to merge in.. I just have had zero time to test it out. I'll try to get to it as soon as I can find the time. Thanks.

eikenb:
Sorry for the long delay getting to this. It is still on my radar and I'm trying to make more time for this project. I hope to get to it soon. Thank you for your patience.

puellanivis:
This one might be good to get in for a 1.12 release.

nightlyone:
@puellanivis / @eikenb This looks like the last pull request that needs to be merged for a v1.12.0 release. 

Would you mind doing that merge and then the release as well? I see no further work blocking the release.

puellanivis:
So, I’m going to merge this, but also immediately start a new branch to add a mutex to protect the `Read()` call from multiple parallel calls.

It will be a lot safer than someone trying to figure out the inevitable race condition.

PR #293, opened 2019-06-12
Allow -sftp to work in server_integration_test

tommie:


eikenb:
Make the flag actually work as a flag. Thanks!

PR #299, opened 2019-08-09
Fixed a problem that caused securefx to crash when errUnknownExtended…

soopsio:
…Packet replied twice in SFTP server mode

### DIFF
--- a/request-server.go
+++ b/request-server.go
@@ -114,18 +114,9 @@ func (rs *RequestServer) Serve() error {
 
 		pkt, err = makePacket(rxPacket{fxp(pktType), pktBytes})
 		if err != nil {
-			switch errors.Cause(err) {
-			case errUnknownExtendedPacket:
-				if err := rs.serverConn.sendError(pkt, ErrSshFxOpUnsupported); err != nil {
-					debug("failed to send err packet: %v", err)
-					rs.conn.Close() // shuts down recvPacket
-					break
-				}
-			default:
-				debug("makePacket err: %v", err)
-				rs.conn.Close() // shuts down recvPacket
-				break
-			}
+			debug("makePacket err: %v", err)
+			rs.conn.Close() // shuts down recvPacket
+			break
 		}
 
 		pktChan <- rs.pktMgr.newOrderedRequest(pkt)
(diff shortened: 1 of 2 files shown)


drakkan:
Do you have a test case or a way to reproduce ths issue?

soopsio:
> Do you have a test case or a way to reproduce ths issue?

Using SecureFX connection examples/go-sftp-server , SecureFX will crash, 
图片

eikenb:
Thanks for the PR @soopsio! I love bugs reported via fixes. :)

Looking over this it the goal seems to be to get the server to better handle unknown extended packets. Sending an unsupported-status packet instead of just crapping out seems reasonable. Checking the spec for what is supposed to happen... turns out it is pretty much just this...

> If the server does not recognize the `extended-request' name, then the server MUST respond with SSH_FXP_STATUS with error/status set to SSH_FX_OP_UNSUPPORTED.

There are a few problems with this PR though and I'd like to ask..  Do you just want this done (ie. have me do it) or would you like to work with me to fix up your PR so it can be merged? I'm happy with either option.

Thanks.

soopsio:
Now I put SSH_FX_OP_UNSUPPORTED in handlePacket of server.go and packetWorker of request-server.go, because I found that all extension-request will be performed in these places, but when it comes to here, the connection will be closed because it cannot handle extension-request. But when makePacket finds errUnknownExtendedPacket it will reply to the client once again SSH_FXP_STATUS packet, these two places are in conflict, so I remove the back packet after makePacket, and in handlePacket and packetWorker to determine whether the SpecificPacket of *sshFxpExtendedPacket is nil, If it is nil, SSH_FXP_STATUS and SSH_FX_OP_UNSUPPORTED should be replied. If this is not appropriate, please help me to fix this pr, thank you!

eikenb:
@soopsio, no need to explain. The purpose of your code is clear and the points you picked made sense. It is more implementation details than overall logic. I'll follow up later to make suggestions with inline comments.

eikenb:
Sorry for the delay. I thought I had a couple quick suggestions but as I dug in I decided there was more to it. I've been looking  at it as I can, but with limited time it might take me a bit. Just wanted to stay in touch and let you know I didn't just forget about it. Thanks.

soopsio:
@eikenb Looking forward to your better suggestions. Thanks

eikenb:
So I think I'm just going to merge this as is. It works and doesn't conflict with the other 2 extended packet related PRs. I'm going to pull in all 3 and then get things working and rework from there.  I'm going to do the merge locally as there will be a few conflicts that I'd prefer to resolve locally vs. githuts tool.

eikenb:
And thanks again for the PR and your patience.

eikenb:
Merged.
I rebased before merging, which changed the ids and so it didn't auto-close.

PR #301, opened 2019-08-19
sftp: support rename extension for server

kardianos:
Previously if a client makes an unsupported operation,
like a POSIX rename, it would exit the server.
    
Both support POSIX rename, and do not abort the connection
if there is an unsupported operation is made by the client.

### DIFF
--- a/request-server.go
+++ b/request-server.go
@@ -174,6 +174,15 @@ func (rs *RequestServer) packetWorker(
 				request = NewRequest("Stat", request.Filepath)
 				rpkt = request.call(rs.Handlers, pkt)
 			}
+		case *sshFxpExtendedPacket:
+			switch expkt := pkt.SpecificPacket.(type) {
+			default:
+				rpkt = statusFromError(pkt, ErrSshFxOpUnsupported)
+			case *sshFxpExtendedPacketPosixRename:
+				request := NewRequest("Rename", expkt.Oldpath)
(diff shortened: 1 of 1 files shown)


kardianos:
@eikenb Let me know if something else is needed. Previous behavior terminated loop (closed connection) when client requested an extension when using `sftp.Handlers` implementation. Now it will return an error if not implemented (not terminate loop) or in the case of openssh rename, just handle it like a normal rename.

Also I might add a go.mod PR if you don't mind.

kardianos:
Ping.

eikenb:
Hey @kardianos, thanks for the PR.

I haven't had a lot of time for this project lately, but I'm trying to make some time. I'll look at this soon.

eikenb:
Actually, I'm already looking at this. Realized when reading through #299 that we didn't handle extended packets in either server and was looking into fixing that.

eikenb:
So I think I'm just going to merge this. It works and doesn't conflict (much) with the other 2 extended packet related PRs. I'm going to pull in all 3 and then get things working and go from there. I'm going to do the merge locally as there will be a few conflicts that I'd prefer to resolve locally vs. githuts tool. Thanks.

eikenb:
Merged.
I rebased before merging, which changed the ids and so it didn't auto-close.

kardianos:
Thanks! (I understand about having little time :) )

PR #302, opened 2019-08-26
Announce the posix-rename extension in the server.

### LINKED ISSUE #298: Extensions are not announced by the server
While working on  #290, I discovered that the version package sent by the server isn't announcing any extensions. E.g. `posix-rename@openssh.com`.

This here: https://github.com/pkg/sftp/blob/79ae07e7783e194ec8f46029774d8c4966a79ff4/server.go#L159
should contain extensions:

```
rpkt = sshFxVersionPacket{
	Version: sftpProtocolVersion,
	Extensions: []sshExtensionPair{
		{"posix-rename@openssh.com", "1"},
	},
}
```

tommie:
Closes #298.

### DIFF
--- a/server.go
+++ b/server.go
@@ -160,6 +160,7 @@ func handlePacket(s *Server, p orderedRequest) error {
 			Version: sftpProtocolVersion,
 			Extensions: []sshExtensionPair{
 				{"hardlink@openssh.com", "1"},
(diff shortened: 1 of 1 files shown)


eikenb:
request-server.go should probably contain something similar. But I need to look that over first. But this addresses the need for server.go.

PR #303, opened 2019-08-27
define supported extensions in one place and use for both server and …

drakkan:
…request-server


eikenb:
Thanks for the PR @drakkan. Definitely makes sense to have these 1 place and include in both servers... at least while we have both servers (eventual plan is to re-write the old filesystem based server using an expanded request based backend).

PR #304, opened 2019-08-27
fix lint issues

drakkan:
This commit fix various lint issues, for example:

- don't use underscores in Go name
- don't use ALL_CAPS in Go name; use CamelCase
- errors should have a name of the form errFoo
- method newOrderId should be newOrderID
- comments on exported methods that don't start with the method name
- error strings should not  be capitalized or end with punctuation or a newline

there are still lint issues for the constants here:

https://github.com/pkg/sftp/blob/master/request-errors.go#L10

for example `ErrSshFxOk` should be `ErrSSHFxOk` but this seems a breaking change,

and probably fixing the lint issues here:

https://github.com/pkg/sftp/blob/master/sftp.go#L12

could be a breaking change too, please let me know what do you think about this pull requests before making additional changes, thanks!


eikenb:
Hey @drakkan, thanks for another PR.

Having given it a quick skim a couple thoughts. I think getting more idiomatic naming is a good idea with one caveat. That anything public needs to maintain backward compatibility. It can either just be left as is, or you can add new/proper name but keep the old ones and make sure they both work (should be able to swap one for the other without issue).

eikenb:
Regarding the questions...

> there are still lint issues for the constants here:
> https://github.com/pkg/sftp/blob/master/request-errors.go#L10
> for example ErrSshFxOk should be ErrSSHFxOk but this seems a breaking change,

Those are public. So they need to stay.

> and probably fixing the lint issues here:
> https://github.com/pkg/sftp/blob/master/sftp.go#L12
> could be a breaking change too, please let me know what do you think about this pull requests before making additional changes, thanks!

Those are all private, so they shouldn't breaking as long as you change them in use. Am I missing something... why did you think they could be breaking?

Thanks.

client.go
```diff
@@ -15,12 +15,12 @@ import (
...
 )
 
-// InternalInconsistency indicates the packets sent and the data queued to be
+// ErrInternalInconsistency indicates the packets sent and the data queued to be
 // written to the file don't match up. It is an unusual error and usually is
 // caused by bad behavior server side or connection issues. The error is
 // limited in scope to the call where it happened, the client object is still
 // OK to use as long as the connection is still open.
-var InternalInconsistency = errors.New("internal inconsistency")
+var ErrInternalInconsistency = errors.New("internal inconsistency")
```
eikenb: This is public, so it can't be changed without breaking backwards compatibility. It either needs to be kept as is or you could make the old name an alias for the new.

client.go
```diff
@@ -953,7 +953,7 @@ func (f *File) WriteTo(w io.Writer) (int64, error) {
 
 		if inFlight == 0 {
 			if firstErr.err == nil && len(pendingWrites) > 0 {
-				return copied, InternalInconsistency
+				return copied, ErrInternalInconsistency
```
eikenb: If you make it an alias above, then this should be OK.

request-attrs.go
```diff
@@ -28,23 +28,23 @@ func (r *Request) Pflags() FileOpenFlags {
 	return newFileOpenFlags(r.Flags)
 }
 
-// Flags that indicate whether SFTP file attributes were passed. When a flag is
+// FileAttrFlags that indicate whether SFTP file attributes were passed. When a flag is
 // true the corresponding attribute should be available from the FileStat
 // object returned by Attributes method. Used with SetStat.
 type FileAttrFlags struct {
-	Size, UidGid, Permissions, Acmodtime bool
+	Size, UIDGid, Permissions, Acmodtime bool
```
eikenb: Public field in public struct. This is going to have to stay for compatibility reasons.

server_integration_test.go
```diff
@@ -71,8 +71,8 @@ var sftpServerDebugStream = ioutil.Discard
 var sftpClientDebugStream = ioutil.Discard
 
 const (
-	GOLANG_SFTP  = true
-	OPENSSH_SFTP = false
+	GolangSFTP  = true
+	OpenSSHSFTP = false
```
eikenb: More public consts... in this case it'd be fine to add the new ones, but we can't get rid of the old ones wihtout breaking compatibility. Aliases would work again.

server_integration_test.go
```diff
@@ -468,8 +468,8 @@ func runSftpClient(t *testing.T, script string, path string, host string, port i
 }
 
 func TestServerCompareSubsystems(t *testing.T) {
-	listenerGo, hostGo, portGo := testServer(t, GOLANG_SFTP, READONLY)
-	listenerOp, hostOp, portOp := testServer(t, OPENSSH_SFTP, READONLY)
+	listenerGo, hostGo, portGo := testServer(t, GolangSFTP, READONLY)
+	listenerOp, hostOp, portOp := testServer(t, OpenSSHSFTP, READONLY)
```
eikenb: New names are fine to use internally as they are constants.

server_integration_test.go
drakkan: we need aliases for server_integration_test?

request-attrs.go
drakkan: ok, this lint issue remains

client.go
drakkan: done

drakkan:
patch updated, please let me know if I missed something, thanks

server_integration_test.go
eikenb: Opps. No. Tests don't need to worry about this. Was rushing a bit I guess. Thanks.

request-errors.go
```diff
@@ -1,40 +1,52 @@
...
-	ErrSshFxOk               = fxerr(ssh_FX_OK)
-	ErrSshFxEof              = fxerr(ssh_FX_EOF)
-	ErrSshFxNoSuchFile       = fxerr(ssh_FX_NO_SUCH_FILE)
-	ErrSshFxPermissionDenied = fxerr(ssh_FX_PERMISSION_DENIED)
-	ErrSshFxFailure          = fxerr(ssh_FX_FAILURE)
-	ErrSshFxBadMessage       = fxerr(ssh_FX_BAD_MESSAGE)
-	ErrSshFxNoConnection     = fxerr(ssh_FX_NO_CONNECTION)
-	ErrSshFxConnectionLost   = fxerr(ssh_FX_CONNECTION_LOST)
-	ErrSshFxOpUnsupported    = fxerr(ssh_FX_OP_UNSUPPORTED)
+	ErrSshFxOk               = ErrSSHFxOk
```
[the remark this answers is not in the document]
puellanivis: Hm, this linting advise is weird. It should be `ErrSSHFxOK`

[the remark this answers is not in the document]
eikenb: Yeah... I find many of houndci-bot's recommendations to be questionable. I was thinking of disabling it.

eikenb:
I think everything is good and we can always fix anything we missed later. I want to get this merged to get the conflicts it might cause out of the way and avoid more in the future.

So here it goes...

PR #307, opened 2019-09-07
notify errors to readerAt and writerAt

### LINKED ISSUE #306: sftp server: detection upload errors
Hi, 

I would like to detect upload errors on the sftp server, for example I kill the sftp client process while an upload is ongoing or I stop the upload, client side, using CTRL+C.

In the server `Filewrite` handler I have a pointer to the `sftp.Request` but `request.Context().Err()` seems always `nil`.

Is this supported? If it is not currently supported and you give me some directions I can try to send a pull request, thanks!

drakkan:
Add an optional interface that readerAt and writerAt can implement
to be notified about errors

Fixes #306


drakkan:
as alternate approach we could do a patch like this:

```
diff --git a/request-server.go b/request-server.go
index 5599151..5c3a7a0 100644
--- a/request-server.go
+++ b/request-server.go
@@ -133,6 +133,20 @@ func (rs *RequestServer) Serve() error {
        // make sure all open requests are properly closed
        // (eg. possible on dropped connections, client crashes, etc.)
        for handle, req := range rs.openRequests {
+               if err != nil {
+                       req.state.RLock()
+                       writer := req.state.writerAt
+                       reader := req.state.readerAt
+                       req.state.RUnlock()
+                       if t, ok := writer.(TransferError); ok {
+                               debug("notify error: %v to writer: %v\n", err, writer)
+                               t.TransferError(err)
+                       }
+                       if t, ok := reader.(TransferError); ok {
+                               debug("notify error: %v to reader: %v\n", err, reader)
+                               t.TransferError(err)
+                       }
+               }
                delete(rs.openRequests, handle)
                req.close()
        }
diff --git a/request.go b/request.go
index 3786461..ba684db 100644
--- a/request.go
+++ b/request.go
@@ -32,6 +32,12 @@ type Request struct {
        cancelCtx context.CancelFunc
 }
 
+// TransferError is an optional interface that the readerAt or the writerAt
+// can implement to be notified about packet receive errors
+type TransferError interface {
+       TransferError(err error)
+}
+
 type state struct {
        *sync.RWMutex
        writerAt io.WriterAt
```

this works too and should catch more errors. Please let me know what do you think about. This approach seems better to me.

Please note that if we kill the sftp client while downloading a file from the server the error is alway `io.EOF` as for normal connections close, so if we use code such as this one:

https://github.com/pkg/sftp/blob/master/examples/request-server/main.go#L124

`server.Close()` will never be called in this case. Calling `server.Close()` multiple times seems to work fine (tested on Linux only), so maybe we could modify the request server example so `server.Close()` is always called. Can you please clarify how it should work?

puellanivis:
So, this provides the ability to transfer errors to a WriterAt/ReaderAt, but does not actually implement that with any of the WriterAt/ReaderAt in the package?

drakkan:
> So, this provides the ability to transfer errors to a WriterAt/ReaderAt, but does not actually implement that with any of the WriterAt/ReaderAt in the package?

This is correct, I tested my modifications using sftpgo that has WriterAt and ReaderAt that implements the new interface:

https://github.com/drakkan/sftpgo/blob/master/sftpd/transfer.go#L41

I was unable to simulate these errors using the sftp client in this package since there isn't an abort method, I tested manually killing `sftp` command line tool while uploading/downloading.

eikenb:
Hey @drakkan, thanks for this.

I like the use of an interface for adding the functionality. This is precisely my plan for expanding the API to implement the full server functionality as needed by the filesystem based backend (server.go, which I plan to re-implement w/ the request-server eventually).

The only change I'd like to see (well, I'd like tests but first we need to be able to replicate the error..) is to have the interface moved into the request-interfaces.go file.

Thanks.

PR #309, opened 2019-09-29
fix: received packet too long

### LINKED ISSUE #308: SFTP access blocked and eat too many memory when received message too long
First set some output, like `echo abc`, under the access user's `.bashrc`, then using sftp to access that user will reproduce the problem.

Go pprof:

```sh
ROUTINE ======================== github.com/pkg/sftp.recvPacket in /Users/wutz/.go/path/pkg/mod/github.com/pkg/sftp@v1.10.1/packet.go
    1.52GB     1.52GB (flat, cum) 99.94% of Total
         .          .    141:	var b = []byte{0, 0, 0, 0}
         .          .    142:	if _, err := io.ReadFull(r, b); err != nil {
         .          .    143:		return 0, nil, err
         .          .    144:	}
         .          .    145:	l, _ := unmarshalUint32(b)
    1.52GB     1.52GB    146:	b = make([]byte, l)
         .          .    147:	if _, err := io.ReadFull(r, b); err != nil {
         .          .    148:		debug("recv packet %d bytes: err %v", l, err)
         .          .    149:		return 0, nil, err
         .          .    150:	}
         .          .    151:	if debugDumpRxPacketBytes {
```

In the openssh project, sftp-client will be check received message whether greater than SFTP_MAX_MSG_LENGTH (256 * 1024 )

> https://github.com/openssh/openssh-portable/blob/master/sftp-client.c#L154-L158

wutz:
Close #308

### DIFF
--- a/packet.go
+++ b/packet.go
@@ -13,11 +13,13 @@ import (
 )
 
 var (
+	errLongPacket            = errors.New("packet too long")
 	errShortPacket           = errors.New("packet too short")
 	errUnknownExtendedPacket = errors.New("unknown extended packet")
 )
 
 const (
+	maxMsgLength           = 256 * 1024
 	debugDumpTxPacket      = false
 	debugDumpRxPacket      = false
 	debugDumpTxPacketBytes = false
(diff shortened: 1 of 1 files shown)


eikenb:
As recvPacket is used by both the client and server I wanted to double check this was OK for the server (as you seemed to be focused on the client). Checking openssh, they use the same defined constant in the server code to check message length.

With that checked and the fact that the tests are happy with it... I think this logic should be  fine. And the code is good as well. I think I just talked myself into merging it. :)

PR #311, opened 2019-10-09
Bug with sub folders/files of a renamed/removed path

danfletcher1:
When removing or renaming a folder, sub files/folders of the path were not being updated. I just created a loop to update anything with a matching path.

### DIFF
--- a/request-example.go
+++ b/request-example.go
@@ -93,10 +93,8 @@ func (fs *root) Filecmd(r *Request) error {
 		delete(fs.files, r.Filepath)
 
 		if file.IsDir() {
-			fmt.Printf("Rename Dir %v, %v, %v\n", r.Target, file.name, r.Filepath)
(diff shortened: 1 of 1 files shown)


eikenb:
Hey @danfletcher1, thanks for the patch. 

Looks good though I'm not sure about the remove behavior. Most SFTP servers (and most filesystems) won't delete a directory when it has content and I think that is more correct. It should return an error when trying to remove a directory that still has content.

It doesn't do that either... if you'd like to change remove to return an error, otherwise just remove that part. I'd have it return an os.PathError type... mimicing os.Remove()'s behaviour.

danfletcher1:
Thanks for the advice. I will change that.

Kind regards
Dan

danfletcher1:
Applied your advice.

eikenb:
Looks good. Thanks for the changes.

PR #312, opened 2019-10-11
fix nil panic error

LeeEirc:
the `shFxpExtendedPacket` field `SpecificPacket `will be nil, as its function `UnmarshalBinary`   got an err , when sftp could not recognize `ExtendedRequest` string.


eikenb:
Hey @LeeEirc, thanks for the fix... I appreciate your taking the time to submit a PR.

PR #313, opened 2019-10-23
Bug: windows specific issue on request-server.go

danfletcher1:
I would like you to review the suggestions please.

Both came from practical issues I had using the package. 

The first problem with the Dir() function, like I said in the commit its more specifically clean() function in the Dir() function. It assumes the filesystem separator based on the OS. As its specified as in the Linux format the cleanup starts replacing / with \

The second issue I came about when renaming things. Previous code from an much older commit, didn't have any rename issues, so i'm unclear how this problem came in. Also I've undone some other edit, which might be solving another issue. I'm not certain this is the correct solution but it solves in my instance, i'd appreciate review on that.

### DIFF
--- a/request-example.go
+++ b/request-example.go
@@ -56,7 +56,7 @@ func (fs *root) Filewrite(r *Request) (io.WriterAt, error) {
 	defer fs.filesLock.Unlock()
 	file, err := fs.fetch(r.Filepath)
 	if err == os.ErrNotExist {
-		dir, err := fs.fetch(filepath.Dir(r.Filepath))
+		dir, err := fs.fetch(dir(r.Filepath))
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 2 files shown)


puellanivis:
Why not https://golang.org/pkg/path/#Dir ?

drakkan:
Regarding the rename issue I noticed it in sftpgo too, it happens for example using `sftp` CLI but it works with filezilla.

The problem is that the server announces the `rename` extension and so `sftp` CLI send a posix rename that does not work, filezilla sends a normal rename and it works.

I wonder if:

1) we can make sftpExtensions (https://github.com/pkg/sftp/blob/master/sftp.go#L88) public so an application can customize them
2) we need to send  a "PosixRename" request and not a "Rename" when posix rename is requested from a client

eikenb:
As thing progress I'd like to have some way to handle extended packets without each having their own case in that particular switch statement (it is already kind of long)... but lets get this fixed.

And thanks for pulling that out and adding the test... makes it easier to just merge. :)

eikenb:
@danfletcher1  The above comment was supposed to be on #316 but I put it here and merged this one by mistake. I don't seem to be able to un-merge it in github as well. 

#316 has the posix-rename fix and kept your attribution, but it left out the `Dir` changes. If you'd like to re-submit the `Dir` changes in a new PR I'll look at that again. I was going to ask you to split them anyways as I like 1 PR to be for 1 thing.

Sorry for the hassle.

PR #315, opened 2019-10-26
make SFTP extensions configurable

drakkan:
This patch allow to configure the supported extensions

### DIFF
--- a/sftp.go
+++ b/sftp.go
@@ -85,10 +85,14 @@ const (
 	sshFxfExcl   = 0x00000020
 )
 
-var sftpExtensions = []sshExtensionPair{
-	{"hardlink@openssh.com", "1"},
-	{"posix-rename@openssh.com", "1"},
-}
+var (
+	// supportedSFTPExtensions defines the supported extensions
+	supportedSFTPExtensions = []sshExtensionPair{
+		{"hardlink@openssh.com", "1"},
+		{"posix-rename@openssh.com", "1"},
+	}
+	sftpExtensions = supportedSFTPExtensions
+)
 
 type fxp uint8
 
@@ -226,4 +230,30 @@ func (s *StatusError) Error() string {
 // FxCode returns the error code typed to match against the exported codes
 func (s *StatusError) FxCode() fxerr {
 	return fxerr(s.Code)
+}
+
+func getSupportedExtensionByName(extensionName string) (sshExtensionPair, error) {
+	for _, supportedExtension := range supportedSFTPExtensions {
+		if supportedExtension.Name == extensionName {
+			return supportedExtension, nil
+		}
+	}
+	return sshExtensionPair{}, fmt.Errorf("Unsupported extension: %v", extensionName)
+}
+
+// SetSFTPExtensions allows to customize the supported server extensions.
+// See the variable supportedSFTPExtensions for supported extensions.
+// This method accepts a slice of sshExtensionPair names for example 'hardlink@openssh.com'.
+// If an invalid extension is given an error will be returned and nothing will be changed
+func SetSFTPExtensions(extensions ...string) error {
+	tempExtensions := []sshExtensionPair{}
+	for _, extension := range extensions {
+		sftpExtension, err := getSupportedExtensionByName(extension)
+		if err != nil {
+			return err
+		}
+		tempExtensions = append(tempExtensions, sftpExtension)
+	}
+	sftpExtensions = tempExtensions
+	return nil
 }
--- a/sftp_test.go
+++ b/sftp_test.go
@@ -26,3 +26,50 @@ func TestErrFxCode(t *testing.T) {
 		assert.Equal(t, statusErr.FxCode(), tt.fx)
 	}
 }
+
+func TestSupportedExtensions(t *testing.T) {
+	for _, supportedExtension := range supportedSFTPExtensions {
+		_, err := getSupportedExtensionByName(supportedExtension.Name)
+		assert.NoError(t, err)
+	}
+	_, err := getSupportedExtensionByName("invalid@example.com")
+	assert.Error(t, err)
+}
+
+func TestExtensions(t *testing.T) {
+	var supportedExtensions []string
+	for _, supportedExtension := range supportedSFTPExtensions {
+		supportedExtensions = append(supportedExtensions, supportedExtension.Name)
+	}
+
+	testSFTPExtensions := []string{"hardlink@openssh.com"}
+	expectedSFTPExtensions := []sshExtensionPair{
+		{"hardlink@openssh.com", "1"},
+	}
+	err := SetSFTPExtensions(testSFTPExtensions...)
+	assert.NoError(t, err)
+	assert.Equal(t, expectedSFTPExtensions, sftpExtensions)
+
+	invalidSFTPExtensions := []string{"invalid@example.com"}
+	err = SetSFTPExtensions(invalidSFTPExtensions...)
+	assert.Error(t, err)
+	assert.Equal(t, expectedSFTPExtensions, sftpExtensions)
+
+	emptySFTPExtensions := []string{}
+	expectedSFTPExtensions = []sshExtensionPair{}
+	err = SetSFTPExtensions(emptySFTPExtensions...)
+	assert.NoError(t, err)
+	assert.Equal(t, expectedSFTPExtensions, sftpExtensions)
+
+	// if we only have an invalid extension nothing will be modified.
+	invalidSFTPExtensions = []string{
+		"hardlink@openssh.com",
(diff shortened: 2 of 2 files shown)


sftp.go
```diff
@@ -227,3 +231,25 @@ func (s *StatusError) Error() string {
 func (s *StatusError) FxCode() fxerr {
 	return fxerr(s.Code)
 }
+
+func isSFTPExtensionSupported(extension SSHExtensionPair) bool {
+	for _, supportedExtension := range supportedSFTPExtensions {
```
puellanivis: Why not a `map` rather than having to iterate through the entire list of supported extensions?

drakkan: well `sftpExtensions` must be a slice, so I created `supportedSFTPExtensions` as a slice too so we can easily define:

```
sftpExtensions = supportedSFTPExtensions
```

We can define a map too but this way we duplicate the values, or we can create a map only and then iterate the map to populate `sftpExtensions`. I have no preference, please let me know how I have to modify the patch, thanks!

Am I missing something?

eikenb: IMO keeping it a slice right now is fine as that is required for the current packet setup. It does mean having to loop over , but it is and probably always will be a very short list so I don't think converting it will make much difference.

If it gets annoying we can change the sshFxVersionPacket struct to hold a map instead of a slice and convert things over.

eikenb:
@drakkan Thanks for the submission and I think it is a good idea. I'll be happy to merge it though there is one thing... Could you add a test or 2 covering the new functions?

drakkan:
@eikenb, sorry for the delay. I added some basic test cases, please let me know if they are enough.

sftp_test.go
```diff
@@ -26,3 +26,34 @@ func TestErrFxCode(t *testing.T) {
 		assert.Equal(t, statusErr.FxCode(), tt.fx)
 	}
 }
+
+func TestSupportedExtensions(t *testing.T) {
+	for _, supportedExtension := range supportedSFTPExtensions {
+		assert.True(t, isSFTPExtensionSupported(supportedExtension))
+	}
```
puellanivis: At least one negative test would be a good idea.

sftp_test.go
```diff
@@ -26,3 +26,34 @@ func TestErrFxCode(t *testing.T) {
...
+}
+
+func TestExtensions(t *testing.T) {
+	testSFTPExtensions := []SSHExtensionPair{
+		{"hardlink@openssh.com", "1"},
+	}
+	err := SetSFTPExtensions(testSFTPExtensions)
+	assert.Nil(t, err)
```
puellanivis: `assert.NoError` is a stronger assertion than `assert.Nil`.

sftp_test.go
```diff
@@ -26,3 +26,34 @@ func TestErrFxCode(t *testing.T) {
...
+	assert.Equal(t, testSFTPExtensions, sftpExtensions)
+	invalidSFTPExtensions := []SSHExtensionPair{
+		{"invalid@openssh.com", "1"},
```
puellanivis: Use `example.com` instead of `openssh.com` for this invalid case, in order to give a strong guarantee that this will not name-collide with a possible future extension.

sftp_test.go
```diff
@@ -26,3 +26,34 @@ func TestErrFxCode(t *testing.T) {
...
+	}
+	err = SetSFTPExtensions(invalidSFTPExtensions)
+	assert.NotNil(t, err)
```
puellanivis: `assert.Error` is the stronger assertion here.

sftp.go
```diff
@@ -227,3 +231,25 @@ func (s *StatusError) Error() string {
...
+// SetSFTPExtensions allows to customize the supported server extensions.
+// See the variable supportedSFTPExtensions for supported extensions.
+// If an invalid extension is given an error will be returned
+func SetSFTPExtensions(extensions []SSHExtensionPair) error {
+	for _, extension := range extensions {
+		if !isSFTPExtensionSupported(extension) {
+			return fmt.Errorf("Unsupported extension, name: %v data: %v", extension.Name, extension.Data)
+		}
+	}
+	sftpExtensions = extensions
```
puellanivis: 😮 I just realized something bad could happen here.

Please include this test in your PR, and then fix your code so that it passes:
```
func TestSetSFTPExtensions_IsNotMutable(t *testing.T) {
	inputExtensions := []SSHExtensionPair{
		{"hardlink@openssh.com", "1"},
	}

	// It is important that this be a separate slice, as we mutate inputExtensions below.
	expectExtensions := []SSHExtensionPair{
		{"hardlink@openssh.com", "1"},
	}

	err := SetSFTPExtensions(inputExtensions)
	assert.NoError(t, err)
	assert.Equal(t, expectExtensions, sftpExtensions)

	inputExtensions[0].Data = 42
	assert.Equal(t, expectExtensions, sftpExtensions)
}
```

drakkan: should be fixed, thanks!

eikenb:
@drakkan You don't have to apologize to me for delays... I'm more guilty than anyone on that front around here.

@puellanivis Thanks for reviewing that. Very much appreciated.

eikenb:
@drakkan Thanks for the tests and I just had one more thought (sorry)...

I was looking it over again and wondered if it might not be better if we kept SSHExtensionPair private and made SetSFTPExtensions take a list of the names they wanted. Changing the signature to...
```
func SetSFTPExtensions(extensions ...string) error {
...
```
And calling it...
```
SetSFTPExtensions("hardlink@openssh.com", "posix-rename@openssh.com")
```
Seems like this would both keep the API as is and seems like it be easier to call. 

What do you think?

puellanivis:
I considered that as well, I was not sure how important the “Data” field of the extension pair was. If it’s just ”always the value 1” then it makes sense.

drakkan:
@eikenb @puellanivis please let me know if there are any remaining fixes/improvements, thanks!

sftp.go
```diff
@@ -227,3 +231,40 @@ func (s *StatusError) Error() string {
...
+	var tempExtensions []sshExtensionPair
+	for _, extension := range extensions {
+		sftpExtension, err := getSupportedExtensionByName(extension)
+		if err != nil {
+			return err
+		}
+		tempExtensions = append(tempExtensions, sftpExtension)
+	}
+	sftpExtensions = make([]sshExtensionPair, len(extensions))
+	copy(sftpExtensions, tempExtensions)
```
puellanivis: You do  not need to copy the slice anymore, as you are already building a new slice in `tempExtensions`.

sftp.go
```diff
@@ -227,3 +231,40 @@ func (s *StatusError) Error() string {
 func (s *StatusError) FxCode() fxerr {
 	return fxerr(s.Code)
 }
+
+// useful for tests
+func getEnabledExtensionByName(extensionName string) (sshExtensionPair, error) {
```
puellanivis: This function seems to be _only_ used for testing. If it is going to be limited to tests, then it should be in the appropriate `*_test.go` file, rather than in non-test code.

I don’t think we should change the way the `assert.Equal` compares against the `sftpExtensions` as an array from how you had it before.

Especially, because otherwise the expectation would be that this function itself should be tested.

drakkan:
@puellanivis removed slice copy, thanks

sftp.go
drakkan: > This function seems to be _only_ used for testing. If it is going to be limited to tests, then it should be in the appropriate `*_test.go` file, rather than in non-test code.

ok I can move it to tests or replace it with a function that allows to get the enabled extensions as []string. This method could be useful to to library users too and maybe we can make it public.

> 
> I don’t think we should change the way the `assert.Equal` compares against the `sftpExtensions` as an array from how you had it before.

the input is now `[]string` while the extenstions are `[]sshExtensionPair` we need the method explained above to get the enabled extensions as `[]string` to compare as before.

Please let me know what you prefer, thanks!

> 
> Especially, because otherwise the expectation would be that this function itself should be tested.

puellanivis: Internal to the library, they should be using the `sftpExtensions` array directly.

The tests can test the value of `sftpExtensions` directly against an `expectedExtensions := []sshExtensionPair{…}` complex literal.

eikenb:
> I considered that as well, I was not sure how important the “Data” field of the extension pair was. If it’s just ”always the value 1” then it makes sense.

It may not always be "1", but it will always be fixed... at least for Openssh extensions. They spec is very vague about what it is used for, only specifying that both fields (name, data) must be present though the data one may be blank. Openssh uses the data field as an incremental version number, and the later versions supersede  the former. 

So it might be something other than "1", but it will still be a fixed setting. In other words, I don't have any plans on supporting multiple versions of a specific extension protocol.

PR #316, opened 2019-10-27
Fix posixrename and add a test case

drakkan:
I extracted the relevant parts from https://github.com/pkg/sftp/pull/313 keeping the author and added a test case

### DIFF
--- a/request-server_test.go
+++ b/request-server_test.go
@@ -241,9 +241,20 @@ func TestRequestRename(t *testing.T) {
 	err = p.cli.Rename("/foo", "/bar")
 	assert.Nil(t, err)
 	f, err := r.fetch("/bar")
+	if err != nil {
+		t.Fatal("unexpected error:", err)
+	}
 	assert.Equal(t, "bar", f.Name())
-	assert.Nil(t, err)
 	_, err = r.fetch("/foo")
+	assert.Equal(t, os.ErrNotExist, err)
+	err = p.cli.PosixRename("/bar", "/baz")
+	assert.Nil(t, err)
+	f, err = r.fetch("/baz")
+	if err != nil {
+		t.Fatal("unexpected error:", err)
+	}
+	assert.Equal(t, "baz", f.Name())
+	_, err = r.fetch("/bar")
 	assert.Equal(t, os.ErrNotExist, err)
 }
 
--- a/request-server.go
+++ b/request-server.go
@@ -190,15 +190,10 @@ func (rs *RequestServer) packetWorker(
 				request = NewRequest("Stat", request.Filepath)
 				rpkt = request.call(rs.Handlers, pkt)
 			}
-		case *sshFxpExtendedPacket:
-			switch expkt := pkt.SpecificPacket.(type) {
-			default:
-				rpkt = statusFromError(pkt, ErrSSHFxOpUnsupported)
-			case *sshFxpExtendedPacketPosixRename:
-				request := NewRequest("Rename", expkt.Oldpath)
-				request.Target = expkt.Newpath
-				rpkt = request.call(rs.Handlers, pkt)
-			}
+		case *sshFxpExtendedPacketPosixRename:
+			request := NewRequest("Rename", pkt.Oldpath)
+			request.Target = pkt.Newpath
+			rpkt = request.call(rs.Handlers, pkt)
 		case hasHandle:
 			handle := pkt.getHandle()
 			request, ok := rs.getRequest(handle)


request-server_test.go
```diff
@@ -245,6 +245,13 @@ func TestRequestRename(t *testing.T) {
 	assert.Nil(t, err)
 	_, err = r.fetch("/foo")
 	assert.Equal(t, os.ErrNotExist, err)
+	err = p.cli.PosixRename("/bar", "/foo")
+	assert.Nil(t, err)
+	f, err = r.fetch("/foo")
+	assert.Equal(t, "foo", f.Name())
```
puellanivis: What happens if `f == nil`?

Ok, so,  I know the code just above at line 244 has the same issue, but it crashes with a nil-pointer deref:
```
--- FAIL: TestRequestRename (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x7a1b05]

goroutine 254 [running]:
testing.tRunner.func1(0xc00012d000)
	/usr/lib64/go/1.12/src/testing/testing.go:830 +0x392
panic(0x82d740, 0xc21d50)
	/usr/lib64/go/1.12/src/runtime/panic.go:522 +0x1b5
github.com/pkg/sftp.(*memFile).Name(...)
	/home/snowgirl/Work/github.com/pkg/sftp/request-example.go:257
github.com/pkg/sftp.TestRequestRename(0xc00012d000)
	/home/snowgirl/Work/github.com/pkg/sftp/request-server_test.go:244 +0x275
testing.tRunner(0xc00012d000, 0x8bc8e8)
	/usr/lib64/go/1.12/src/testing/testing.go:865 +0xc0
created by testing.(*T).Run
	/usr/lib64/go/1.12/src/testing/testing.go:916 +0x35a
exit status 2
FAIL	github.com/pkg/sftp	0.013s
```

We really should rework the tests so that in the case that when we’re testing `f.Name()` that we are first ensuring that `if err != nil { t.Fatal(…) }` happens. The `testify/require` package does this. But I tend to much prefer being explicit about this fatal condition case.

In fact, every single one of the `assert.Nil(t, err)` here should be `require.NoError(t, err)` or `if err != nil { t.Fatal("unexpected error:", err) }` Because an error anywhere along the line will cause a cascading failure.

Speaking of which, for good testing sanitation, we should probably use `err = p.cli.PosixRename("/bar", "/baz")` to avoid the repeat of a variable value, just in case.

drakkan: Thanks for the review, the `f == nil` crash in test case should be fixed now.

Regarding "the repeat of a variable value" I don't understand what you mean sorry, why using `p.cli.PosixRename("/bar", "/baz")` is better than `p.cli.PosixRename("/bar", "/foo")` ? 

Can you please clarify? Thanks!

puellanivis: We started the “file” off with the name `/foo`, then we rename it to `/bar`. Renaming back to `/foo` could potentially obscure an error in renaming to `/bar`.

Renaming to `/baz` gives a third filename, that could not be there as a result of a bad test earlier in the process.

Essentially, if both renames failed, then at least some of the tests of state in the second rename would report as passing, even though the underlying functionality was broken. When I say “variable” I mean a test variable, like in a science experiment. We set it up to switch back and forth between the values here, when we really want to make sure that three distinct states have been gone through.

drakkan: Thanks, should be better now

request-server_test.go
```diff
@@ -241,10 +241,23 @@ func TestRequestRename(t *testing.T) {
...
 	_, err = r.fetch("/foo")
 	assert.Equal(t, os.ErrNotExist, err)
+	err = p.cli.PosixRename("/bar", "/baz")
+	assert.Nil(t, err)
+	f, err = r.fetch("/baz")
+	if err != nil {
+		t.Fatal("unexpected error:", err)
+	}
+	assert.Equal(t, "baz", f.Name())
+	assert.Nil(t, err)
```
puellanivis: You don’t need the `assert.Nil` anymore, when you have the `if err != nil { t.Fatal(…) }`  :wink:

drakkan: fixed, thanks!

eikenb:
As thing progress I'd like to have some way to handle extended packets without each having their own case in that particular switch statement (it is already kind of long)... but lets get this fixed.

And thanks for pulling that out and adding the test... makes it easier to just merge. :)

I mistakenly merged the other one (went to wrong tab)... but I reverted it and will merge this one.

PR #326, opened 2019-12-15
WIP: add support for check-file extension

drakkan:
This patch add support for file hashing extension

https://tools.ietf.org/html/draft-ietf-secsh-filexfer-extensions-00#section-3

This is useful to check hash of remote files, for example `rclone` use md5sum/sha1sum system commands over SSH to check files hash, it could use this extension instead.

I tested using the request-server example and this python script (this patch is needed https://github.com/paramiko/paramiko/pull/1563)

```
import paramiko
import os

local_file_path = "/tmp/test_file.txt"
remote_file_path = os.path.basename(local_file_path)

local_file = open(local_file_path,"w")
local_file.write("test content\n")
local_file.close()

transport = paramiko.Transport(("127.0.0.1", 2022))
transport.connect(None, "testuser", "tiger", None)
sftpclient = paramiko.SFTPClient.from_transport(transport)
sftpclient.put(local_file_path,remote_file_path)
dirlist = sftpclient.listdir(".")
print(dirlist)
f=sftpclient.open(remote_file_path)
print(f.check("md5",0,0,0))

sftpclient.close()
```

This patch is missing client side code and test cases. Before finishing this work I would like to know if this is an acceptable addition for pkg/sftp since it supports v3 protocol and this extension comes from v6 protocol. Since this is an extension maybe we can reasonably support it backported.

The extension should be announced as `supported2` extension (v6 protocol) so I should avoid to announce it for now or we should advertise v6 protocol server version.

Please let me know what do you think about this patch. If acceptable I could add support for other file xfer extensions too.

### DIFF
--- a/packet.go
+++ b/packet.go
@@ -842,6 +842,26 @@ func (p *sshFxpDataPacket) UnmarshalBinary(b []byte) error {
 	return nil
 }
 
+type sshFxpExtendedPacketCheckFileReply struct {
+	ID uint32
+	CheckFileResponse
+}
+
+func (p sshFxpExtendedPacketCheckFileReply) id() uint32 { return p.ID }
+
+func (p sshFxpExtendedPacketCheckFileReply) MarshalBinary() ([]byte, error) {
+	l := 1 + 4 + // type(byte) + uint32
+		len(p.HashAlgoUsed) +
+		len(p.Hash)
+
+	b := make([]byte, 0, l)
+	b = append(b, sshFxpExtendedReply)
+	b = marshalUint32(b, p.ID)
+	b = marshalString(b, p.HashAlgoUsed)
+	b = append(b, p.Hash...)
+	return b, nil
+}
+
 type sshFxpStatvfsPacket struct {
 	ID   uint32
 	Path string
@@ -937,6 +957,10 @@ func (p *sshFxpExtendedPacket) UnmarshalBinary(b []byte) error {
 		p.SpecificPacket = &sshFxpExtendedPacketPosixRename{}
 	case "hardlink@openssh.com":
 		p.SpecificPacket = &sshFxpExtendedPacketHardlink{}
+	case "check-file-handle":
+		p.SpecificPacket = &sshFxpExtendedPacketCheckFileHandle{}
+	case "check-file-name":
+		p.SpecificPacket = &sshFxpExtendedPacketCheckFileName{}
 	default:
 		return errors.Wrapf(errUnknownExtendedPacket, "packet type %v", p.SpecificPacket)
 	}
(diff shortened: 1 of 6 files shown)


packet.go
```diff
@@ -1020,3 +1044,64 @@ func (p sshFxpExtendedPacketHardlink) respond(s *Server) responsePacket {
...
+}
+
+func (p sshFxpExtendedPacketCheckFile) id() uint32     { return p.ID }
+func (p sshFxpExtendedPacketCheckFile) readonly() bool { return true }
+func (p *sshFxpExtendedPacketCheckFile) unmarshalBinary(b []byte) (string, error) {
+	var err error
+	var pathOrHandle string
+	if p.ID, b, err = unmarshalUint32Safe(b); err != nil {
+		return pathOrHandle, err
+	} else if p.ExtendedRequest, b, err = unmarshalStringSafe(b); err != nil {
```
puellanivis: None of these `else`s are necessary. The other path will always terminate, so nothing is gained by putting every following statement in an `else` clause.

request-example.go
```diff
@@ -69,6 +75,48 @@ func (fs *root) Filewrite(r *Request) (io.WriterAt, error) {
 	return file.WriterAt()
 }
 
+func (fs *root) CheckFile(r *Request) (CheckFileResponse, error) {
+	response := CheckFileResponse{}
+	requestedAlgos := r.CheckFile.GetHashAlgoAsList()
+	var h hash.Hash
+	for _, algo := range requestedAlgos {
+		if algo == "md5" {
```
puellanivis: This should be a `switch … case` especially since every condition is `algo == A_VALUE`.

request-example.go
```diff
@@ -69,6 +75,48 @@ func (fs *root) Filewrite(r *Request) (io.WriterAt, error) {
...
+		} else if algo == "sha256" {
+			h = sha256.New()
+			response.HashAlgoUsed = "sha256"
+		} else if algo == "sha384" {
+			h = sha512.New384()
+			response.HashAlgoUsed = "sha384"
+		} else if algo == "sha512" {
+			h = sha512.New()
+			response.HashAlgoUsed = "sha512"
+		}
```
puellanivis: And what if `default`?

Ah… you handle it below with the `h == nil`. Wouldn’t it be more clear to put it as an actual case in a `switch`? Then, if we come out of the `switch` we will know that `h` must be valid.

request-example.go
```diff
@@ -69,6 +75,48 @@ func (fs *root) Filewrite(r *Request) (io.WriterAt, error) {
...
 }
 
+func (fs *root) CheckFile(r *Request) (CheckFileResponse, error) {
+	response := CheckFileResponse{}
+	requestedAlgos := r.CheckFile.GetHashAlgoAsList()
+	var h hash.Hash
+	for _, algo := range requestedAlgos {
+		if algo == "md5" {
+			h = md5.New()
+			response.HashAlgoUsed = "md5"
```
puellanivis: It looks like `response.HashAlgoUsed = algo` is invariant to each branch.

request-example.go
```diff
@@ -69,6 +75,48 @@ func (fs *root) Filewrite(r *Request) (io.WriterAt, error) {
 	return file.WriterAt()
 }
 
+func (fs *root) CheckFile(r *Request) (CheckFileResponse, error) {
+	response := CheckFileResponse{}
```
puellanivis: The proper way to declare the zero value of a struct is `var response CheckFileResponse`.

request-example.go
```diff
@@ -69,6 +75,48 @@ func (fs *root) Filewrite(r *Request) (io.WriterAt, error) {
...
+	requestedAlgos := r.CheckFile.GetHashAlgoAsList()
+	var h hash.Hash
+	for _, algo := range requestedAlgos {
```
puellanivis: I do not understand the purpose of ranging through the `requestedAlgos` when you will only ever pick one. Does the standard define that it will only ever respond with a single algorithm in response? If it does, then what does the standard say about how we should go about picking the proper response algorithm?

request-example.go
```diff
@@ -69,6 +75,48 @@ func (fs *root) Filewrite(r *Request) (io.WriterAt, error) {
...
+		} else if algo == "sha512" {
+			h = sha512.New()
+			response.HashAlgoUsed = "sha512"
+		}
+	}
+	if h == nil {
+		return response, errors.New("Unsupported hash algorithm requested")
+	}
+	// in this example we only support the full file hash
+	if r.CheckFile.StartOffset > 0 || r.CheckFile.Length > 0 || r.CheckFile.BlockSize > 0 {
```
puellanivis: Nothing about `r.CheckFile.Length == 0` or `r.CheckFile.BlockSize == 0` says to me that it is a full-file hash. In such a case, you want to document what the standard says and why.

request-example.go
```diff
@@ -69,6 +75,48 @@ func (fs *root) Filewrite(r *Request) (io.WriterAt, error) {
...
+		return response, errors.New("Only full file hash is supported")
+	}
+	file, err := fs.fetch(r.Filepath)
+	if err != nil {
+		return response, err
```
puellanivis: Why are we returning a `response` where? Oh. You aren’t returning a pointer type to the response.

Return a pointer-type to the response, so you can `return nil, err` here, and make it clear that the struct being returned is invalid.

request-example.go
```diff
@@ -69,6 +75,48 @@ func (fs *root) Filewrite(r *Request) (io.WriterAt, error) {
...
+	}
+	h.Write(file.content)
+	response.Hash = append(response.Hash, []byte(fmt.Sprintf("%x", h.Sum(nil)))...)
+
+	return response, nil
```
puellanivis: This could be:
```
return &CheckFileResponse{
  HashAlgoUsed: hashAlgoUsed,
  Hash:         []byte(fmt.Sprintf("%x", h.Sum(nil))),
}, nil
```

request.go
```diff
@@ -15,16 +16,43 @@ import (
...
+type CheckFileRequest struct {
+	HashAlgoList string
+	StartOffset  uint64
+	Length       uint64
+	BlockSize    uint32
+}
+
+// GetHashAlgoAsList returns the comma separataed HashAlgoList string as slice
+func (r CheckFileRequest) GetHashAlgoAsList() []string {
+	result := []string{}
```
puellanivis: https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices

Use `var result []string` to get an empty list.

request.go
```diff
@@ -15,16 +16,43 @@ import (
...
-	Attrs    []byte // convert to sub-struct
-	Target   string // for renames and sym-links
-	handle   string
+	Method    string
+	Filepath  string
+	Flags     uint32
+	Attrs     []byte // convert to sub-struct
+	Target    string // for renames and sym-links
+	handle    string
+	CheckFile CheckFileRequest // for CheckFile requests
```
puellanivis: Not sure why this couldn’t be in its own separate “paragraph” and not cause a reflow of all the other fields’ whitespace.

drakkan:
@puellanivis thanks for your review. I can improve the patch as you suggest but before doing more work here I would like to know if this patch could be accepted or I need to add full v6 protocol support too

eikenb:
Hey @drakkan, thanks for the contribution.

I'm definitely open to this. After an initial skim my only hesitation is the additional handler and backwards compatibility. I've thought about this before and am thinking about adding/changing how that works.

My idea is to make it so that the handlers can be defined on handlers object itself (instead of as fields on the struct) and to assert the different supported types. This could then be a way to extend the interfaces supported without requiring additional Handers fields and the additional extensions could all be optional.

This moves toward how I'm thinking about extending the system as is to support the rest of the functionality that would be required to fully implement the spec. I'm working towards getting a new release out right now, but after I'll be looking into this.

drakkan:
@eikenb thanks for you suggestions. 

Meantime I tested this WIP with other clients and it does not work for example with WinSCP, to make it work with WinSCP we have to announce the extension and so we need to be v6 compatible.

Currently I'm using my free time to add s3/cloud support to sftpgo (thaks for pipeat, very useful!), 

I'll try to understand what changes are required to be v6 compatible later and if I can I'll try to send a pull request. 

I'll try to use proftpd code for inspiration: for example proftpd's mod_sftp reads client version and announces a compatibile server version, for example if client version is 6 it sets the server version to 6 and announces the v6 extensions it supports

drakkan:
This could be closed, eventually I'll refactor it in future

PR #329, opened 2019-12-30
fix S_IFMT value on windows, plan9 and js/wasm

### LINKED ISSUE #291: Wrong value of S_IFMT mask when converting sftp filemode bits to os.FileMode
I encountered this issue when synchronizing a directory using rclone (with sftp) to a server running AIX. Rclone was sending a SSH_FXP_STAT request to test if a directory is, well, a directory.

The STAT request was giving the directory's status as 'not a directory' despite being one.
The problem was in the toFileMode function in attrs.go

```
input filemode bits are                  1 0100 0011 1111 1111‬
syscall.S_IFMT mask bits are             1 1111 0000 0000 0000
```

The different values tested are:

> 

```
 S_IFBLK                                 0 0110 0000 0000 0000
 S_IFCHR                                 0 0010 0000 0000 0000
 S_IFDIR                                 0 0100 0000 0000 0000
 S_IFIFO                                 0 0001 0000 0000 0000
 S_IFLNK                                 0 1010 0000 0000 0000
 S_IFREG                                 0 1000 0000 0000 0000
 S_IFSOCK                                0 1100 0000 0000 0000
```

As you can see, none of the values matchs the input, and the leftmost bit is not used by any value. 
The input would match S_IFDIR if the S_IFMT mask didn't test the leftmost bit.

I have searched the value of S_IFMT mask in other sftp implentations and in the ones I have found, they define S_IFMT as: 1111 0000 0000 0000

for example, the python implementation of stat:
https://github.com/python/cpython/blob/2.7/Lib/stat.py

Or this javascript implementation of sftp:
http://maverick-ng-javadocs.s3-website-eu-west-1.amazonaws.com/constant-values.html#com.sshtools.client.sftp.SftpFileAttributes.S_IFMT

eikenb:
Go defines S_IFMT on windows, plan9 and js/wasm as 0x1f000 instead of 0xf000. None of the the other S_IFxyz values include the "1" (in 0x1f000) which prevents them from matching the bitmask.

This fixes that by overriding the S_IFMT value on the effected platforms to be 0xf000, as it it on all the others.

Fixes #291

### DIFF
--- a/syscall_fixed.go
+++ b/syscall_fixed.go
@@ -1,0 +1,9 @@
+// +build plan9 windows js,wasm
+
+// Go defines S_IFMT on windows, plan9 and js/wasm as 0x1f000 instead of
+// 0xf000. None of the the other S_IFxyz values include the "1" (in 0x1f000)
+// which prevents them from matching the bitmask.
+
+package sftp
+
+const S_IFMT = 0xf000
--- a/attrs.go
+++ b/attrs.go
@@ -16,8 +16,8 @@ const (
 	sshFileXferAttrACmodTime   = 0x00000008
 	sshFileXferAttrExtented    = 0x80000000
 
-	sshFileXferAttrAll = sshFileXferAttrSize|sshFileXferAttrUIDGID|sshFileXferAttrPermissions|
-		sshFileXferAttrACmodTime|sshFileXferAttrExtented
+	sshFileXferAttrAll = sshFileXferAttrSize | sshFileXferAttrUIDGID | sshFileXferAttrPermissions |
+		sshFileXferAttrACmodTime | sshFileXferAttrExtented
 )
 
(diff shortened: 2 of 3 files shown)


syscall_fixed.go
```diff
@@ -0,0 +1,5 @@
+// +build plan9 windows
```
puellanivis: We probably also want this for `js,wasm`:

 https://github.com/golang/go/blob/master/src/syscall/syscall_js.go#L159

eikenb: done.

eikenb:
I added js/wasm, updated the commit message and added a comment to the _fixed version noting the reasoning.

eikenb:
Thanks.

I had to double check the build tags as well. First time I had a that logic combo.

PR #330, opened 2020-01-02
Issue 295 corrupt resumes

### LINKED ISSUE #295: sftp resume produces corrupted files
I'm writing an SFTP server using this great library, I have a problem with uploads resume.

In my app I'm using `sftp.NewRequestServer` but the problem is reproducible also with the simple sftp example here 

https://github.com/pkg/sftp/blob/master/examples/go-sftp-server/main.go

please try this:

```
sftp -P 2022 testuser@127.0.0.1
sftp> put <path to some large file>
```

interrupt the transfer using CTRL+C and then resume the upload

```
sftp> reput <path to some large file>
```
the uploaded file has the same size as the original one but a different checksum

get/reget works as expected.

thanks

eikenb:
Test case and fix for using sftp's append mode.

Fixes #295

### DIFF
--- a/server_integration_test.go
+++ b/server_integration_test.go
@@ -649,6 +649,53 @@ func TestServerPut(t *testing.T) {
 	}
 }
 
+func TestServerResume(t *testing.T) {
+	listenerGo, hostGo, portGo := testServer(t, GolangSFTP, READONLY)
+	defer listenerGo.Close()
+
+	tmpFileLocal := "/tmp/" + randName()
+	tmpFileRemote := "/tmp/" + randName()
+	defer os.RemoveAll(tmpFileLocal)
+	defer os.RemoveAll(tmpFileRemote)
+
+	t.Logf("put: local %v remote %v", tmpFileLocal, tmpFileRemote)
+
+	// create a local file with random contents to be pushed to the server
+	tmpFileLocalData := randData(2 * 1024 * 1024)
+	// only write half the data to simulate a split upload
+	half := 1024 * 1024
+	err := ioutil.WriteFile(tmpFileLocal, tmpFileLocalData[:half], 0644)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// sftp the first half of the file to the server
+	output, err := runSftpClient(t, "put "+tmpFileLocal+" "+tmpFileRemote,
+		"/", hostGo, portGo)
+	if err != nil {
+		t.Fatalf("runSftpClient failed: %v, output\n%v\n", err, output)
+	}
+
+	// write the full file out
+	err = ioutil.WriteFile(tmpFileLocal, tmpFileLocalData, 0644)
+	if err != nil {
+		t.Fatal(err)
+	}
+	// re-sftp the full file with the append flag set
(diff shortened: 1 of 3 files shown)


server_integration_test.go
```diff
@@ -649,6 +649,53 @@ func TestServerPut(t *testing.T) {
 	}
 }
 
+func TestServerResume(t *testing.T) {
+	listenerGo, hostGo, portGo := testServer(t, GolangSFTP, READONLY)
+	defer listenerGo.Close()
+
+	tmpFileLocal := "/tmp/" + randName()
```
puellanivis: Any reason to not use https://golang.org/pkg/io/ioutil/#TempFile ?

server_integration_test.go
```diff
@@ -649,6 +649,53 @@ func TestServerPut(t *testing.T) {
...
+	output, err = runSftpClient(t, "put -a "+tmpFileLocal+" "+tmpFileRemote,
+		"/", hostGo, portGo)
+	if err != nil {
+		t.Fatalf("runSftpClient failed: %v, output\n%v\n", err, output)
+	}
+
+	// tmpFileRemote should now exist, with the same contents
+	if tmpFileRemoteData, err := ioutil.ReadFile(tmpFileRemote); err != nil {
+		t.Fatal(err)
+	} else if string(tmpFileLocalData) != string(tmpFileRemoteData) {
```
puellanivis: `bytes.Equal` might be shorter than casting both to string, just to do the compare?

server_integration_test.go
```diff
@@ -649,6 +649,53 @@ func TestServerPut(t *testing.T) {
...
+	}
+	// re-sftp the full file with the append flag set
+	output, err = runSftpClient(t, "put -a "+tmpFileLocal+" "+tmpFileRemote,
+		"/", hostGo, portGo)
+	if err != nil {
+		t.Fatalf("runSftpClient failed: %v, output\n%v\n", err, output)
+	}
+
+	// tmpFileRemote should now exist, with the same contents
+	if tmpFileRemoteData, err := ioutil.ReadFile(tmpFileRemote); err != nil {
```
puellanivis: While go does prefer use of preconditional initializations in `if` blocks, in this particular case, it’s recommended to avoid it, because it’s causing the much more ugly redundant codepath `else` use below.

server_integration_test.go
eikenb: This and all the below are due to my having cut-n-pasted the TestServerPut test to make this. Many other tests have these same issues and I'm not sure if correcting them in one place is better than leaving them as is for consistency.

I think I'll leave this commit as is (after adding the documentation updates I forgot) but make a quick pass fixing these sorts of things where I see them in a separate commit. Do you think that commit would be better included in this PR or out on its own?

eikenb:
@drakkan. You are 100% correct that the fix is to just remove the O_APPEND flag and nothing else. When using WriteAt() the current file offset is ignored, so that seek is pointless. It is entirely up to the client to send the correct offset.

I'll re-work this PR to reflect that.

Thanks for catching that.

server_integration_test.go
puellanivis: Sounds good to just clean them all up in one pass together.

Issue #331, opened 2020-01-03
File.WriteTo implementation moved to Stat instead of FStat breaks IBM Sterling SFTP


erans:
Due to #288 it seems the code of File.WriteTo was changed to use Stat instead of Fstat. That causes problems with IBM Sterling SFTP servers when extractability is set to 1.

It throws the error:
```The message [{FILENAME}] is not extractable!\" (SSH_FX_FAILURE)```

Debugging it further shows that if a certain configuration (which is very common for such SFTP servers in financial institutions) is set calling Stat after opening a file will cause it to freak out as it thinks someone tries to open/download more than 1 file at the same time.

Removing the call for Stat and just pulling the file until there are no more bytes to get seems to solve this.

We are working on a PR that will contain a ClientOption flag that will allow to control this behaviour, hopefully not breaking other servers by supporting both of these strange behaviours :-)

erans:
Here is the pull request to allow both behaviours:
https://github.com/pkg/sftp/pull/332

eikenb:
Hey @erans, thanks for the report  and sorry about the regression. SFTP servers and clients are all over the place in what/how they support and its an ongoing wack-a-mole to get them all happy.

I'll take a look at your PR...

eikenb:
Forgot to add that the PR fixes this. Merged the PR so closing this as well.

PR #332, opened 2020-01-03
Added a ClientOption to determine whetehr to use Fstat or Stat when F…

erans:
…ile.WriteTo is being called to support strange behaviour on some servers. See #331

### DIFF
--- a/client.go
+++ b/client.go
@@ -51,6 +51,23 @@ func MaxPacketChecked(size int) ClientOption {
 	}
 }
 
+// UseFstat sets whether to use Fstat or Stat when File.WriteTo is called (usually when copying files).
+// Some servers limit the amount of open files and calling Stat after opening the file will throw an error
+// From the server. Setting this flag will call Fstat instead of Stat which is suppose to be called on an
+// open file handle.
+//
+// From our experience it happens often with IBM Sterling SFTP servers that have "extractability" level
+// set to 1 which means only 1 file can be open at any given time.
+//
+// If the server you are working with still has an issue with both Stat and Fstat calls you can always
+// open a file and just loop through multiple Read calls until the file is completely read.
+func UseFstat(value bool) ClientOption {
+	return func(c *Client) error {
+		c.useFstat = value
+		return nil
+	}
+}
+
 // MaxPacketUnchecked sets the maximum size of the payload, measured in bytes.
 // It accepts sizes larger than the 32768 bytes all servers should support.
 // Only use a setting higher than 32768 if your application always connects to
@@ -161,6 +178,7 @@ type Client struct {
 	maxPacket             int // max packet size read or written.
 	nextid                uint32
 	maxConcurrentRequests int
+	useFstat              bool
 }
 
 // Create creates the named file mode 0666 (before umask), truncating it if it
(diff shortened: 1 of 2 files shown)


eikenb:
Looks good. Nice documentation.

A couple requests.
1. Could you wrap the comments/documentation lines at 80 characters.
2. Could you add a simple test in client_test.go,  sort of like are there for the MaxPacketChecked, et. al. 

Thanks @erans.

client.go
```diff
@@ -51,6 +51,23 @@ func MaxPacketChecked(size int) ClientOption {
 	}
 }
 
+// UseFstat sets whether to use Fstat or Stat when File.WriteTo is called (usually when copying files).
+// Some servers limit the amount of open files and calling Stat after opening the file will throw an error
+// From the server. Setting this flag will call Fstat instead of Stat which is suppose to be called on an
+// open file handle.
+//
+// From our experience it happens often with IBM Sterling SFTP servers that have "extractability" level
```
puellanivis: “our experience“ is a dangling reference. It will not be clear once checked in who “we” is. Turning this into “It was found that this happens often with…” should be enough, unless you would like to expand fully who figured it out.

client.go
```diff
@@ -51,6 +51,23 @@ func MaxPacketChecked(size int) ClientOption {
...
+// set to 1 which means only 1 file can be open at any given time.
+//
+// If the server you are working with still has an issue with both Stat and Fstat calls you can always
+// open a file and just loop through multiple Read calls until the file is completely read.
```
puellanivis: We should probably recommend `io.Copy(ioutil.Discard, file)` as a work around, which is semantically more clear of purpose, rather than recommending a looped Read.

client.go
```diff
@@ -913,15 +931,27 @@ func (f *File) Read(b []byte) (int, error) {
 // maximise throughput for transferring the entire file (especially
 // over high latency links).
 func (f *File) WriteTo(w io.Writer) (int64, error) {
-	fi, err := f.c.Stat(f.path)
-	if err != nil {
-		return 0, err
+	var err error
+	fileSize := uint64(0)
```
puellanivis: `var fileSize uint64` is more clear of purpose.

client.go
```diff
@@ -913,15 +931,27 @@ func (f *File) Read(b []byte) (int, error) {
...
+	if f.c.useFstat {
+		var fileStat *FileStat
+		if fileStat, err = f.c.fstat(f.handle); err != nil {
```
puellanivis: Because you’re going to continue using the `fileStat` variable, it’s way better to use the `:=` short declaration and initialization outside, rather than using this very unwieldy declaration and precondition.

So:
```
fs, err := f.c.fstat(f.handle)
if err != nil {
  return 0, err
}
fileSize = fs.Size // casting is unnecessary, as the types already match
```

Basically, the overriding interest of avoiding explicit type declarations outweighs the interest in putting things into an initializing precondition.

client.go
```diff
@@ -913,15 +931,27 @@ func (f *File) Read(b []byte) (int, error) {
...
+			return 0, err
+		}
+		fileSize = uint64(fileStat.Size)
+
+	} else {
+		var fi os.FileInfo
+		if fi, err = f.c.Stat(f.path); err != nil {
```
puellanivis: Same as above, it is far more unwieldy to use an initializing precondition here, rather than just using the short declaration `:=` operator.

client.go
erans: The problem is that some SFTP servers, specifically those on mainframes, have the notion that once a file has been completely read it gets deleted, so doing `io.Copy(ioutil.Discard, file)` will get you the size, but the SFTP server will assume you have read it all and delete it :-|

client.go
erans: Thanks. Will update.

client.go
puellanivis: Makes sense and 🤦‍♀ of course it does that. Although, I’m not saying we should do that in order to find the file size, but rather replace the comment here suggesting “loop through multiple Read calls until the file is completely read.”

erans:
I see. I do want to say what to do when both Stat and FStat fails. It would
save people a whole lot of time :)

I will rephrase it better.

erans:
I've updated the code based off the comments + added a small test. Thanks all!

PR #337, opened 2020-02-14
server integration test: add /usr/lib/ssh/sftp-server to search paths

drakkan:
os some distro, for example ArchLinux, the sftp-server CLI is in this
path


eikenb:
Thanks!

PR #340, opened 2020-02-25
bump ssh dep to fix vuln

twmb:
https://groups.google.com/forum/#!topic/golang-dev/bhd3Qn9_Q1k

Via `go get -u`, so includes one more day of commits rather than the exact commit referenced in the link.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,5 +6,5 @@ require (
 	github.com/kr/fs v0.1.0
 	github.com/pkg/errors v0.8.1
 	github.com/stretchr/testify v1.4.0
-	golang.org/x/crypto v0.0.0-20190820162420-60c769a6c586
+	golang.org/x/crypto v0.0.0-20200221231518-2aa609cf4a9d
 )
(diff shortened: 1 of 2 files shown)


puellanivis:
The vulnerability seems to be unclear about when the issue was introduced, and just says the 20200220 commit version is vulnerable, and that is not the one we are using.

I’m not saying that we should not update the version, but it would be nice if security vulnerabilities always explained what versions are impacted.

twmb:
This likely affects most backwards versions. [Here'](https://github.com/golang/crypto/commit/bac4c82f69751a6dd76e702d54b3ceb88adab236)s the fix commit.

I've pinned ssh where I need to an appropriate version, so I don't mind if this bump is not merged, but I recommend it be.

puellanivis:
Yeah, `go.mod` intends that transitive dependencies should be updated regardless of what the middle-package says it requires. But this is still a good merge, considering the bug.

Best to not “recommend” a bad version.

eikenb:
Thanks so much @twmb and @puellanivis for fixing this! I was AFK all week at a work function and didn't even see this until now. Really, I appreciate it so much.

PR #342, opened 2020-03-07
fix connection close race

### LINKED ISSUE #333: data race when deferring a Close()
I am doing:

```
[myfunc]
mymutex.Lock()
defer mymutex.Unlock()

sshClient, _, err := s.SSHClient(ctx)
if err != nil {
	return err
}

client, err := sftp.NewClient(sshClient)
if err != nil {
	return err
}
defer client.Close()

[use client]

return err
```

(where `s.SSHClient` is creating/storing/reusing an `*ssh.Client`.)

On rare occasion I get this:

```
WARNING: DATA RACE
Write at 0x00c000c18d20 by goroutine 340:
  golang.org/x/crypto/ssh.(*channel).CloseWrite()
      /home/ubuntu/go/pkg/mod/golang.org/x/crypto@v0.0.0-20191011191535-87dc89f01550/ssh/channel.go:542 +0x5e
  golang.org/x/crypto/ssh.(*sessionStdin).Close()
      /home/ubuntu/go/pkg/mod/golang.org/x/crypto@v0.0.0-20191011191535-87dc89f01550/ssh/session.go:536 +0x55
  github.com/pkg/sftp.(*clientConn).recv.func1()
      /home/ubuntu/go/pkg/mod/github.com/pkg/sftp@v1.10.1/conn.go:71 +0x6c
  github.com/pkg/sftp.(*clientConn).recv()
      /home/ubuntu/go/pkg/mod/github.com/pkg/sftp@v1.10.1/conn.go:77 +0x392
  github.com/pkg/sftp.(*clientConn).loop()
      /home/ubuntu/go/pkg/mod/github.com/pkg/sftp@v1.10.1/conn.go:60 +0x72

Previous write at 0x00c000c18d20 by goroutine 366:
  golang.org/x/crypto/ssh.(*channel).CloseWrite()
      /home/ubuntu/go/pkg/mod/golang.org/x/crypto@v0.0.0-20191011191535-87dc89f01550/ssh/channel.go:542 +0x5e
  golang.org/x/crypto/ssh.(*sessionStdin).Close()
      /home/ubuntu/go/pkg/mod/golang.org/x/crypto@v0.0.0-20191011191535-87dc89f01550/ssh/session.go:536 +0x55
  github.com/pkg/sftp.(*clientConn).Close()
      /home/ubuntu/go/pkg/mod/github.com/pkg/sftp@v1.10.1/conn.go:55 +0x95
  github.com/pkg/sftp.(*Client).Close()
      <autogenerated>:1 +0x43
[...]
```

Both goroutines are inside mutex-protected `myfunc`.

So the issue here is that my own `Close()` call in `myfunc`'s defer is competing with the internal `Close()` call that gets triggered in the event of a `recvPacket()` error.

I'm not sure what I should do about this.

eikenb:
Need to prevent Close from being called at the same time. The lock
already existed, but was only used in one place where it should be
applied generally.

Fixes #333

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -31,6 +31,12 @@ func (c *conn) sendPacket(m encoding.BinaryMarshaler) error {
 	return sendPacket(c, m)
 }
 
+func (c *conn) Close() error {
+	c.Lock()
+	defer c.Unlock()
+	return c.WriteCloser.Close()
+}
+
 type clientConn struct {
 	conn
 	wg         sync.WaitGroup
@@ -67,9 +73,7 @@ func (c *clientConn) loop() {
 // appropriate channel.
 func (c *clientConn) recv() error {
 	defer func() {
-		c.conn.Lock()
 		c.conn.Close()
-		c.conn.Unlock()
 	}()
 	for {
 		typ, data, err := c.recvPacket()


eikenb:
Threw this up as a pull request in case anyone could come up with a test case. I know of no good way to reproduce data races in general so skipped it.

PR #343, opened 2020-03-09
fileget: allocate a slice with enough capacity

drakkan:
so a new allocation is not needed in MarshalBinary and sendPacket.

Here are some profiling results while downloading a file (file size is about 1GB),

before this patch:

```
1254.24MB 55.18% 55.18%  1254.24MB 55.18%  github.com/pkg/sftp.sshFxpDataPacket.MarshalBinary
  991.81MB 43.63% 98.81%   991.81MB 43.63%  github.com/pkg/sftp.fileget
       1MB 0.044% 98.86%  1255.24MB 55.22%  github.com/pkg/sftp.(*packetManager).maybeSendPackets
    0.50MB 0.022% 98.88%  1260.24MB 55.44%  github.com/pkg/sftp.(*packetManager).controller
         0     0% 98.88%   991.81MB 43.63%  github.com/pkg/sftp.(*Request).call
         0     0% 98.88%   993.31MB 43.70%  github.com/pkg/sftp.(*RequestServer).Serve.func1.1
         0     0% 98.88%   993.31MB 43.70%  github.com/pkg/sftp.(*RequestServer).packetWorker
         0     0% 98.88%  1254.24MB 55.18%  github.com/pkg/sftp.(*conn).sendPacket
         0     0% 98.88%  1254.24MB 55.18%  github.com/pkg/sftp.sendPacket
```

with this patch:

```
1209.48MB 98.46% 98.46%  1209.48MB 98.46%  github.com/pkg/sftp.fileget
       2MB  0.16% 98.63%     7.50MB  0.61%  github.com/pkg/sftp.recvPacket
         0     0% 98.63%        8MB  0.65%  github.com/drakkan/sftpgo/sftpd.Configuration.handleSftpConnection
         0     0% 98.63%  1209.48MB 98.46%  github.com/pkg/sftp.(*Request).call
         0     0% 98.63%        8MB  0.65%  github.com/pkg/sftp.(*RequestServer).Serve
         0     0% 98.63%  1209.98MB 98.50%  github.com/pkg/sftp.(*RequestServer).Serve.func1.1
         0     0% 98.63%  1209.98MB 98.50%  github.com/pkg/sftp.(*RequestServer).packetWorker
         0     0% 98.63%     7.50MB  0.61%  github.com/pkg/sftp.(*conn).recvPacket (inline)
```

### DIFF
--- a/packet.go
+++ b/packet.go
@@ -826,6 +826,8 @@ type sshFxpDataPacket struct {
 	Data   []byte
 }
 
+// MarshalBinary encodes the receiver into a binary form and returns the result.
+// To avoid a new allocation the Data slice must have a capacity >= Length + 9
 func (p sshFxpDataPacket) MarshalBinary() ([]byte, error) {
 	b := append(p.Data, make([]byte, 9)...)
 	copy(b[9:], p.Data[:p.Length])
--- a/packet.go
+++ b/packet.go
@@ -820,10 +820,11 @@ type sshFxpDataPacket struct {
 }
 
 func (p sshFxpDataPacket) MarshalBinary() ([]byte, error) {
-	b := []byte{sshFxpData}
-	b = marshalUint32(b, p.ID)
-	b = marshalUint32(b, p.Length)
-	b = append(b, p.Data[:p.Length]...)
+	b := append(p.Data, make([]byte, 9)...)
+	copy(b[9:], p.Data[:p.Length])
+	b[0] = sshFxpData
+	binary.BigEndian.PutUint32(b[1:5], p.ID)
+	binary.BigEndian.PutUint32(b[5:9], p.Length)
 	return b, nil
 }
 
--- a/packet.go
+++ b/packet.go
@@ -586,9 +586,9 @@ func (p *sshFxpReadPacket) UnmarshalBinary(b []byte) error {
 
 func (p *sshFxpReadPacket) getDataSlice() []byte {
 	dataLen := clamp(p.Len, maxTxPacket)
-	// we allocate a slice with a bigger capacity so we avoid a new allocation in MarshalBinary and in sendPacket
-	// we need 9 bytes in MarshalBinary and 4 bytes in sendPacket
-	return make([]byte, dataLen, dataLen+13)
+	// we allocate a slice with a bigger capacity so we avoid a new allocation in sshFxpDataPacket.MarshalBinary
+	// and in sendPacket, we need 9 bytes in MarshalBinary and 4 bytes in sendPacket
+	return make([]byte, dataLen, dataLen+9+4)
 }
 
 type sshFxpRenamePacket struct {
--- a/packet.go
+++ b/packet.go
@@ -584,6 +584,13 @@ func (p *sshFxpReadPacket) UnmarshalBinary(b []byte) error {
 	return nil
 }
 
+func (p *sshFxpReadPacket) getDataSlice() []byte {
+	dataLen := clamp(p.Len, maxTxPacket)
+	// we allocate a slice with a bigger capacity so we avoid a new allocation in MarshalBinary and in sendPacket
+	// we need 9 bytes in MarshalBinary and 4 bytes in sendPacket
+	return make([]byte, dataLen, dataLen+13)
+}
+
 type sshFxpRenamePacket struct {
 	ID      uint32
 	Oldpath string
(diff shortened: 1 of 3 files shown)


drakkan:
Here are the profile results for uploads and downloads

[file_download_profile_results.zip](https://github.com/pkg/sftp/files/4311984/file_download_profile_results.zip)
[file_upload_profile_results.zip](https://github.com/pkg/sftp/files/4311985/file_upload_profile_results.zip)

packet.go
```diff
@@ -584,6 +584,13 @@ func (p *sshFxpReadPacket) UnmarshalBinary(b []byte) error {
 	return nil
 }
 
+func (p *sshFxpReadPacket) getDataSlice() []byte {
+	dataLen := clamp(p.Len, maxTxPacket)
+	// we allocate a slice with a bigger capacity so we avoid a new allocation in MarshalBinary and in sendPacket
+	// we need 9 bytes in MarshalBinary and 4 bytes in sendPacket
+	return make([]byte, dataLen, dataLen+13)
```
puellanivis: Writing this as `+9+4` makes the magic number parallel the code comment more directly, and we don’t have to do arithmetic in our head every time to ensure the numbers line up.

packet.go
```diff
@@ -820,10 +827,11 @@ type sshFxpDataPacket struct {
 }
 
 func (p sshFxpDataPacket) MarshalBinary() ([]byte, error) {
-	b := []byte{sshFxpData}
-	b = marshalUint32(b, p.ID)
-	b = marshalUint32(b, p.Length)
-	b = append(b, p.Data[:p.Length]...)
+	b := append(p.Data, make([]byte, 9)...)
```
puellanivis: Hm… if this triggers a reallocation, then we end up duplicating everything twice.

But I’m not sure if we want to have a conditional in here to avoid that in the case that should not happen.

server.go
```diff
@@ -258,7 +256,7 @@ func handlePacket(s *Server, p orderedRequest) error {
 		f, ok := s.getHandle(p.Handle)
 		if ok {
 			err = nil
-			data := make([]byte, clamp(p.Len, s.maxTxPacket))
+			data := p.getDataSlice()
```
puellanivis: It might be worth mentioning down below at the `Data:` field that we need to _not_ clamp the capacity, as we use it later in the marshalling. e.g. `// do not use data[:n:n] here to clamp the capacity, we allocated extra capacity above to avoid reallocations`

packet.go
drakkan: A reallocation should not happen since we allocated the needed 9 extra bytes before

puellanivis: Exported function though. :wink: 

But yeah, I don’t think it’s probably worth it, because like you said, it’s use is almost exclusively internal, and we allocate enough space.

drakkan: Added some documentation about the required capacity

eikenb:
@drakkan Thanks for your continued work on performance improvements. Very much appreciated. 

Also thanks @puellanivis for the great job reviewing it.

My only comments would have been similar to @puellanivis about explaining why the numbers are what they are. I think the comments @drakkan added do a pretty good job.

Thanks again.

drakkan:
@eikenb thanks, my next pull request will contain the allocator. 

As you can see in the profile results above, pkg/sftp only allocates in 2 places now:

- [here](https://github.com/pkg/sftp/blob/master/packet.go#L591) for downloads
- [here](https://github.com/pkg/sftp/blob/master/packet.go#L142) when receiveing any packets including write packets

and we need to release the allocated slice after sending the packet to the network, [here](https://github.com/pkg/sftp/blob/master/packet-manager.go#L183)
 
It doesn't seem so complex, anyway,obviously, feel free to reject it

puellanivis:
Additional idea to simplify the allocator: why don’t we always allocate the max len slice, rather than a variety of different slice lengths?

In fact, if we just change the current allocations to strictly use the one length, would we see a performance increase before even doing the whole allocator? It’s possible it might simplify garbage collection, and reallocation of backing arrays.

drakkan:
@puellanivis, I evaluated this strategy too, I think it can be used in fileget for downloads, in recvPacket the max allowed packet size is 262144 this seems too much.

As alternative strategy I thought to preallocate a slice for each [rwChan](https://github.com/pkg/sftp/blob/master/packet-manager.go#L111) but we need a way to identify each rwChan and to get this information to use the right slice, I don't know how to do this with the current code (the allocator seems less intrusive), do you have ideas? Additionally recvPacket is not related to rwChan,we receive the packet (allocating a slice) we read the type and then we send to the right channel so for uploads we have an already allocated slice inside rwChan.

Thanks!

PR #344, opened 2020-03-14
add optional allocator

### LINKED ISSUE #334: Performance improvements
Hi,

SFTPGo uses this library for SFTP and an home made implementation for SCP. 

Generally the bottleneck is the encryption, anyway for some ciphers this is not the case, for example `aes128-gcm@openssh.com` has implicit MAC and is really fast, other ciphers can be optimized, for example for AES CTR we can apply this patch:

https://go-review.googlesource.com/c/go/+/51670

and sha256, used for MAC, can be optimized this way:

https://github.com/drakkan/crypto/commit/cc78d71334be7b503d6e71e9ffeed80cdee863d4

Now encryption is not more the bottleneck and SFTPGo' SCP implementation has a speed comparable with OpenSSH, sadly SFTP is still slower than OpenSSH.

I understand that SCP is a simpler protocol than SFTP and my implementation is really basic too: I have only one goroutine that read/write to/from a preallocated []byte, no packet manager or multiple goroutines as in pkg/sftp. 

In SFTPGo if I put this line:

https://github.com/drakkan/sftpgo/blob/master/sftpd/scp.go#L402

inside the for loop I have a 25% performance regression, pkg/sftp seems to do exactly this, looking at pkg/sftp code for each packet  there is at least a line such as this one:

```
data := make([]byte, clamp(length, maxTxPacket))
```

I think we can improve performance using preallocated slices. I'm thinking about something like a global or per request sync.Pool but it does't seem so easy to follow allocation/deallocation in all cases, do you have better suggestions? We could make pkg/sftp as fast as OpenSSH implementation!

Please take a look here for some numbers and profiling results:

https://github.com/drakkan/sftpgo/issues/69

thanks!

drakkan:
after processing a packet we keep in memory the allocated slices and we reuse
them for new packets.
Slices are allocated in:

- recvPacket
- when we receive a sshFxpReadPacket (downloads)

The allocated slices have a fixed size = maxMsgLength.

Allocated slices are referenced to the request order id and are marked for reuse
after a request is served in maybeSendPackets.

The allocator is added to the packetManager struct and it is cleaned at the end
of the Serve() function.

This allocation mode is optional and disabled by default.

I tested several different approachs, please take a look at my branches here:

- https://github.com/drakkan/sftp/tree/allocator
- https://github.com/drakkan/sftp/tree/allocator2
 
the included benchmark show that this is the better approach. Here are some profiling results:

## Upload 1GB file ##

```
Showing nodes accounting for 1116.80MB, 97.66% of 1143.54MB total
Dropped 63 nodes (cum <= 5.72MB)
Showing top 10 nodes out of 16
      flat  flat%   sum%        cum   cum%
 1110.30MB 97.09% 97.09%  1110.30MB 97.09%  golang.org/x/crypto/ssh.(*connectionState).readPacket
       6MB  0.52% 97.62%       10MB  0.87%  golang.org/x/crypto/ssh.Marshal (inline)
    0.50MB 0.044% 97.66%    10.50MB  0.92%  golang.org/x/crypto/ssh.(*channel).adjustWindow
         0     0% 97.66%    17.95MB  1.57%  github.com/drakkan/sftpgo/sftpd.Configuration.handleSftpConnection
         0     0% 97.66%    17.95MB  1.57%  github.com/pkg/sftp.(*RequestServer).Serve
         0     0% 97.66%    15.45MB  1.35%  github.com/pkg/sftp.(*conn).recvPacket (inline)
         0     0% 97.66%        6MB  0.52%  github.com/pkg/sftp.(*packetManager).controller
         0     0% 97.66%    15.45MB  1.35%  github.com/pkg/sftp.recvPacket
         0     0% 97.66%    10.50MB  0.92%  golang.org/x/crypto/ssh.(*channel).Read
         0     0% 97.66%    10.50MB  0.92%  golang.org/x/crypto/ssh.(*channel).ReadExtended
```

## Download 1GB file ##

```
Showing nodes accounting for 34.34MB, 80.94% of 42.43MB total
Showing top 10 nodes out of 131
      flat  flat%   sum%        cum   cum%
   15.84MB 37.34% 37.34%    15.84MB 37.34%  github.com/pkg/sftp.(*allocator).GetPage
       5MB 11.78% 49.12%     6.50MB 15.32%  golang.org/x/crypto/ssh.Marshal
    2.50MB  5.89% 55.01%     4.50MB 10.61%  github.com/pkg/sftp.orderedPackets.Sort
       2MB  4.71% 59.73%        2MB  4.71%  internal/reflectlite.Swapper
       2MB  4.71% 64.44%        2MB  4.71%  github.com/pkg/sftp.(*packetManager).incomingPacket
       2MB  4.71% 69.15%        2MB  4.71%  strings.genSplit
    1.50MB  3.54% 72.69%     1.50MB  3.54%  github.com/pkg/sftp.makePacket
    1.50MB  3.54% 76.22%     8.45MB 19.91%  github.com/pkg/sftp.fileget
       1MB  2.36% 78.58%        1MB  2.36%  golang.org/x/crypto/ssh.(*connectionState).readPacket
       1MB  2.36% 80.94%        1MB  2.36%  github.com/pkg/sftp.(*packetManager).readyPacket
```

Fixes #334


packet.go
```diff
@@ -584,10 +591,13 @@ func (p *sshFxpReadPacket) UnmarshalBinary(b []byte) error {
 	return nil
 }
 
-func (p *sshFxpReadPacket) getDataSlice() []byte {
+func (p *sshFxpReadPacket) getDataSlice(alloc *allocator, orderID uint32) []byte {
 	dataLen := clamp(p.Len, maxTxPacket)
 	// we allocate a slice with a bigger capacity so we avoid a new allocation in sshFxpDataPacket.MarshalBinary
-	// and in sendPacket, we need 9 bytes in MarshalBinary and 4 bytes in sendPacket
+	// and in sendPacket, we need 9 bytes in MarshalBinary and 4 bytes in sendPacket.
+	if alloc != nil {
```
puellanivis: Move this conditional above the comment, as the comment is much more specific to the code below.

packet.go
```diff
@@ -584,10 +591,13 @@ func (p *sshFxpReadPacket) UnmarshalBinary(b []byte) error {
...
+		return alloc.GetPage(orderID)[:dataLen]
```
puellanivis: Should `GetPage` also be returning a slice with capacity of `maxTxPacket+9+4`?

request-server.go
```diff
@@ -107,11 +107,11 @@ func (rs *RequestServer) Serve() error {
 	var pktType uint8
 	var pktBytes []byte
 	for {
-		pktType, pktBytes, err = rs.recvPacket()
+		pktType, pktBytes, err = rs.recvPacket(rs.pktMgr.alloc, rs.pktMgr.getNextOrderID())
```
puellanivis: Here we’re calling a receiver method on `rs` and passing values from the `rs` object as arguments. We shouldn’t do that, the receiver method will already have access to these values.

request-server.go
```diff
@@ -150,6 +150,9 @@ func (rs *RequestServer) Serve() error {
 		delete(rs.openRequests, handle)
 		req.close()
 	}
+	if rs.pktMgr.alloc != nil {
```
puellanivis: This might be more reliable as a `defer`? Just in case someone does an early `return`.

server.go
```diff
@@ -22,6 +22,34 @@ const (
 	SftpServerWorkerCount = 8
 )
 
+// AllocationMode defines allocation modes
+type AllocationMode uint8
```
puellanivis: When defining enums use the type `int` (which should have the most efficient size for the architecture).

Bytes are not any faster to use than `int`s, and rarely save any space as well, due to memory alignment.

server.go
```diff
@@ -138,9 +166,8 @@ func (svr *Server) sftpServerWorker(pktChan chan orderedRequest) error {
 		// If server is operating read-only and a write operation is requested,
 		// return permission denied
 		if !readonly && svr.readOnly {
-			svr.sendPacket(orderedResponse{
-				responsePacket: statusFromError(pkt, syscall.EPERM),
-				orderid:        pkt.orderID()})
+			svr.pktMgr.readyPacket(
+				svr.pktMgr.newOrderedResponse(statusFromError(pkt, syscall.EPERM), pkt.orderID()))
```
puellanivis: I know the original code ends the last line with a `)`, but if one is going to break things up onto separate lines, each argument should go on its own line, ending with a `,`, and then the ending `)` should be on its own line, the same as one should fill in the fields of a struct with `{` and `}`.

request-server.go
```diff
@@ -188,30 +192,30 @@ func (rs *RequestServer) packetWorker(
 				rpkt = statusFromError(pkt, syscall.EBADF)
 			} else {
 				request = NewRequest("Stat", request.Filepath)
-				rpkt = request.call(rs.Handlers, pkt)
+				rpkt = request.call(rs.Handlers, pkt, rs.pktMgr.alloc, orderID)
```
puellanivis: The caller should be able to call `pkt.orderID()` itself, right?

allocator.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+// Free removes all the used and free pages.
+// Call this method when the allocator is not needed anymore
+func (a *allocator) Free() {
+	a.Lock()
+	defer a.Unlock()
+
+	a.available = nil
+	a.used = make(map[uint32][][]byte)
```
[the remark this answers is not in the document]
drakkan: we just forgive improper usage, I can add a comment about this or, if you prefer, I can set it to nil

conn.go
```diff
@@ -18,8 +18,10 @@ type conn struct {
 	sendPacketTest func(w io.Writer, m encoding.BinaryMarshaler) error
 }
 
-func (c *conn) recvPacket() (uint8, []byte, error) {
-	return recvPacket(c)
+// the allocator and the orderID are used in server mode if AllocationModeOptimized is enabled.
+// For the client just pass nil and 0
+func (c *conn) recvPacket(alloc *allocator, orderID uint32) (uint8, []byte, error) {
```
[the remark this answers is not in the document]
drakkan: Ok, I'll add the allocator to the conn struct

server.go
```diff
@@ -22,6 +22,34 @@ const (
...
+const (
+	// AllocationModeStandard a new slice is allocated for each new SFTP packet
+	AllocationModeStandard AllocationMode = iota
+	// AllocationModeOptimized after processing a packet we keep in memory the allocated slices
+	// and we reuse them for new packets. This mode is experimental
+	AllocationModeOptimized
+)
+
+// enabledAllocationMode defines the allocation mode.
+var enabledAllocationMode = AllocationModeStandard
```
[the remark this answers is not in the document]
drakkan: For `Server` with can just add a new option like this:

```
// WithAllocationMode sets the allocation mode.
// Allowed values are:
// - AllocationModeStandard, default
// - AllocationModeOptimized
func WithAllocationMode(mode AllocationMode) ServerOption {
	return func(s *Server) error {
		var alloc *allocator
		if mode == AllocationModeOptimized {
			alloc = newAllocator()
		}
		s.pktMgr.alloc = alloc
		s.conn.alloc = alloc
		return nil
	}
}
```

for RequestServer is ok a new costructor like this one? 

```
// NewRequestServerWithAllocationMode creates/allocates/returns new RequestServer using the
// specified allocation mode.
// Allowed allocation mode are:
//
// - AllocationModeStandard, default
// - AllocationModeOptimized
//
// Normally there will be one server per user-session.
func NewRequestServerWithAllocationMode(rwc io.ReadWriteCloser, h Handlers, mode AllocationMode) *RequestServer 
```

packet.go
```diff
@@ -139,27 +139,34 @@ func sendPacket(w io.Writer, m encoding.BinaryMarshaler) error {
...
 	}
 	l, _ := unmarshalUint32(b)
 	if l > maxMsgLength {
 		debug("recv packet %d bytes too long", l)
 		return 0, nil, errLongPacket
 	}
-	b = make([]byte, l)
-	if _, err := io.ReadFull(r, b); err != nil {
+	if alloc == nil {
+		b = make([]byte, l)
```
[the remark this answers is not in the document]
drakkan: this is `l` not `1`where `l` change based on the previous 4 bytes read

packet.go
drakkan: GetPage returns a slice with capacity = maxMsgLength this is enough to avoid new allocations, I'll add a comment to clarify this

request-server.go
drakkan: You should read this as `rs.serverConn.recvPacket(rs.pktMgr.alloc, rs.pktMgr.getNextOrderID())`

anyway I'll add alloc to conn struct as you suggested before, so this could be simplified in:

```
pktType, pktBytes, err = rs.recvPacket(rs.pktMgr.getNextOrderID())
```

request-server.go
drakkan: pkt is an orderedRequest and so has orderID method, `request.call` requires a requestPacket that cannot call orderID itself

allocator.go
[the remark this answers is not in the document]
puellanivis: 🤷‍♀ I think it’s up to you. Neither choice is going to be a big deal.

server.go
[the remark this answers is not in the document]
puellanivis: If there is a `NewRequestServer` and what is essentially `NewRequestServerWithXYZ(…, xyz bool)`, then better would be to presume that the later call would only be called if `xyz` was going to be true.

But you should still be able to extend `NewRequestServer` as well, so: `NewRequestServer(io.ReadWriteCloser, Handlers, ServerOption...)` as well though, right?

drakkan:
@puellanivis thank you for your review. Please let me know if I need to change other things.

request-server.go
```diff
@@ -33,20 +33,50 @@ type RequestServer struct {
 }
 
 // NewRequestServer creates/allocates/returns new RequestServer.
-// Normally there there will be one server per user-session.
+// Normally there will be one server per user-session.
 func NewRequestServer(rwc io.ReadWriteCloser, h Handlers) *RequestServer {
+	rs, _ := NewRequestServerWithOptions(rwc, h, nil...)
```
puellanivis: You don’t need to specify: `nil...`, you just don’t have to include any options at all.

Is there a reason to split up `NewRequestServer` and `NewRequestServerWithOptions`? Unless we expect that people are assigning this into a variable with a specific type, then there should not be no reason why we couldn’t use `NewRequestServer(io.ReadWriteCloser, Handlers, ...RequestServerOption) (*RequestServer, error)`

drakkan: > You don’t need to specify: `nil...`, you just don’t have to include any options at all.
> 

you are right, sorry for that

> Is there a reason to split up `NewRequestServer` and `NewRequestServerWithOptions`? Unless we expect that people are assigning this into a variable with a specific type, then there should not be no reason why we couldn’t use `NewRequestServer(io.ReadWriteCloser, Handlers, ...RequestServerOption) (*RequestServer, error)`

the reason is backward compatibility, currently `NewRequestServer` does not return an error while `NewServer` returns `(*Server, error)`, the error can be returned there only when setting an option and it can never happen with the current code.
So for `NewRequestServer`:

1) we can break compatibility using the method you suggest
2) introduce a new method as I did
3) don't allow the RequestServerOption to return an error: this should work with current code but could not work in future

puellanivis: Right, but the only actual option available right now, cannot return a non-`nil` error. So, do we really need to build having an `error` return value into the `RequestServerOption`?

drakkan: Ok, done, thanks!

puellanivis:
Wanted to add some feedback on circumstances on my end. I do intend to evaluate the last changes, and you’ve probably covered all the important stuff. But work has been nuts with the COVID stuff, and work from home, and everything has just been quite hectic here, and i have not been able to give this the attention it deserves.

I’m sure I’ll be back to being able to provide the help to get this improvement put in soon.

drakkan:
@puellanivis no problem, take your time and thanks for all your suggestions

drakkan:
The allocator is now [enabled](https://github.com/drakkan/sftpgo/commit/5a5912ea668bb0faeaedc90db08019b9ef0d8689) in SFTPGo. I don't expect any issue, anyway this patch will be now tested by more users. Thanks

puellanivis:
Awesome. I think the code is probably all set here, so with some actual real-world testing, we can then have extra safety on top.

eikenb:
@puellanivis Do you think this is good to merge? I keep meaning to review it but I've just had no time. I had a bit today and did a quick skim review and things looked good. I'll trust your opinion and if you think it is ready, please go ahead and merge it (or let me know and I can do it if you don't want to).

puellanivis:
Yeah, I think any concerns I would have would be lint-like comments anyways. The solid chunk of concerns have all be addressed.

drakkan:
a polite ping here.

I would like to release SFTPGo 1.0.0 before July 15, do you think this patch can be merged before then?
If this is not possibile I'll continue to use my pkg/sftp branch for the release too, thank you!

puellanivis:
Oh, sorry. I was up-to-my-eyeballs in stuff as of late.

drakkan:
> We’ll probably need to release at least a version 1.11.1 right?
> 
> Should we instead do a 1.12? I think we’re reasonably backwards compatible that all users of the code will still work with the same behavior if untouched, right?

Thank you! Yes it should be full backwards compatible. 

Please note that I pushed by mistake #355 to this branch (without the test case) since I included it in SFTPGo 1.0.0, do you want a patch to revert it? Or a patch to add the test case only? Sorry my bad

puellanivis:
Uh, depends on if it’s actually a bug or not. Various parts of the SFTP standard dictate that we should be using POSIX `path` not per-OS `filepath` so… 🤷

eikenb:
> Should we instead do a 1.12?

I think so. Not so much due to compatibility issues but to help highlight all the performance work drakkan contributed.

I'm going to add a 1.12.0 release milestone and add all the relevant issues/prs to it. So we have some semblance of a changelog.

PR #352, opened 2020-05-28
Update example_test.go

eleijonmarck:
Overwriting the package name of the sftp.

Should use client in the example instead


puellanivis:
Good call. Will leave it up for other review over night, and commit tomorrow.

eikenb:
Thanks for this!

PR #354, opened 2020-06-02
Update example_test.go

eleijonmarck:
Add Closing of the file. Had trouble finding the error in which a sftp did not catch my file through this example.

Need to close the file in order for it completely see the newly created file.

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -41,6 +41,7 @@ func Example() {
 	if _, err := f.Write([]byte("Hello world!")); err != nil {
 		log.Fatal(err)
 	}
+	defer f.Close()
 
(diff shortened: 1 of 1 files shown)


example_test.go
```diff
@@ -41,6 +41,7 @@ func Example() {
 	if _, err := f.Write([]byte("Hello world!")); err != nil {
 		log.Fatal(err)
 	}
+	defer f.Close()
```
eleijonmarck, line 44:
```suggestion
	f.Close()
```

eleijonmarck: writing to file is finished. closing file rather than defering

PR #355, opened 2020-06-06
cleanPath: use path.IsAbs after converting ToSlash

drakkan:
we need a POSIX path, filepath.IsAbs could give unexpected results on Windows.

I don't have any specific test case for this anyway I think that filepath.IsAbs could consider an absolute path things like `C:/path` on Windows

### DIFF
--- a/request-server.go
+++ b/request-server.go
@@ -232,7 +232,7 @@ func cleanPacketPath(pkt *sshFxpRealpathPacket) responsePacket {
 // Makes sure we have a clean POSIX (/) absolute path to work with
 func cleanPath(p string) string {
 	p = filepath.ToSlash(p)
-	if !filepath.IsAbs(p) {
+	if !path.IsAbs(p) {
 		p = "/" + p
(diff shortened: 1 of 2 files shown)


puellanivis:
Seems likely, but it’s not immediately clear that this is the more correct behavior. I will take a look sometime (probably next week) and make sure it’s solid before merging.

drakkan:
Hi, sorry for the delay, I just found some time to test on Windows and I added a test case.

As expected without the patch it fails on Windows:

```
 request-server_test.go:426:
             Error Trace:    request-server_test.go:426
             Error:          Not equal:
                             expected: "/C:/a"
                             actual  : "C:/a"

                             Diff:
                             --- Expected
                             +++ Actual
                             @@ -1 +1 @@
                             -/C:/a
                             +C:/a
             Test:           TestCleanPath
```

puellanivis:
Looks legit. The function even says, “a clean POSIX path”.

PR #361, opened 2020-07-15
Remove tautology statement

wn:
Remove tautology statement. Consider the for-loop above, there is no way to break out of the loop unless `err != nil`.

### DIFF
--- a/request-server.go
+++ b/request-server.go
@@ -133,19 +133,17 @@ func (rs *RequestServer) Serve() error {
 	// make sure all open requests are properly closed
 	// (eg. possible on dropped connections, client crashes, etc.)
 	for handle, req := range rs.openRequests {
-		if err != nil {
-			req.state.RLock()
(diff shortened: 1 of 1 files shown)


puellanivis:
Looks pretty legit. At the very least, the value of `err` cannot change in the loop, and as a loop invariant should be tested outside of the loop entirely. But then, as noted, the only way out of the loop above, is if `err != nil`.

wn:
Could i get an LGTM?

puellanivis:
I just spotted a bug, not in your code, but in the code above it. I’ll fix that, and then push the two together.

puellanivis:
Closing with deference to https://github.com/pkg/sftp/pull/363 which fixes a bunch of bugs spotted while reviewing this PR. (They’re not bugs from this PR.)

eikenb:
Thanks for handling this @puellanivis!

PR #362, opened 2020-07-16
Example needs to use client.Close

puellanivis:
When changing the example code to use `client` instead of shadowing the `sftp` package with the client variable, it was missed that we need to call `client.Close()` instead of `sftp.Close()`

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -22,7 +22,7 @@ func Example() {
 	if err != nil {
 		log.Fatal(err)
 	}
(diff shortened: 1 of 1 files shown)


puellanivis:
Would prefer to merge this tomorrow, so that we can release the next version with the optional allocator.

puellanivis:
Defering to https://github.com/pkg/sftp/pull/356 from other author who contributed it before me.

PR #363, opened 2020-07-18
RequestServer.Serve bugs found looking at PR-361

puellanivis:
So, I end up moving the “tautological statement” into a function under `Request` (`transferError(err error)`). The new function better encapsulates behavior of the `Request` object to calls on the `Request` object. So, while the check now still exists, it at least is an appropriate guard statement, guarding against potential misbehavior.

The thing I noticed during PR-361, is that the `break` in the `switch { default: … }` does [not break out of the `for { … }` loop, but rather only out of the `switch`](https://play.golang.org/p/Nh_EgEpNI_f). Meaning, the `break` as it originally appeared was actually a no-op, and execution wasn’t properly ending in that situation.

The solution was to move the `for { … }` loop into its own independent function, where the loop can then be broken clearly and unambiguously with a `return`.

Additionally, I noticed that although we’re looping over `openRequests`, which is supposed to only be accessed under lock, we were not actually holding the lock at that point.

As well, `Request.Close` really should be closing both the `readerAt` and the `writerAt` if they implement, `io.Closer`, even if the first one returns an `err != nil` error.


puellanivis:
No, I don’t think so. There’s a bit of a hope to avoid touching as much as possible, so as little can break as possible. 😆

request.go
```diff
@@ -138,19 +138,49 @@ func (r *Request) close() error {
...
+			err = err2
+		}
+	}
+
+	if c, ok := wr.(io.Closer); ok {
+		if err2 := c.Close(); err == nil {
+			// update error if it is still nil
+			err = err2
+		}
+	}
```
nightlyone, line 161: Could you switch those around? Close errors of writers are sometimes data loss, while close errors or readers are mostly not relevant to the user. So I would prefer to get the close error of the writer here.

request-server.go
```diff
@@ -158,8 +160,7 @@ func (rs *RequestServer) Serve() error {
 
 	err := rs.serveLoop(pktChan)
 
-	close(pktChan) // shuts down sftpServerWorkers
-	wg.Wait()      // wait for all workers to exit
+	wg.Wait() // wait for all workers to exit
```
nightlyone, line 163: Won't this wait forever now? 

The previous sequence seems more correct, since you usually close the channels so workers can detect end of record stream and exit. The you wait for the exit.

request.go
puellanivis: Good point.

request-server.go
puellanivis: The `pktChan` is closed as a `defer` in the `rs.serveLoop(pktChan)`.

request.go
puellanivis: I have inverted both `Close()` and `transferError()` (to keep them consistent in ordering), and specifically added documentation about why the Writer Close goes first.

drakkan:
I just did a quick test and it seems fine for me, all my test cases passed and also manual checking for transfer errors seems ok, thanks!

puellanivis:
:grumble-grumble: merge conflicts when a delete and a delete/insert end up head-to-tail next to each other.

request-server.go
nightlyone: Ah yes, I overlooked somehow that the function has been factored out. So the changes you did are completely correct in this regard.

nightlyone:
@eikenb this seems now good to go. Wanna merge it?

puellanivis:
I’m usually hesitant to commit my own merges if there are no other checks on code. But sure, I think everyone is pretty satisfied that this won’t massively break anything.

PR #365, opened 2020-07-24
update deps

drakkan:
Maybe is a good idea before the next tag :)


puellanivis:
Not necessarily all that important, since dependencies are largely overridden by a projects own `go.mod`, but a good idea regardless. I mean, then we can run our tests against the latest versions, and ensure that they pass.

PR #366, opened 2020-07-25
remove redundant types declaration

drakkan:
.


puellanivis:
I suppose, ideally, we would also change them to use proper field names in assignment, rather than implicit field names. But 🤷‍♀️ this is still a good code polish to make.

PR #369, opened 2020-08-10
Exclude fakeFileInfoSys and testOsSys in non-test builds

tklauser:
These funcs are only used in tests, so rename the files they are defined
in to request_*_test.go and add appropriate build tags where necessary
to exclude them from non-test builds.


request_windows_test.go
```diff
@@ -1,3 +1,5 @@
+// +build windows
```
davecheney, line 1: this line should not be needed.

PR #370, opened 2020-08-10
Exclude test/example code in non-test builds

tklauser:
The fakeFileInfoSys and testOsSys funcs are only used in tests, so
rename the files they are defined in to request_*_test.go to exclude
them from non-test builds.

Also rename request-example.go to request_example_test.go to exclude it
as well, since it uses said funcs in example code.


request_windows_test.go
```diff
@@ -1,3 +1,5 @@
+// +build windows
```
davecheney, line 1: this line should not be needed

tklauser: Done, thanks.

tklauser:
@puellanivis @davecheney Is there anything else that needs to be done before this can be merged? Thanks.

puellanivis:
No, I will merge it now.

tklauser:
Given this currently breaks the build and I currently lack the time to investigate a solution which is not introducing a breaking change I sent PR#375 to revert this one

PR #373, opened 2020-08-20
request server: add support for SSH_FXP_FSETSTAT

drakkan:
we need to add a case for this packet inside the packet worker otherwise
it will be handled in hasHandle case and it will become a "Put" request.

Client side if a Truncate request is called on the open file we should
send a FSETSTAT packet, the request is on the handle, and not a SETSTAT
packet that should be used for paths and not for handle.
We should probably do the same for Chmod/Chtimes/Stat too.

This patch fix file corruption using `sshfs`. Simply create a text file using sshfs and then delete some lines, the file will be corrupted.

See the truncate issue reported [here](https://github.com/drakkan/sftpgo/issues/155) for more details


request_example_test.go
```diff
@@ -302,6 +309,18 @@ func (f *memFile) WriteAt(p []byte, off int64) (int, error) {
 	return len(p), nil
 }
 
+func (f *memFile) Truncate(size int64) error {
+	f.contentLock.Lock()
+	defer f.contentLock.Unlock()
+	nc := make([]byte, size)
```
puellanivis, line 315: We should be able to truncate the byte slice if `size` is shorter, and just `append(f.content, make([]byte, grow)...)` otherwise:

```
grow := size - len(f.content)
if grow <= 0 {
  f.content = f.content[:size]
} else {
  f.content = append(f.content, make([]byte, grow)...)
}
```

This uses some [Slice Tricks](https://github.com/golang/go/wiki/SliceTricks#extend) that the compiler understands, and will turn into “just expand, don’t actually allocate the `make` buffer”. [The source code](https://github.com/golang/go/blob/037ac2bd84480e0a06c4d1e7a2c1f133109466a5/src/cmd/compile/internal/gc/walk.go#L662-L664)

puellanivis:
Hm… interesting. Is this something really critical that we should address before releasing 1.12? File corrupt _does_ seems like a showstopper.

drakkan:
> Hm… interesting. Is this something really critical that we should address before releasing 1.12? File corrupt _does_ seems like a showstopper.

It depends on what the client do. We will have corrupted files if the client use `SSH_FXP_FSETSTAT` to truncate the file. But we have corrupted files also if the client read and write to the same file, see #374

Probably we should also add an optional `Truncate` interface and call it  on the aleady opened file, for example in SFTPGo I have to do something like [this](https://github.com/drakkan/sftpgo/blob/master/common/connection.go#L494) because if a issue an `os.Truncate` on an already opened file I get errors on some OS (for example on Windows). Anyway I can handle this in SFTPGo. Please let me know what do you think about, thank you

request-server_test.go
```diff
@@ -345,6 +345,40 @@ func TestRequestFstat(t *testing.T) {
 	checkRequestServerAllocator(t, p)
 }
 
+func TestRequestFsetstat(t *testing.T) {
+	p := clientRequestServerPair(t)
+	defer p.Close()
+	_, err := putTestFile(p.cli, "/foo", "hello")
+	assert.Nil(t, err)
```
puellanivis, line 352: `require.NoError`. If this errors, we want to stop right here, and now, because nothing else will work.

request-server_test.go
```diff
@@ -345,6 +345,40 @@ func TestRequestFstat(t *testing.T) {
...
+	fp, err := p.cli.OpenFile("/foo", os.O_WRONLY)
+	assert.Nil(t, err)
```
puellanivis, line 354: `require.NoError`

request-server_test.go
```diff
@@ -345,6 +345,40 @@ func TestRequestFstat(t *testing.T) {
...
+	err = fp.Truncate(2)
+	if assert.NoError(t, err) {
```
puellanivis, line 356: Why do we have different behavior on whether there is an error or not?

We should just `require.NoError` and/or `if err != nil { t.Fatal(…) }` any time a precondition fails. There’s no point in running the rest of any of the test if these things fail.

request-server_test.go
```diff
@@ -345,6 +345,40 @@ func TestRequestFstat(t *testing.T) {
...
+	assert.NoError(t, err)
+	assert.Equal(t, 5, n)
+	err = fp.Close()
+	assert.NoError(t, err)
+	rf, err := p.cli.Open("/foo")
+	assert.Nil(t, err)
+	defer rf.Close()
+	contents := make([]byte, 20)
+	n, err = rf.Read(contents)
+	if err != nil && err != io.EOF {
```
puellanivis, line 374: Unless we actually do not know if this will return `err == nil` or `err == io.EOF`, then we should instead write the stronger test of `err != nil` or `err != io.EOF`. (That is, there is no reason to write a test that accepts two possible values, unless both of those values are possible/reasonable/in-contract.)

request-server_test.go
```diff
@@ -345,6 +345,40 @@ func TestRequestFstat(t *testing.T) {
...
+		t.Fatalf("err: %v", err)
```
puellanivis, line 375: `t.Fatal("unexpected error:", err)`

client.go
```diff
@@ -1280,8 +1299,17 @@ func (f *File) Chmod(mode os.FileMode) error {
 // that if the size is less than its current size it will be truncated to fit,
 // the SFTP protocol does not specify what behavior the server should do when setting
 // size greater than the current size.
+// We send a SSH_FXP_FSETSTAT here since we have a file handle
 func (f *File) Truncate(size int64) error {
-	return f.c.Truncate(f.path, size)
+	err := f.c.setfstat(f.handle, sshFileXferAttrSize, uint64(size))
+	if err == nil {
```
puellanivis, line 1305: Note that [`os.File.Truncate`](https://golang.org/pkg/os/#File.Truncate) says: “It does not change the I/O offset.”

An in fact [`man ftruncate`](https://linux.die.net/man/2/ftruncate) also says: “The file offset is not changed.”

Additional information about how to handle “what if we’re past the end of the new file length before a `truncate`? Available here: https://stackoverflow.com/questions/60950996/how-file-offset-of-read-or-write-change-when-file-is-truncated-to-zero-by-ot

Note: a `Read()` when `offset > length` should return zero-bytes read. A `Write()` when `offset > length` should ftruncate to the new offset, before extending further with the write.

drakkan: Oops I missed that thank you. 

The `Read()` and `Write()` test cases are client side tests I didn't include them

drakkan:
Regarding CI I updated travis to go 1.15 however I would switch to GitHub actions in future, Actions are better integrated with GitHub and we cannot merge a pull request without noticing that it breaks build, do you have objections? 

Do you think we can break compatibility after 1.12? I would like to do the following backward incompatible changes:

- don't hide `Fstat`: currently the request server call `Stat` for both `sshFxpLstat` and `sshFxpFstat`
- remove `Put` and `Get` methods and add a generic `OpenFile` method. This should fix file corruption if a client try to read and write to the same handle. Do you think this should be fixed before 1.12?
- add support for golangci-lint (as GitHub action) and so remove all the current warnings
- maybe remove support for the old server mode and leave request-server only

puellanivis:
If you want to do the work to switch us over to GitHub actions over travis, :woman_shrugging: I wouldn‌’t complain :laughing: 

I’m honestly not all that impressed with `golangci-lint`. If we want to clean up the code, then we can manually run any linting we want and resolve any raised issues when the make sense. Enforcing a list of linting to pass in order to greenlight a PR looks good on paper (which is why it shows up so often in corporate codebases), but it is actually a nightmare of appeasing arbitrary and often poorly considered rules. (In particular, `golangci-lint` seems to be focused on embedding via copy-paste All The Linters, rather than focusing on targetting a well defined and scoped Best Practice.)

As for reworking the whole RequestServer mechanics, I’m kind of :grimacing: about. The only reason to have RequestServer is that it is a helpful abstraction layer hiding facets of SFTP from the implementing service. If the only thing it does is thinly wrap the whole of the SFTP protocol, then it stops being all that useful. If suddenly now one has to implement _all_ the features to get any functionality at all? :-1: More ideally, we want to give a quick access to 80% of the functionality without excessive overhead. If someone needs that 20% we don’t support, then they can deal with a more primitive layer directly (that perhaps we could also build out and expose as a subpackage).

request-server_test.go
```diff
@@ -345,6 +346,38 @@ func TestRequestFstat(t *testing.T) {
...
 }
 
+func TestRequestFsetstat(t *testing.T) {
+	p := clientRequestServerPair(t)
+	defer p.Close()
+	_, err := putTestFile(p.cli, "/foo", "hello")
+	require.NoError(t, err)
+	fp, err := p.cli.OpenFile("/foo", os.O_WRONLY)
+	require.NoError(t, err)
+	err = fp.Truncate(2)
```
puellanivis, line 356: If I comment out this line, and the `assert.Equal(t, fi.Size(), int64(2))` line, then the test still passes. This is because all of the extra writing to `/foo` is done after a `fp.Seek(int64(2), io.SeekStart)`, which means it will end up as `"hehello"` regardless of whether we’ve done the truncate or not.

But really, we should only be testing `Fsetstat` here, so we should probably just test to make sure that the file after the truncate is `fi.Name() == "foo"`, `fi.Size() == 2` and `body == "he"`.

Maybe followed by a truncate to five, and and a test that `fi.Size() == 5` and `body == []byte{ 'h', 'e', 0, 0 , 0 }`. (Depending on if we want to guarantee behavior as per XSI-standards, or not)

drakkan:
> If you want to do the work to switch us over to GitHub actions over travis,  I wouldn‌’t complain 
> 

Ok I'll submit a pull request

> I’m honestly not all that impressed with `golangci-lint`. If we want to clean up the code, then we can manually run any linting we want and resolve any raised issues when the make sense. Enforcing a list of linting to pass in order to greenlight a PR looks good on paper (which is why it shows up so often in corporate codebases), but it is actually a nightmare of appeasing arbitrary and often poorly considered rules. (In particular, `golangci-lint` seems to be focused on embedding via copy-paste All The Linters, rather than focusing on targetting a well defined and scoped Best Practice.)
> 

Ok, we will discuss better about this in future, I'll try to send a pull request with some basic linters

> As for reworking the whole RequestServer mechanics, I’m kind of  about. The only reason to have RequestServer is that it is a helpful abstraction layer hiding facets of SFTP from the implementing service. If the only thing it does is thinly wrap the whole of the SFTP protocol, then it stops being all that useful. If suddenly now one has to implement _all_ the features to get any functionality at all?  More ideally, we want to give a quick access to 80% of the functionality without excessive overhead. If someone needs that 20% we don’t support, then they can deal with a more primitive layer directly (that perhaps we could also build out and expose as a subpackage).

Request-Server is the only choice if you want to interact with non local fs (in SFTPGo I support S3 and GCS and I could add Azure blog storage in future). 

I think we can add optional interfaces for the missing features instead of a new subpackage. I need the generic openfile method to fix the reported file corruption issue in SFTPGo. What do you think about an optional interface? If it is implemented it will be called instead of "Get" and "Put", we can do the same for "Lstat" if an optional interface is implemented we call it otherwise we call "Stat" as now.

I don't know the history of this project, based on your answer I understand that the Request-Server is not intented to replace the server implementation. OK.

puellanivis:
> What do you think about an optional interface?

Yeah, this is a good idea. We should probably work on converging towards something similar to the `io/fs` proposal that has been proposed, and the coolest idea I think to come from that proposal is the idea of extension interfaces providing optional support for features.

Exactly like you said, this would be nice since a “simplest” implementation of `Lstat`/`Stat` would be one function doing `Stat`, but maybe you need and/or can provide an `Lstat` then we could then use implementation of the extension interface to know we could do a specific `Lstat` instead of just the same `Stat`.

The project wasn’t probably as well defined of scope when it started unfortunately. And I think most of the API popped in without a multi-draft proposal process, so some choices were made that were maybe not ideal. (I know `pkg/errors` suffers from some legacy choices that are too late to change now without significant breakage, and the breaking changes would be of little benefit since most of the package’s functionality has been replaced by `fmt.Errorf("… %w")`.)

What with the (probable) future roll out of `io/fs` proposal, that might provide us with a good opportunity to rethink how this whole package should work and what it should look like, and then we could run a few drafts of a v2, and make a solid break towards a better implementation overall,  rather than piecemeal breaking changings.

request-server_test.go
```diff
@@ -345,6 +346,35 @@ func TestRequestFstat(t *testing.T) {
...
+	require.NoError(t, err)
+	err = fp.Truncate(5)
+	require.NoError(t, err)
+	fi, err := fp.Stat()
+	if assert.NoError(t, err) {
+		assert.Equal(t, fi.Name(), "foo")
+		assert.Equal(t, fi.Size(), int64(5))
```
puellanivis, line 363: We still want a test of `f.Size() == 2` after the `Truncate(2)`. Imagine an implementation doesn’t do anything with `Truncate` at all, then after `Write("12345"): Truncate(2); Truncate(5)` it would still have a length of `5`, because it never changed the length at all.

Also, use a `require.NoError(t, err)` for both, and don’t embed them in an `if … { … }` block.

drakkan:
> > What do you think about an optional interface?
> 
> Yeah, this is a good idea. We should probably work on converging towards something similar to the `io/fs` proposal that has been proposed, and the coolest idea I think to come from that proposal is the idea of extension interfaces providing optional support for features.
> 

Ok, I'll try to send a new patch for this (later this week or the next one). Thank you for your reviews :-)

> Exactly like you said, this would be nice since a “simplest” implementation of `Lstat`/`Stat` would be one function doing `Stat`, but maybe you need and/or can provide an `Lstat` then we could then use implementation of the extension interface to know we could do a specific `Lstat` instead of just the same `Stat`.
> 
> The project wasn’t probably as well defined of scope when it started unfortunately. And I think most of the API popped in without a multi-draft proposal process, so some choices were made that were maybe not ideal. (I know `pkg/errors` suffers from some legacy choices that are too late to change now without significant breakage, and the breaking changes would be of little benefit since most of the package’s functionality has been replaced by `fmt.Errorf("… %w")`.)
> 
> What with the (probable) future roll out of `io/fs` proposal, that might provide us with a good opportunity to rethink how this whole package should work and what it should look like, and then we could run a few drafts of a v2, and make a solid break towards a better implementation overall, rather than piecemeal breaking changings.

I quicky readed the proposal some weeks ago, I'll read it better and in future I'll try to help with v2

puellanivis:
I don’t think we need a new patch right now. And most of this patch is already pretty necessary to avoid calling a `Put` request on something that is _definitely_ not a `Put`.

We can get this done, and hopefully checked in before EOW, then cut a release for Friday? I hope? And then we can start looking for the future on how we want a v2 to look.

drakkan:
> I don’t think we need a new patch right now. And most of this patch is already pretty necessary to avoid calling a `Put` request on something that is _definitely_ not a `Put`.
> 

Hi, I'm not fully understanding your answer. We have two issues here:

1) an `SSH_FXP_FSETSTAT` that will become a `Put` request. This issue is fixed in this pull request and if the test cases are now ok I can merge it myself just now
2) if a client open a file for both reading and writing the code [here](https://github.com/pkg/sftp/blob/master/request.go#L212) will set the request method as `Put` and if the client try to read it will corrupt the opened file. We discussed about this issue in #374. It is not clear to me if we want to fix this issue before or after 1.12. If we can break compatibility the fix is quite easy otherwise we should try to add an optional interface and document that if it is not implemented we can have a file corruption issue.

I can work on this today, I'll will be unavailable from tomorrow until Friday

> We can get this done, and hopefully checked in before EOW, then cut a release for Friday? I hope? And then we can start looking for the future on how we want a v2 to look.

puellanivis:
1. Yes, we want this fixed, and the tests are good now. You can merge it. Then we should be good for 1.12. (I will probably do a go module update just before release tomorrow.)

2. Yeah, we need to think about this a bit more than we properly have time for right now. Ideally, we need a better design over all, which better considers this possible case from the ground up. We’ll work on a possible v2 design and iterate on the ideas and see where we go from there, sound good?

PR #375, opened 2020-08-24
Revert PR #370

tklauser:
This reverts commit 3cda2a1c6bfd3bf7af40da1424fbf64a377494b5 / PR #370.

Reason for revert: this broke the Travis CI build, see
https://github.com/pkg/sftp/pull/370#issuecomment-678610816

/cc @puellanivis @davecheney @drakkan


puellanivis:
From the other PR: the example code exports certain types that people might be (incorrectly) relying upon. As this would make it a breaking change, we’re backing out the PR.

PR #377, opened 2020-08-30
request-server: implement Open method

### LINKED ISSUE #374: Request server: allow to open a file for both reading and writing
Hi,

based on the SFTP specs opening a file for both reading and writing is allowed

```
SSH_FXF_READ
      Open the file for reading.

SSH_FXF_WRITE
      Open the file for writing.  If both this and SSH_FXF_READ are
      specified, the file is opened for both reading and writing.
```

pkg/sftp currently does not support this mode. I think this is the reason for this [issue](https://github.com/drakkan/sftpgo/issues/155#issuecomment-676798820). 

Here are the OpenSSH logs when opening a file using sshfs-win:

```
ago 20 16:29:10 p1 sftp-server[9254]: debug3: request 168: open flags 3
ago 20 16:29:10 p1 sftp-server[9254]: open "/tmp/a/Nuovo documento di testo (2).txt" flags READ,WRITE mode 00
ago 20 16:29:10 p1 sftp-server[9254]: debug1: request 168: sent handle handle 2
ago 20 16:29:10 p1 sftp-server[9254]: debug3: request 169: lstat
ago 20 16:29:10 p1 sftp-server[9254]: lstat name "/tmp/a/Nuovo documento di testo (2).txt"
ago 20 16:29:10 p1 sftp-server[9254]: debug1: request 169: sent attrib have 0xf
ago 20 16:29:10 p1 sftp-server[9254]: debug1: request 170: write "/tmp/a/Nuovo documento di testo (2).txt" (handle 2) off 0 len 15
ago 20 16:29:10 p1 sftp-server[9254]: debug3: request 170: sent status 0
ago 20 16:29:10 p1 sftp-server[9254]: sent status Success
ago 20 16:29:10 p1 sftp-server[9254]: debug1: request 171: fstat "/tmp/a/Nuovo documento di testo (2).txt" (handle 2)
ago 20 16:29:10 p1 sftp-server[9254]: debug1: request 171: sent attrib have 0xf
ago 20 16:29:10 p1 sftp-server[9254]: debug1: request 172: fsetstat handle 2
ago 20 16:29:10 p1 sftp-server[9254]: set "/tmp/a/Nuovo documento di testo (2).txt" size 15
ago 20 16:29:10 p1 sftp-server[9254]: debug3: request 172: sent status 0
ago 20 16:29:10 p1 sftp-server[9254]: sent status Success
ago 20 16:29:10 p1 sftp-server[9254]: debug1: request 173: fstat "/tmp/a/Nuovo documento di testo (2).txt" (handle 2)
ago 20 16:29:10 p1 sftp-server[9254]: debug1: request 173: sent attrib have 0xf
ago 20 16:29:10 p1 sftp-server[9254]: debug1: request 174: read "/tmp/a/Nuovo documento di testo (2).txt" (handle 2) off 0 len 4096
ago 20 16:29:10 p1 sftp-server[9254]: debug1: request 174: sent data len 15
ago 20 16:29:10 p1 sftp-server[9254]: debug3: request 175: close handle 2
ago 20 16:29:10 p1 sftp-server[9254]: close "/tmp/a/Nuovo documento di testo (2).txt" bytes read 15 written 15
```

as you can see OpenSSH uses the same handle for both reads and writes.

How do you suggest to support this mode in pkg/sftp? Is ok a new request method, for example "PutGet" mapped to a new interface, for example Filereadwrite that returns both an io.ReaderAt and an io.WriterAt? Do you have other suggestions? Thank you

drakkan:
add an optional interface to FileWriter to allow to open a file for both
writing and reading

Fixes #374

I noticed that an `Open` method is documented but not implemented, see [here](https://github.com/pkg/sftp/blob/master/request-interfaces.go#L30). I implemented this method. The change is backward compatible and implementing the new optional Filewriter interface the request server can handle both reading and writing to the same file fixing the corruption issue reported in #374

I know that we want something different for v2 and so I'll not push this change myself.  

However I cannot wait until we implement v2 to fix a file corruption issue in SFTPGo, so if we don't agree to merge this change I'll maintain it in my fork until we have an alternative, thank you

### DIFF
--- a/request-example.go
+++ b/request-example.go
@@ -10,7 +10,6 @@ import (
 	"io"
 	"os"
 	"path"
-	"path/filepath"
 	"sort"
 	"strings"
 	"sync"
@@ -122,7 +121,7 @@ func (fs *root) Filecmd(r *Request) error {
 			}
 		}
 	case "Rmdir", "Remove":
-		file, err := fs.fetch(filepath.Dir(r.Filepath))
+		file, err := fs.fetch(path.Dir(r.Filepath))
 		if err != nil {
 			return err
 		}
@@ -142,7 +141,7 @@ func (fs *root) Filecmd(r *Request) error {
 		delete(fs.files, r.Filepath)
 
 	case "Mkdir":
-		_, err := fs.fetch(filepath.Dir(r.Filepath))
+		_, err := fs.fetch(path.Dir(r.Filepath))
 		if err != nil {
 			return err
 		}
@@ -203,7 +202,7 @@ func (fs *root) Filelist(r *Request) (ListerAt, error) {
 		}
 		orderedNames := []string{}
 		for fn := range fs.files {
-			if filepath.Dir(fn) == r.Filepath {
+			if path.Dir(fn) == r.Filepath {
 				orderedNames = append(orderedNames, fn)
 			}
 		}
@@ -274,7 +273,7 @@ func newMemFile(name string, isdir bool) *memFile {
 }
 
 // Have memFile fulfill os.FileInfo interface
-func (f *memFile) Name() string { return filepath.Base(f.name) }
+func (f *memFile) Name() string { return path.Base(f.name) }
 func (f *memFile) Size() int64  { return int64(len(f.content)) }
 func (f *memFile) Mode() os.FileMode {
 	ret := os.FileMode(0644)
--- a/request-example.go
+++ b/request-example.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path"
 	"path/filepath"
 	"sort"
 	"strings"
@@ -47,7 +48,7 @@ func (fs *root) Fileread(r *Request) (io.ReaderAt, error) {
 	return file.ReaderAt()
 }
 
-func (fs *root) Filewrite(r *Request) (io.WriterAt, error) {
+func (fs *root) getFileForWrite(r *Request) (*memFile, error) {
 	if fs.mockErr != nil {
 		return nil, fs.mockErr
 	}
@@ -56,29 +57,7 @@ func (fs *root) Filewrite(r *Request) (io.WriterAt, error) {
 	defer fs.filesLock.Unlock()
 	file, err := fs.fetch(r.Filepath)
 	if err == os.ErrNotExist {
-		dir, err := fs.fetch(filepath.Dir(r.Filepath))
-		if err != nil {
-			return nil, err
-		}
-		if !dir.isdir {
-			return nil, os.ErrInvalid
-		}
-		file = newMemFile(r.Filepath, false)
-		fs.files[r.Filepath] = file
-	}
-	return file.WriterAt()
-}
-
-func (fs *root) OpenFile(r *Request) (WriterAtReaderAt, error) {
-	if fs.mockErr != nil {
-		return nil, fs.mockErr
-	}
-	_ = r.WithContext(r.Context()) // initialize context for deadlock testing
-	fs.filesLock.Lock()
-	defer fs.filesLock.Unlock()
-	file, err := fs.fetch(r.Filepath)
-	if err == os.ErrNotExist {
-		dir, err := fs.fetch(filepath.Dir(r.Filepath))
+		dir, err := fs.fetch(path.Dir(r.Filepath))
 		if err != nil {
 			return nil, err
 		}
@@ -89,6 +68,18 @@ func (fs *root) OpenFile(r *Request) (WriterAtReaderAt, error) {
 		fs.files[r.Filepath] = file
 	}
 	return file, nil
+}
+
+func (fs *root) Filewrite(r *Request) (io.WriterAt, error) {
+	file, err := fs.getFileForWrite(r)
+	if err != nil {
+		return nil, err
+	}
+	return file.WriterAt()
+}
+
+func (fs *root) OpenFile(r *Request) (WriterAtReaderAt, error) {
+	return fs.getFileForWrite(r)
 }
 
 func (fs *root) Filecmd(r *Request) error {
--- a/request-example.go
+++ b/request-example.go
@@ -69,6 +69,28 @@ func (fs *root) Filewrite(r *Request) (io.WriterAt, error) {
 	return file.WriterAt()
 }
 
+func (fs *root) OpenFile(r *Request) (WriterAtReaderAt, error) {
+	if fs.mockErr != nil {
+		return nil, fs.mockErr
+	}
+	_ = r.WithContext(r.Context()) // initialize context for deadlock testing
+	fs.filesLock.Lock()
+	defer fs.filesLock.Unlock()
+	file, err := fs.fetch(r.Filepath)
+	if err == os.ErrNotExist {
+		dir, err := fs.fetch(filepath.Dir(r.Filepath))
+		if err != nil {
+			return nil, err
+		}
+		if !dir.isdir {
+			return nil, os.ErrInvalid
+		}
+		file = newMemFile(r.Filepath, false)
+		fs.files[r.Filepath] = file
+	}
+	return file, nil
+}
+
 func (fs *root) Filecmd(r *Request) error {
 	if fs.mockErr != nil {
 		return fs.mockErr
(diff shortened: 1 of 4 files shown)


request-example.go
```diff
@@ -69,6 +69,28 @@ func (fs *root) Filewrite(r *Request) (io.WriterAt, error) {
 	return file.WriterAt()
 }
 
+func (fs *root) OpenFile(r *Request) (WriterAtReaderAt, error) {
+	if fs.mockErr != nil {
+		return nil, fs.mockErr
+	}
+	_ = r.WithContext(r.Context()) // initialize context for deadlock testing
```
puellanivis, line 76: Is this necessary for non-testing situations?

request-example.go
```diff
@@ -69,6 +69,28 @@ func (fs *root) Filewrite(r *Request) (io.WriterAt, error) {
...
+	file, err := fs.fetch(r.Filepath)
+	if err == os.ErrNotExist {
+		dir, err := fs.fetch(filepath.Dir(r.Filepath))
+		if err != nil {
+			return nil, err
+		}
+		if !dir.isdir {
+			return nil, os.ErrInvalid
+		}
+		file = newMemFile(r.Filepath, false)
```
puellanivis, line 88: Hm, we control this function, and it’s not exported. So we might be able to refactor it to return an `error` as well, and we can then put the `fs.fetch(filepath.Dir(…))` and `!dir.isdir` checks into that function? As well, it could be a receiver method on the `fs` struct, and assign the `fs.files[filepath] = file` as well?

Also, I’m not sure if we want `filepath` here, do we? We’re working with memory files, which means we should be able to use the POSIX abstraction regardless of underlying OS. But I guess if the other code is already using `filepath`, then no. 😆

request-example.go
```diff
@@ -309,6 +331,13 @@ func (f *memFile) WriteAt(p []byte, off int64) (int, error) {
 	return len(p), nil
 }
 
+func (f *memFile) ReadAt(p []byte, off int64) (int, error) {
+	f.contentLock.Lock()
+	defer f.contentLock.Unlock()
+	reader := bytes.NewReader(f.content)
+	return reader.ReadAt(p, off)
```
puellanivis, line 338: I think I would probably rather see an actual implementation rather than exploiting the `bytes.NewReader` type, but I doubt there ends up being any difference in speed really, and this is more readable.

Why can’t the `memFile` contain a `bytes.Buffer` and then we could get a bunch of the functionalit— oh, huh, `bytes.Buffer` doesn’t implement `ReaderAt` or `WriterAt`? 😬

request.go
```diff
@@ -142,6 +143,7 @@ func (r *Request) close() error {
 	r.state.RLock()
 	wr := r.state.writerAt
 	rd := r.state.readerAt
+	writerReader := r.state.writerReaderAt
```
puellanivis, line 146: `rw` is the preferred abbreviation for `ReadWriter`s.

request.go
```diff
@@ -210,8 +227,13 @@ func (r *Request) open(h Handlers, pkt requestPacket) responsePacket {
 	var err error
 	switch {
 	case flags.Write, flags.Append, flags.Creat, flags.Trunc:
-		r.Method = "Put"
-		r.state.writerAt, err = h.FilePut.Filewrite(r)
+		if openFileWriter, ok := h.FilePut.(OpenFileWriter); ok && flags.Read {
```
puellanivis, line 230: Seems like we could use an 'if flags.Read { … }` surrounding block to shortcircuit even attempting the type conversion if we know we won’t need it. But I don’t think it will end up being any clearer code, without an early return in the block that makes the `openFileWriter.OpenFile(…)` call.

puellanivis (commented):
I don’t think there’s much if any recommendations for change. Just musings over possible alternatives.

drakkan:
Hi,

after changing the code like this

```
func (fs *root) getFileForWrite(r *Request) (*memFile, error) {
	if fs.mockErr != nil {
		return nil, fs.mockErr
	}
	_ = r.WithContext(r.Context()) // initialize context for deadlock testing
	fs.filesLock.Lock()
	defer fs.filesLock.Unlock()
	file, err := fs.fetch(r.Filepath)
	if err == os.ErrNotExist {
		dir, err := fs.fetch(path.Dir(r.Filepath))
		if err != nil {
			return nil, err
		}
		if !dir.isdir {
			return nil, os.ErrInvalid
		}
		file = newMemFile(r.Filepath, false)
		fs.files[r.Filepath] = file
	}
	return file, nil
}

func (fs *root) Filewrite(r *Request) (io.WriterAt, error) {
	file, err := fs.getFileForWrite(r)
	if err != nil {
		return nil, err
	}
	return file.WriterAt()
}

func (fs *root) OpenFile(r *Request) (WriterAtReaderAt, error) {
	return fs.getFileForWrite(r)
}
```

I'm getting test failures like this one:

```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x58 pc=0x680b45]

goroutine 229 [running]:
github.com/pkg/sftp.(*memFile).TransferError(0x0, 0x7b8da0, 0xc000410140)
	/home/nicola/goprojects/sftp/request-example.go:345 +0x25
github.com/pkg/sftp.(*Request).transferError(0xc000720750, 0x7b8da0, 0xc000410140)
	/home/nicola/goprojects/sftp/request.go:195 +0x185
github.com/pkg/sftp.(*RequestServer).Serve(0xc000616a00, 0x0, 0x0)
	/home/nicola/goprojects/sftp/request-server.go:171 +0x259
github.com/pkg/sftp.clientRequestServerPair.func1(0xc000020ea0, 0xc000313380, 0xc000318020)
	/home/nicola/goprojects/sftp/request-server_test.go:55 +0x245
created by github.com/pkg/sftp.clientRequestServerPair
	/home/nicola/goprojects/sftp/request-server_test.go:40 +0xe5
exit status 2
```

for some reason now in code like this

```
if t, ok := wr.(TransferError); ok {
	t.TransferError(err)
}
```

`wr` implements the interface but is nil (probably because is a pointer to a struct, the struct implements the interface but it is nil). I got the same error in SFTPGo in the past when returning a pointer to a struct. What do you think to add something like this

```
if t, ok := wr.(TransferError); ok && t != nil && !reflect.ValueOf(t).IsNil() {
	t.TransferError(err)
}
```

to avoid this issue? Thank you

puellanivis:
> wr implements the interface but is nil (probably because is a pointer to a struct, the struct implements the interface but it is nil).

Yeah, this is a source of a lot of issues, where a `(*ConcreteType)(nil)` implements a method, but does not expect a `nil`-pointer.

drakkan:
> > wr implements the interface but is nil (probably because is a pointer to a struct, the struct implements the interface but it is nil).
> 
> Yeah, this is a source of a lot of issues, where a `(*ConcreteType)(nil)` implements a method, but does not expect a `nil`-pointer.

OK, so is this PR ready to merge for you? Thank you

request-example.go
```diff
@@ -66,9 +66,21 @@ func (fs *root) Filewrite(r *Request) (io.WriterAt, error) {
...
+func (fs *root) Filewrite(r *Request) (io.WriterAt, error) {
+	file, err := fs.getFileForWrite(r)
+	if err != nil {
+		return nil, err
+	}
 	return file.WriterAt()
 }
 
+func (fs *root) OpenFile(r *Request) (WriterAtReaderAt, error) {
+	return fs.getFileForWrite(r)
```
puellanivis, line 81: Implement this function with:
```
file, err := fs.getFileForWrite(r)
if err != nil {
  return nil, err
}
return file, nil
```

And then you don’t have to do the hacky check for `&& c != nil && reflect.ValueOf(c).IsNil()`.

puellanivis:
More preferably, the `Request.open(…)` handler should handle `err != nil` and not assign any values into `r.state` if there is an error.

It shouldn’t be hard to copy-paste the `if err != nil { return statusFromError(pkt, err) }` into each of the funcitons, and implement them properly, rather than cheating with a “fallthrough” logic pattern…

request.go
```diff
@@ -225,48 +224,55 @@ func (r *Request) call(handlers Handlers, pkt requestPacket, alloc *allocator, o
...
-			openFileWriter, ok := h.FilePut.(OpenFileWriter)
-			if ok {
+			if openFileWriter, ok := h.FilePut.(OpenFileWriter); ok {
 				r.Method = "Open"
-				r.state.writerReaderAt, err = openFileWriter.OpenFile(r)
+				rw, err := openFileWriter.OpenFile(r)
+				if err != nil {
+					return statusFromError(pkt, err)
+				}
+				r.state.writerReaderAt = rw
```
puellanivis, line 237: If we copy-paste the `return &sshFxpHandlePacket{ID: pkt.id(), Handle: r.handle}` here, we don’t need the `if r.Method == "" { … }` anymore. 😄

puellanivis (approved):

puellanivis:
> Great! I wish you could review all my code!

You say that. And then a month in, you’re just like, “but it works, right?” 😰

drakkan:
> > Great! I wish you could review all my code!
> 
> You say that. And then a month in, you’re just like, “but it works, right?”

I don't think so :)

puellanivis:
All good!

PR #378, opened 2020-09-06
align structures

drakkan:


puellanivis:
Not wholly sure what this PR is intended to accomplish.

drakkan:
> Not wholly sure what this PR is intended to accomplish.

this is a micro optimization, it fixes warnings like this one

```
packet.go:707:24: struct of size 32 bytes could be of size 24 bytes
```

puellanivis:
That’s what I thought it was, but that’s struct “packing” less so “alignment”, so I was confused.

Packing (of course) being ordering fields so that the containing struct has less size, while alignment is where the struct itself aligns, which is automatic in Go. (I think that it aligns on the same alignment boundary as its first field?)

I’m sad Tetris didn’t exist when the idea of struct packing was being invented, so we could call it tetrising the struct.

puellanivis (approved):
+1 lgtm

PR #379, opened 2020-09-06
request-server: add lstat support

drakkan:
Add LstatFileLister, an optional interface for FileLister, if this
interface is implemented Lstat requests will call it otherwirse they
will be handled in the same way as Stat, as before

### DIFF
--- a/request-example.go
+++ b/request-example.go
@@ -213,6 +213,12 @@ func (fs *root) Filelist(r *Request) (ListerAt, error) {
 		}
 		return listerat(list), nil
 	case "Stat":
+		if file.symlink != "" {
+			file, err = fs.fetch(file.symlink)
+			if err != nil {
+				return nil, err
+			}
+		}
 		return listerat([]os.FileInfo{file}), nil
 	case "Readlink":
 		if file.symlink != "" {
@@ -224,6 +230,22 @@ func (fs *root) Filelist(r *Request) (ListerAt, error) {
 		return listerat([]os.FileInfo{file}), nil
 	}
 	return nil, nil
+}
+
+// implements LstatFileLister interface
+func (fs *root) Lstat(r *Request) (ListerAt, error) {
+	if fs.mockErr != nil {
+		return nil, fs.mockErr
+	}
+	_ = r.WithContext(r.Context()) // initialize context for deadlock testing
+	fs.filesLock.Lock()
+	defer fs.filesLock.Unlock()
+
+	file, err := fs.fetch(r.Filepath)
+	if err != nil {
+		return nil, err
+	}
+	return listerat([]os.FileInfo{file}), nil
 }
 
 // In memory file-system-y thing that the Hanlders live on
--- a/request-interfaces.go
+++ b/request-interfaces.go
@@ -63,6 +63,13 @@ type FileLister interface {
 	Filelist(*Request) (ListerAt, error)
 }
 
+// LstatFileLister is a FileLister that implements the Lstat method.
+// If this interface is implemented Lstat requests will call it
+// otherwise they will be handled in the same way as Stat
+type LstatFileLister interface {
+	Lstat(*Request) (ListerAt, error)
+}
+
 // ListerAt does for file lists what io.ReaderAt does for files.
 // ListAt should return the number of entries copied and an io.EOF
 // error if at end of list. This is testable by comparing how many you
--- a/request-server_test.go
+++ b/request-server_test.go
@@ -405,6 +405,19 @@ func TestRequestStatFail(t *testing.T) {
 	checkRequestServerAllocator(t, p)
 }
 
+func TestRequestLstat(t *testing.T) {
+	p := clientRequestServerPair(t)
+	defer p.Close()
+	_, err := putTestFile(p.cli, "/foo", "hello")
+	require.NoError(t, err)
+	err = p.cli.Symlink("/foo", "/bar")
+	require.NoError(t, err)
+	fi, err := p.cli.Lstat("/bar")
+	require.NoError(t, err)
+	assert.True(t, fi.Mode()&os.ModeSymlink == os.ModeSymlink)
+	checkRequestServerAllocator(t, p)
+}
+
 func TestRequestLink(t *testing.T) {
 	p := clientRequestServerPair(t)
 	defer p.Close()
--- a/request.go
+++ b/request.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"strings"
 	"sync"
 	"syscall"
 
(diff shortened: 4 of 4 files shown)


request-example.go
```diff
@@ -213,6 +213,12 @@ func (fs *root) Filelist(r *Request) (ListerAt, error) {
 		}
 		return listerat(list), nil
 	case "Stat":
+		if file.symlink != "" {
+			file, err = fs.fetch(file.symlink)
```
puellanivis, line 217: This functionality should be in a separate `"Lstat"` method. The two are different system calls for a reason. And `“Stat”` is the one that does not follow symlinks.

drakkan: `stat` should follow the symlink, it was broken before, please try this sample code:

```
package main

import (
	"fmt"
	"io/ioutil"
	"os"
)

func main() {
	fileName := "testfile.txt"
	fileNameLink := "testfile.link"
	os.Remove(fileName)
	os.Remove(fileNameLink)
	err := ioutil.WriteFile(fileName, []byte("hello"), os.ModePerm)
	if err != nil {
		fmt.Printf("error creating file: %v\n", err)
		os.Exit(1)
	}
	err = os.Symlink(fileName, fileNameLink)
	if err != nil {
		fmt.Printf("error creating symlink: %v\n", err)
		os.Exit(1)
	}
	fi, err := os.Stat(fileName)
	if err != nil {
		fmt.Printf("stat error: %v\n", err)
		os.Exit(1)
	}
	fmt.Printf("stat file info: %+v size: %v is symlink: %v\n", fi, fi.Size(), fi.Mode()&os.ModeSymlink)

	fi, err = os.Stat(fileNameLink)
	if err != nil {
		fmt.Printf("stat error: %v\n", err)
		os.Exit(1)
	}
	fmt.Printf("stat link info: %+v size: %v is symlink: %v\n", fi, fi.Size(), fi.Mode()&os.ModeSymlink)

	fi, err = os.Lstat(fileNameLink)
	if err != nil {
		fmt.Printf("lstat error: %v\n", err)
		os.Exit(1)
	}
	fmt.Printf("lstat link info: %+v size: %v is symlink: %v\n", fi, fi.Size(), fi.Mode()&os.ModeSymlink)
}
```

puellanivis: No, I misread the text from `man stat`:
```
lstat() is identical to stat(), except that if path is a symbolic link, then the link itself is stat-ed, not the file that it refers to.
```

🤦‍♀️

puellanivis: I think we still need an `Lstat` Request handler, though?

puellanivis: We should be able to do it with a `fallthrough` as well:
```
case "Stat":
  if file.symlink != "" {
    file, err := fs.fetch(file.symlink)
    if err != nil { … }
  }
  fallthrough
case "Lstat":
  return listerat([]os.FileInfo{file}), nil
```

drakkan: The point here is to keep backward compatibility. If we add a `Lstat` method to `Filelist` in request-example.go then all the implementors need to do the same. I added an optional interface so `Lstat` is called only if the optional LstatFileLister interface is implemented otherwise Lstat will become a Stat as before. I agree that this is not ideal

puellanivis: But this isn’t `FileList`, it’s a `Request` handler on `switch r.Method { … }`. I don’t see how expanding the functionality of a struct in `request-example.go` would require all implementors to do the same. 😕 Especially, since no where in this `Filelist` receiver method, we are calling a `.Lstat` on anything.

drakkan: Maybe I'm missing something, are we talking about the change I did to the Filelister interface implementation for the in memory [request example](https://github.com/drakkan/sftp/blob/lstat/request-example.go#L185)?

puellanivis: Yes, I’m talking about this specific piece of code that this thread is attached to, which is in `request-example.go`. Good that we now have the correct behavior for `"Stat"` Method, but we should also include an `"Lstat"` example as well, right?

drakkan: `Filelist` is an interface called by the request server and implemented by other software. We are talking about the sample implementation inside pkg/sftp (request-example.go), [here](https://github.com/drakkan/sftpgo/blob/master/sftpd/handler.go#L193) is the implementation for this interface in SFTPGo.

If the request server set the request method to `Lstat` then both request-example.go and the SFTPGo implementation must be updated to handle the new `Lstat` method. The current request-server go never set the request method to `Lstat`. It sets the method to `Stat` for both `stat` and `lstat` so the implementors don't expect the `Lstat` method. For this reson I wrote a backward compatible path. To handle `lstat` request-example must implements the `LstatFileLister` and handle `lstat` inside this optional interface. In SFTPGo I'll do the same. If the interface is not implemented lstat will be hidden as with the current code and the `Filelister` interface will receive `Stat` as request method.

If you prefer I can update the patch to be backward incompatible

puellanivis: OK, I think I see where I’m getting absolutely confused. If the handler implements `LstatFileLister`, then we call the `Lstat(…)` receiver, otherwise we call `Filelist(…)` with `r.Method = "Stat"`?

So, we’re handling two things in very different methods. (Which I suppose is how we did the `OpenFile` extension?)

I think I have been so confused, because I didn’t expect us to handle things in very different directions. 🤔 hm…

drakkan: > OK, I think I see where I’m getting absolutely confused. If the handler implements `LstatFileLister`, then we call the `Lstat(…)` receiver, otherwise we call `Filelist(…)` with `r.Method = "Stat"`?
> 

yes, this is to keep backward compatibility if we call `Filelist(…)` with `r.Method = "Lstat"` all the current implementors must be updated

> So, we’re handling two things in very different methods. (Which I suppose is how we did the `OpenFile` extension?)
> 
> I think I have been so confused, because I didn’t expect us to handle things in very different directions.  hm…

a backward incompatible change, calling `Filelist(…)` with `r.Method = "Lstat"`, is cleaner but we could cause troubles

drakkan:
@puellanivis now that we have clarified the implementations details, what do you think about this pull request?

For what I know this is one of the last missing piece for the request-server, the other one is the support for the statvs packet.

What are your plans for 2.0? I'm thinking to completly remove server/request-server and implementing a single SFTPFs interface with optional extension interfaces. Basically a model very similar to the one proposed [here](https://go.googlesource.com/proposal/+/master/design/draft-iofs.md).

I also would like to work on SFTP v6 support, using the new SFTPFs interface or extending the request-server. The server implementation is useless for my use case since it is only usable with local filesystem. Thank you

puellanivis:
I need to once over the code again, just to make sure I haven’t missed anything, now that I know the way it is working.

For v2, I’ve been kind of thinking it might be good to separate the underlying network “wire format” away from the rest of the logic. So, providing a library that minimally exposes semantically-contained packets (you know, how we packed the structs just recently? We could change their order because the object had “semantic” scope, not “syntactic” scope.) Then ideally, if someone decides they would rather work with the raw packets or something, all power to them.

Then, on top of that low-level → lowest meaningful-semantic-packets, we use the semantic “atoms” to build the design of a new unified API similar to the `io/fs` package, which you linked to.  When the `io/fs` project is rolled out, we might even develop a compatibility layer package that takes a compatible new API client, and makes it compatible with an `io/fs` directly, so you can “mount” or whatever necessary to get that to be compatible.

Ideas… but I haven’t really had much compelling force on such a massive refactor.

puellanivis (approved):
Now that I understand things, I don’t see anything wrong.

PR #380, opened 2020-09-10
Fix build on Plan 9

fhs:
All test are passing on Plan 9, and I've also verified
`examples/go-sftp-server` is working.


attrs.go
```diff
@@ -172,75 +171,3 @@ func marshalFileInfo(b []byte, fi os.FileInfo) []byte {
...
-	case syscall.S_IFIFO:
-		fm |= os.ModeNamedPipe
-	case syscall.S_IFLNK:
-		fm |= os.ModeSymlink
-	case syscall.S_IFREG:
-		// nothing to do
-	case syscall.S_IFSOCK:
-		fm |= os.ModeSocket
-	}
-	if mode&syscall.S_ISGID != 0 {
```
puellanivis, line 195: Huh… one would think that these could be set as (for example) `syscall.S_ISGID = 0` and then this would be recognized as always false, and thus optimized out for us…

I‘m sure there are subtle reasons why they chose to do so.

stat_unix.go
```diff
@@ -0,0 +1,117 @@
+// +build !plan9
```
puellanivis, line 1: I would expect a `_unix.go` file would also exclude `!windows`, and likely `!js` as well. But then they are compatible with this code only because Go provides “fictitious” values so that they are otherwise compable without code changes like we‘ve had to do here for plan9…

Not sure I have an answer for what to do here, though. I think technically, it‘s `_posix.go`. But only because it‘s the part of POSIX that we‘re able to build support for out of the very rudimentary baseline WIndows POSIX support… Maybe `_posix_compat.go`? or just `_compat.go`?

server.go
```diff
@@ -589,14 +575,12 @@ func statusFromError(p ider, err error) sshFxpStatusPacket {
 	ret.StatusError.Code = sshFxFailure
 	ret.StatusError.msg = err.Error()
 
+	if code, ok := translateSyscallError(err); ok {
```
puellanivis, line 578: From the other PR, for better Windows compat of a file underneath a non-existing directory, we probably want to switch to using `if os.IsNotExist(err) { ret.StatusError.Code = sshFxNoSuchFile ; return ret }` which should simplify up a lot of your code handling here? As after, we would not need any special syscall error code handling for file-not-found errors.

puellanivis (commented):
Probably pretty good. Really more nits/lint than anything?

attrs.go
fhs: The `syscall` package is not meant to be portable. It's actually currently frozen, so I think don't we can add `S_ISGID` for plan9 even if we wanted to. In fact, the `syscall` package might be [removed](https://github.com/golang/go/issues/5616#issuecomment-349783401) in the furture in favor of `golang.org/x/sys`.

puellanivis: Indeed. And that‘s probably a good thing. But it‘s weird that they stopped at extended compatibility for windows and JS, but not for plan9.

Maybe it lost priority, and didn‘t keep up? Not sure. Just seems weird in my head to make all the fictional values for windows and JS, but not for plan9. :laughing:

stat_unix.go
fhs: I'll use `_posix.go` because that's what the standard library uses.

server.go
fhs: Yeah, it's better to use `os.IsNotExist` and `os.Permission` here. I was trying to keep the changes to minimum, but I can certainly make that change.

fhs: We still need syscall error code handing for the case when `errno == 0` (we return `sshFxOk`), and when errno is non-zero but not `ENOENT` or `EPERM`, in which case we return `sshFxFailure`.

I'm inclined to leave this change out of this PR, since we're just trying to fix build on plan9 (and not break other systems in the process).

stat_unix.go
fhs: Done

server.go
puellanivis: I’m going to recommend that you hold off of that change. We’ll probably merge the other PR before yours, and then you’ll have a merge conflict to clean up, but hopefully it shouldn’t be too bad. Especially, since their code only touches a strict subset of your code, so you should be pretty good and familiar with the code to make the right conflict resolution.

puellanivis: The other PR has been merged, and as expected, github now reports there are merge conflicts to resolve. :smiley: 

Once you‘ve addressed those, I think we‘re good to commit. Thanks for the work! I know it can be hard to find people who can even test on plan9, let alone do any support work necessary.

fhs: Done. Thanks for the review!

stat_posix.go
```diff
@@ -0,0 +1,117 @@
...
+
+import (
+	"os"
+	"syscall"
+)
+
+const EBADF = syscall.EBADF
+
+func wrapPathError(filepath string, err error) error {
+	switch err.(type) {
```
puellanivis, line 13: Now that I look at this, this could probably better be:
```
if errno, ok := err.(syscall.Errno); ok {
  return &os.PathError{Path: filepath, Err: errno }
}
return err
```

With something similar for the plan9 with `ErrorString` instead of `Errno`.

A good compiler would make both equivalent, so this really should be about what is more readable, and/or if we expect this type-switch to be further expanded.

Your input is desired regarding this matter. It’s probably fine either way.

fhs: I agree that an if-statement is better here. It was using a switch in the original code, so I just left it like that.

I've converted it to an if-statement.

puellanivis (approved):

puellanivis:
I don‘t see anything. I‘ll leave open a day for anyone else who might want to chime in.

PR #382, opened 2020-09-14
server: use os.IsNotExist to map sshFxNoSuchFile

### LINKED ISSUE #381: server Stat call returns incorrect error for some non-existent Windows file paths
(Apologies if I get some of the details wrong here, as I'm just digging into filesystem syscalls as a result of this bug.  This is all foreign territory for me.  But I do feel pretty confident about the fix I have)

---

`os.Stat` on Windows returns `ENOENT` if the requested file or directory does not exist, but `ENOTDIR` if the requested file is a child of a non-existent directory.  For example, given the code:

``` go
_, err1 := os.Stat("/doesnotexist")
fmt.Printf("%#v\n", err1)

_, err2 := os.Stat("/doesnotexist/a/b")
fmt.Printf("%#v\n", err2)
```

On windows, this prints:

``` go
&os.PathError{Op:"CreateFile", Path:"/doesnotexist", Err:0x2}      // ENOENT
&os.PathError{Op:"CreateFile", Path:"/doesnotexist/a/b", Err:0x3}  // ENOTDIR
```

Compared to linux, which treats these both the same:

``` go
&os.PathError{Op:"stat", Path:"/doesnotexist", Err:0x2}      // ENOENT
&os.PathError{Op:"stat", Path:"/doesnotexist/a/b", Err:0x2}  // ENOENT
```

sftp's server implementation maps these errors to ssh status codes in [statusFromError](https://github.com/pkg/sftp/blob/ec5b80d98efe8089fc64b9acafabc0af1d1d70b2/server.go#L568).  In the case of `os.PathError`, this calls [translateErrno](https://github.com/pkg/sftp/blob/ec5b80d98efe8089fc64b9acafabc0af1d1d70b2/server.go#L555) which properly translates the `ENOENT` to `sshFxNoSuchFile`, but the `ENOTDIR` gets left as a generic `sshFxFailure`.

This means that when the response gets [converted back into a standard error](https://github.com/pkg/sftp/blob/ec5b80d98efe8089fc64b9acafabc0af1d1d70b2/client.go#L1313) in an sftp client, it is not seen as a not exists error.  Here is a simple reproduction case:

``` go
func TestStatNonExistent(t *testing.T) {
	cPipe, sPipe := net.Pipe()
	srv, _ := sftp.NewServer(sPipe)
	go srv.Serve()
	client, _ := sftp.NewClientPipe(cPipe, cPipe)

	for _, file := range []string{"/doesnotexist", "/doesnotexist/a/b"} {
		_, err := client.Stat(file)
		if err == nil || !os.IsNotExist(err) {
			t.Errorf("expected 'does not exist' err for file %q.  got: %v", file, err)
		}
	}
}
```

This works fine on Linux, but on Windows returns:

``` sh
--- FAIL: TestStatNonExistent (0.00s)
    c:\Users\willnorris\src\sftptest\sftp_test.go:30: expected 'does not exist' err for file "/doesnotexist/a/b".  
got: sftp: "CreateFile /doesnotexist/a/b: The system cannot find the path specified." (SSH_FX_FAILURE)
```

It's tempting to just have `translateErrno` handle `ENOTDIR` as well, but golang/go#18974 suggests that is not always a good idea.  I'm not 100% sure if the use cases Ian talks about in that bug apply here, but it's not necessary anyway.  As of [a919b76](https://github.com/golang/go/commit/a919b760378343a211719991a9e0b1b819ac9f3d), `os.IsNotExist` properly identifies `ENOTDIR` on windows, so it's as simple as checking `os.IsNotExist` inside of `statusFromError`, and mapping it to `sshFxNoSuchFile`.  I've verified this resolves the reproduction case above, and will followup with that in a PR shortly.

(For what it's worth, I found this while trying to debug [this Windows test failure](https://github.com/willnorris/perkeep/runs/1104803999?check_suite_focus=true#step:6:800) in Perkeep)

willnorris:
Always use os.IsNotExist to identify any OS specific error types that represent missing files or directories.  This resolves an issue on Windows where some system errors (ENOTDIR) were not being identified as 'not found' errors and mapped to sshFxNoSuchFile.

fixes #381

### DIFF
--- a/server.go
+++ b/server.go
@@ -589,24 +589,24 @@ func statusFromError(p ider, err error) sshFxpStatusPacket {
 	ret.StatusError.Code = sshFxFailure
 	ret.StatusError.msg = err.Error()
 
+	if os.IsNotExist(err) {
+		ret.StatusError.Code = sshFxNoSuchFile
+		return ret
+	}
+
 	switch e := err.(type) {
 	case syscall.Errno:
 		ret.StatusError.Code = translateErrno(e)
 	case *os.PathError:
 		debug("statusFromError,pathError: error is %T %#v", e.Err, e.Err)
-		if os.IsNotExist(err) {
-			ret.StatusError.Code = sshFxNoSuchFile
-		} else if errno, ok := e.Err.(syscall.Errno); ok {
+		if errno, ok := e.Err.(syscall.Errno); ok {
 			ret.StatusError.Code = translateErrno(errno)
 		}
 	case fxerr:
 		ret.StatusError.Code = uint32(e)
 	default:
-		switch e {
-		case io.EOF:
+		if e == io.EOF {
 			ret.StatusError.Code = sshFxEOF
-		case os.ErrNotExist:
-			ret.StatusError.Code = sshFxNoSuchFile
 		}
 	}
 
--- a/server.go
+++ b/server.go
@@ -594,7 +594,9 @@ func statusFromError(p ider, err error) sshFxpStatusPacket {
 		ret.StatusError.Code = translateErrno(e)
 	case *os.PathError:
 		debug("statusFromError,pathError: error is %T %#v", e.Err, e.Err)
-		if errno, ok := e.Err.(syscall.Errno); ok {
+		if os.IsNotExist(err) {
+			ret.StatusError.Code = sshFxNoSuchFile
+		} else if errno, ok := e.Err.(syscall.Errno); ok {
 			ret.StatusError.Code = translateErrno(errno)
 		}
 	case fxerr:
--- a/server_test.go
+++ b/server_test.go
@@ -354,7 +354,7 @@ func TestStatNonExistent(t *testing.T) {
 
 	for _, file := range []string{"/doesnotexist", "/doesnotexist/a/b"} {
 		_, err := client.Stat(file)
-		if err == nil || !os.IsNotExist(err) {
+		if !os.IsNotExist(err) {
 			t.Errorf("expected 'does not exist' err for file %q.  got: %v", file, err)
 		}
 	}
--- a/server_test.go
+++ b/server_test.go
@@ -344,3 +344,18 @@ func TestOpenStatRace(t *testing.T) {
 	os.Remove(tmppath)
 	checkServerAllocator(t, server)
 }
+
+// Ensure that proper error codes are returned for non existent files, such
+// that they are mapped back to a 'not exists' error on the client side.
+func TestStatNonExistent(t *testing.T) {
+	client, server := clientServerPair(t)
+	defer client.Close()
+	defer server.Close()
+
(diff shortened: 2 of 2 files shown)


server_test.go
```diff
@@ -344,3 +344,18 @@ func TestOpenStatRace(t *testing.T) {
...
+// Ensure that proper error codes are returned for non existent files, such
+// that they are mapped back to a 'not exists' error on the client side.
+func TestStatNonExistent(t *testing.T) {
+	client, server := clientServerPair(t)
+	defer client.Close()
+	defer server.Close()
+
+	for _, file := range []string{"/doesnotexist", "/doesnotexist/a/b"} {
+		_, err := client.Stat(file)
+		if err == nil || !os.IsNotExist(err) {
```
puellanivis, line 357: `os.IsNotExist(nil)` will already return false, so we don‘t really need the `err == nil` check. (While not explicitly documented behavior, this should follow logically from the contract of the function, that is, since `err == nil` is not a file not found error, and should therefore never return true.)

https://play.golang.org/p/lFB2yI3oQ5i

server.go
```diff
@@ -594,7 +594,9 @@ func statusFromError(p ider, err error) sshFxpStatusPacket {
 		ret.StatusError.Code = translateErrno(e)
 	case *os.PathError:
 		debug("statusFromError,pathError: error is %T %#v", e.Err, e.Err)
-		if errno, ok := e.Err.(syscall.Errno); ok {
+		if os.IsNotExist(err) {
```
puellanivis, line 597: Hm… we might be able to move this up, before the type switch, with an early return.

```
…
ret.StatusErr.msg = err.Error()

if os.IsNotExist(err) {
  ret.StatusError.Code = sshFxNoSuchFile
  return ret
}

switch e := err.(type) {
…
```

Then we would also cover the below `default: switch { case os.ErrNotExist: … }` case in a more full and proper way as well.

puellanivis (commented):
We might be able to refactor to cover all `os.IsNotExist(…)` cases better.

At the very least, we should be using `os.IsNotExist(e)` as well, instead of the more fragile `case os.ErrNotExist:`

server.go
willnorris: done.  I went ahead and simplified the final switch statement since it now only has a single case.

server_test.go
willnorris: done

willnorris:
technically, this means we could also remove the `syscall.ENOENT` case from inside of `translateError`, since that is only ever called from `statusFromError`.  I think it could still be reasonable to leave it there, but happy to remove it if you'd like.

puellanivis:
Up to you, if you want to remove the ENOENT case. Like you say, we should never hit it anymore. But I am going to guess that it’s not actually too much code or binary bloat, so it might be better to keep it around as a just in case backup? 🤷‍♀️

I’m fine either way.

willnorris:
yeah, I think probably leave it in.  👍

Issue #383, opened 2020-09-17
InMemHandler does not obey Truncate flag


jimrobinson:
I've written a client that uses github.com/pkg/sftp and my tests were failing due to what I think is a bug in sftp.InMemHandler.

It looks to me as though sftp.InMemHandler does not truncate files when a client Create operation reuses an existing file.

I've attached an example as [inmemsftp.tar.gz](https://github.com/pkg/sftp/files/5235746/inmemsftp.tar.gz)

The test client makes three distinct `sftp.Client#OpenFile` calls, specifying os.O_TRUNC:
```
fh, err := client.OpenFile(path, os.O_WRONLY|os.O_CREATE|os.O_TRUNC)
```

the intent being to overwrite the file and replace its contents with new ones.

In this example I use a python sftp server to show the client code does appear to do the correct thing when using it.

I haven't bundled the env files from the python test because I imagine that'd be considered a security concern, but the README.txt shows the steps I followed:
```
# we're using an sftp server impl in python to compare to inmem results
$ python -V
Python 3.7.8

# we setup an virtual environment for python
$ python -m venv env

# we activate the environment
$ source env/bin/activate

# within the environment we install sftpserver
(env) $ pip install --upgrade pip && pip install sftpserver
... output elided ...

# we start the sftp server
(env) $ sftpserver -k id_rsa  &
[1] 987395

# build the inmemsftp test client from main.go
(env) $ go build

# run the test client and we see expected output
# for our write operations
(env) $ ./inmemsftp
INFO:paramiko.transport:Connected (version 2.0, client Go)
INFO:paramiko.transport:Auth rejected (none).
INFO:paramiko.transport:Auth granted (password).
/test 000
/test 11
/test 2
ERROR:paramiko.transport:Socket exception: Connection reset by peer (104)

# kill the sftpserver
(env) $ kill %1

# now try again using the inmem sftp handlers and we
# we see that the truncate operation is not being
# satisfied
(env) $ ./inmemsftp -inmem
Listening on 127.0.0.1:3373
2020/09/16 18:34:44 login detected: user
/test 000
/test 110
/test 210
2020/09/16 18:34:44 sftp client exited session.
```

drakkan:
Hi,

[...]

PR #384, opened 2020-09-17
Refactor: request-example openfile

puellanivis:
While looking into [Issue-383](https://github.com/pkg/sftp/issues/383), I realized that there were a small bunch of corner cases not being covered properly. This ended up being a bit more than just the openfile refactor, but in the end ends up with much improved semantics when dealing with the MemFile handler.

During development a large number of tests started panicking when I broke file creation. The tests should not panic, and their `assert.Nil` or `assert.NoError` tests have been replaced with a `require.NoError`. All `assert.Nil` of error values have been replaced with either `assert.NoError` or `require.NoError`.

As such there are a heavy number of test lines changed, which do not actually change any behavior of the tests at all.

Resolved:
* `O_TRUNC` was being ignored on existing files.
* `O_EXCL` and `O_CREATE` were being ignored.
* in most places where symlinks should be followed, they are now followed until a non-symlink file/dir, or `os.ErrNotExist`
* infinite symlinks will eventually error; following 5 symlinks deep will assume a symlink loop, and error
* creating dangling symlink _files_ should actually be allowed, mkdir dangling symlinks should fail
* `O_CREATE` with a dangling symlink should create the target filename, not the symlink‘s name
* a file being made in a symlink to a directory should be made under the target directory, not under the symlink
* new symlinks were not checking for an existing parent directory
* symlink should not be able to overwrite an existing file/dir/symlink
* new directories were not checking for an existing parent directory
* mkdir should not be able to overwrite an existing file/dir/symlink
* new file/dir/symlink creation has been unified to ensure proper locating at a canonical name is made
* rmdir should only work on directories
* remove should only work on files

Known issue:
* Absolutely sure there are still a ton of symlink corner-case issues. These would require a pretty complete and total redesign though.


drakkan:
Hi @puellanivis this seems a great work, thank you!

puellanivis:
OK, I think all that is left, is a bit of verification of SSH vs POSIX semantics (like `rename` does not allow overwrite in SSH, but does in POSIX) but I think there’s a bit of deeper problems with compliance on that, i.e. the normal `_RENAME` packet actually calls `os.Rename(‌…)` which will actually replace existing files, unilike SSH standard. :upside_down_face:

puellanivis:
Alright, getting any more compliance of out this fake system would definitely require significantly more effort, and a pretty complete overhaul. So, I’m going to leave it here, and we can attack potential further non-compliance issues as they come, or reimplement the in-memory-filesystem more properly in another PR.

I have taken the time to expand support for a `PosixRenameFileCmder` in the style of the `OpenFileFileWriter` and `LstatFileLister` so that the proper semantics of `SSH_FXP_RENAME` not allowing for replacing existing files can be maintained at least for our in-memory-filesystem.

drakkan:
It looks good to me, thank you!

PR #387, opened 2020-10-23
Clean up after Client tests + some minor stuff

greatroar:
I noticed that running the tests left a lot of clutter in $TMPDIR. This PR fixes that; if I run

```
mkdir /tmp/sftp
TMPDIR=/tmp/sftp go test -integration -testserver
```

I end up with an empty /tmp/sftp.

Also:

* Some tests were generating safe temporary paths with ioutil.TempFile, then assumed that a path derived from that (f.Name() + ".new" etc.) was also safe. These now use a directory, which also produces less clutter in $TMPDIR and means we only need one defer to clean up.
* Most tests generate files with more of the test name embedded in them, making missing cases easier to spot. This isn't necessary, but may serve as an example for future contributions.
* Tests that require uid = 0 now skip earlier.

### DIFF
--- a/client_integration_test.go
+++ b/client_integration_test.go
@@ -1743,9 +1743,13 @@ var matchTests = []MatchTest{
 	{"[", "a", false, ErrBadPattern},
 	{"[^", "a", false, ErrBadPattern},
 	{"[^bc", "a", false, ErrBadPattern},
-	{"a[", "a", false, nil},
 	{"a[", "ab", false, ErrBadPattern},
 	{"*x", "xxx", true, nil},
+
+	// The following test behaves differently on Go 1.15.3 and Go tip as
+	// https://github.com/golang/go/commit/b5ddc42b465dd5b9532ee336d98343d81a6d35b2
+	// (pre-Go 1.16). TODO: reevaluate when Go 1.16 is released.
+	//{"a[", "a", false, nil},
 }
 
 func errp(e error) string {
--- a/client_integration_test.go
+++ b/client_integration_test.go
@@ -149,11 +149,11 @@ func testClientGoSvr(t testing.TB, readonly bool, delay time.Duration) (*Client,
 	return client, exec.Command("true")
 }
 
-// testClient returns a *Client connected to a localy running sftp-server
+// testClient returns a *Client connected to a locally running sftp-server
 // the *exec.Cmd returned must be defer Wait'd.
 func testClient(t testing.TB, readonly bool, delay time.Duration) (*Client, *exec.Cmd) {
 	if !*testIntegration {
-		t.Skip("skipping intergration test")
+		t.Skip("skipping integration test")
 	}
 
 	if *testServerImpl {
--- a/client_integration_test.go
+++ b/client_integration_test.go
@@ -202,7 +202,7 @@ func TestClientLstat(t *testing.T) {
 	defer cmd.Wait()
 	defer sftp.Close()
 
-	f, err := ioutil.TempFile("", "sftptest")
+	f, err := ioutil.TempFile("", "sftptest-lstat")
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -229,7 +229,7 @@ func TestClientLstatIsNotExist(t *testing.T) {
 	defer cmd.Wait()
 	defer sftp.Close()
 
-	f, err := ioutil.TempFile("", "sftptest")
+	f, err := ioutil.TempFile("", "sftptest-lstatisnotexist")
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -246,10 +246,12 @@ func TestClientMkdir(t *testing.T) {
 	defer cmd.Wait()
 	defer sftp.Close()
 
-	dir, err := ioutil.TempDir("", "sftptest")
+	dir, err := ioutil.TempDir("", "sftptest-mkdir")
 	if err != nil {
 		t.Fatal(err)
 	}
+	defer os.RemoveAll(dir)
+
 	sub := path.Join(dir, "mkdir1")
 	if err := sftp.Mkdir(sub); err != nil {
 		t.Fatal(err)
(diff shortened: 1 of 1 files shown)


puellanivis:
All of this looks pretty good, except I have concerns about the possible incompatibilities in how Windows and plan9(?) can potentially handle `filepath` generated paths different from `path`.

I will admit I haven’t looked too deeply into any of these specific cases, but we’ve have been bitten before with inconsistent application of `path` and `filepath` (former for remote paths, and later for local paths).

greatroar:
All paths are sent to a locally running server, so the interpretation of path names should be the same. Also, TempFile uses filepath.Join internally, so you should already have noticed if there were any problem with it.

greatroar:
Test failure seems unrelated:

```
=== RUN   TestMatch
    client_integration_test.go:1797: Match(`a[`, `a`) = false, "syntax error in pattern" want false, "<nil>"
--- FAIL: TestMatch (0.00s)
```

on Go tip.

greatroar:
The cause seems to be golang/go@b5ddc42b465dd5b9532ee336d98343d81a6d35b2.

puellanivis:
I think we should comment out that test with a link to the commit you linked to, and a `TODO` statement that go versions are going to disagree on this test all the way until (likely) go1.16 is our last supported version.

greatroar:
Commented out with a TODO. Tests pass.

client_integration_test.go
```diff
@@ -635,13 +645,14 @@ func TestClientRename(t *testing.T) {
 	defer cmd.Wait()
 	defer sftp.Close()
 
-	f, err := ioutil.TempFile("", "sftptest")
-	if err != nil {
-		t.Fatal(err)
-	}
+	dir, err := ioutil.TempDir("", "sftptest-rename")
+	require.Nil(t, err)
```
puellanivis, line 649: `require.NoError(t, err)` is better.

puellanivis (commented):
When checking errors, it’s highly recommended to use `(require|assert).NoError()` and `(require|assert).Error()` as they have more semantic context than `IsNil` and `IsNotNil`. The output from the `assert.Nil/NotNil` is quite a bit less useful:

```
--- FAIL: TestFoo (0.00s)
    foo_test.go:15: 
        	Error Trace:	foo_test.go:15
        	Error:      	Received unexpected error:
        	            	oh noes!
        	Test:       	TestFoo
    foo_test.go:16: 
        	Error Trace:	foo_test.go:16
        	Error:      	Received unexpected error:
        	            	pkg/errors
        	            	github.com/puellanivis/test-notnil-vs-error.TestFoo
        	            		/home/snowgirl/Work/tmp/go/test-notnil-vs-error/foo_test.go:16
        	            	testing.tRunner
        	            		/usr/lib64/go/1.14/src/testing/testing.go:1054
        	            	runtime.goexit
        	            		/usr/lib64/go/1.14/src/runtime/asm_amd64.s:1373
        	Test:       	TestFoo
    foo_test.go:17: 
        	Error Trace:	foo_test.go:17
        	Error:      	Expected nil, but got: &errors.errorString{s:"oh noes!"}
        	Test:       	TestFoo
    foo_test.go:19: 
        	Error Trace:	foo_test.go:19
        	Error:      	An error is expected but got nil.
        	Test:       	TestFoo
    foo_test.go:20: 
        	Error Trace:	foo_test.go:20
        	Error:      	Expected value not to be nil.
        	Test:       	TestFoo
```

client_integration_test.go
greatroar: Fixed. I didn't know about that function. I'm going to use it in my own code now too :)

client_integration_test.go
```diff
@@ -635,13 +645,14 @@ func TestClientRename(t *testing.T) {
...
 
-	f, err := ioutil.TempFile("", "sftptest")
-	if err != nil {
-		t.Fatal(err)
-	}
+	dir, err := ioutil.TempDir("", "sftptest-rename")
+	require.NoError(t, err)
+	defer os.RemoveAll(dir)
+	f, err := os.Create(filepath.Join(dir, "old"))
+	require.Nil(t, err)
```
puellanivis, line 652: I only mentioned it on the first line that was relevant, because I don’t want to spam your PR with a whole bunch of linting messages.

However, please sweep the whole file for `require.Nil`, you‘re the only code using `require` and/or `assert` right now (well, and a previous commit from you that I so unhelpfully didn’t vet fully), so it shouldn’t be a whole bunch of change.

A quick look seems like you should be able to a file-wide search-replace of `require.Nil` to `require.NoError`.

greatroar: Sorry, amend/rebase error on my part. I've now caught all instances, I think, including one that wasn't part of this PR to begin with:

```
$ git grep Nil
request-server_test.go: assert.NotNil(t, _foo.ctx)
request-server_test.go: assert.Nil(t, rf)
request-server_test.go: assert.Nil(t, fi)
```

puellanivis: Yeah, there were like three from a previous PR that I totally missed during review. :laughing:

greatroar:
@puellanivis Anything more to do here?

client_integration_test.go
```diff
@@ -773,6 +794,7 @@ func TestClientChmod(t *testing.T) {
 	if err != nil {
```
puellanivis, line 794: Just above this, don’t we want to name the tempfile `sftptest-chmod`?

client_integration_test.go
```diff
@@ -795,6 +817,7 @@ func TestClientChmodReadonly(t *testing.T) {
...
```
puellanivis, line 817: Same here, `sftptest-chmodreadonly`?

client_integration_test.go
```diff
@@ -971,6 +1005,7 @@ func TestClientTruncateReadonly(t *testing.T) {
...
```
puellanivis, line 1005: `sftptest-truncreadonly`

puellanivis (commented):
pedantry: three tempfiles should probably be renamed to match the new pattern?

greatroar:
Done, and renamed the remaining twelve too.

puellanivis:
Looks good. :+1: :)

PR #389, opened 2020-10-28
Don't crash when the packet length is zero

greatroar:
A packet with an invalid length causes a panic in recvPacket. Patch includes the fuzz test that found this out in a few seconds.

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -1,6 +1,7 @@
 package sftp
 
 import (
+	"bytes"
 	"errors"
 	"io"
 	"os"
@@ -207,4 +208,23 @@ func testFstatOption(t *testing.T, o ClientOption, value bool) {
 func TestUseFstatChecked(t *testing.T) {
 	testFstatOption(t, UseFstat(true), true)
 	testFstatOption(t, UseFstat(false), false)
+}
+
+type sink struct{}
+
+func (*sink) Close() error                { return nil }
+func (*sink) Write(p []byte) (int, error) { return len(p), nil }
+
+func TestClientZeroLengthPacket(t *testing.T) {
+	// Packet length zero (never valid). This used to crash the client.
+	packet := []byte{0, 0, 0, 0}
+
+	r := bytes.NewReader(packet)
+	c, err := NewClientPipe(r, &sink{})
+	if err == nil {
+		t.Error("expected an error, got nil")
+	}
+	if c != nil {
+		c.Close()
+	}
 }
--- a/fuzz.go
+++ b/fuzz.go
@@ -1,0 +1,22 @@
+// +build gofuzz
+
+package sftp
+
+import "bytes"
+
+type sink struct{}
+
+func (*sink) Close() error                { return nil }
+func (*sink) Write(p []byte) (int, error) { return len(p), nil }
+
+var devnull = &sink{}
+
+// To run: go-fuzz-build && go-fuzz
+func Fuzz(data []byte) int {
+	c, err := NewClientPipe(bytes.NewReader(data), devnull)
+	if err != nil {
+		return 0
+	}
+	c.Close()
+	return 1
+}
--- a/packet.go
+++ b/packet.go
@@ -154,6 +154,10 @@ func recvPacket(r io.Reader, alloc *allocator, orderID uint32) (uint8, []byte, e
 		debug("recv packet %d bytes too long", length)
 		return 0, nil, errLongPacket
 	}
+	if length == 0 {
+		debug("recv packet of 0 bytes too short")
+		return 0, nil, errShortPacket
+	}
 	if alloc == nil {
 		b = make([]byte, length)
 	}
--- a/server_test.go
+++ b/server_test.go
@@ -1,6 +1,7 @@
 package sftp
 
 import (
+	"bytes"
 	"io"
 	"os"
 	"path"
@@ -13,6 +14,7 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 const (
@@ -363,5 +365,33 @@ func TestStatNonExistent(t *testing.T) {
 		if !os.IsNotExist(err) {
 			t.Errorf("expected 'does not exist' err for file %q.  got: %v", file, err)
 		}
+	}
+}
+
+func TestServerWithBrokenClient(t *testing.T) {
+	validInit := sp(sshFxInitPacket{Version: 3})
+	brokenOpen := sp(sshFxpOpenPacket{Path: "foo"})
+	brokenOpen = brokenOpen[:len(brokenOpen)-2]
+
+	for _, clientInput := range [][]byte{
+		// Packet length zero (never valid). This used to crash the server.
+		{0, 0, 0, 0},
+		append(validInit, 0, 0, 0, 0),
+
+		// Client hangs up mid-packet.
+		append(validInit, brokenOpen...),
+	} {
+		srv, err := NewServer(struct {
+			io.Reader
+			io.WriteCloser
+		}{
+			bytes.NewReader(clientInput),
+			&sink{},
+		})
+		require.NoError(t, err)
+
+		err = srv.Serve()
+		assert.Error(t, err)
+		srv.Close()
 	}
 }


packet.go
```diff
@@ -153,6 +154,9 @@ func recvPacket(r io.Reader, alloc *allocator, orderID uint32) (uint8, []byte, e
 	if length > maxMsgLength {
 		debug("recv packet %d bytes too long", length)
 		return 0, nil, errLongPacket
+	} else if length < minMsgLength {
```
davecheney, line 157: not need for else after return thanks

davecheney (changes requested):
thank you for working on this. I don't see much value in fuzzing the private API. Can you demonstrate that this crash can be trigged via a public API?

greatroar:
Not an API, but a client panics when it connects to a broken server like the following one:

```go
func TestBrokenServer(t *testing.T) {
	cr, sw := io.Pipe() // client reads, server writes
	sr, cw := io.Pipe() // server reads, client writes

	// Broken server that doesn't speak proper SFTP.
	go func() {
		go io.Copy(ioutil.Discard, sr)
		sw.Write([]byte{0, 0, 0, 0})
		sw.Close()
	}()

	_, err := NewClientPipe(cr, cw)
	if err != nil {
		t.Fatal(err)
	}
}
```

packet.go
greatroar: Fixed.

greatroar:
I can change the fuzz test to do something similar. It would be more expensive to run, obviously.

davecheney:
Can you write that as a failing test case?

davecheney:
Remember that new client pipe doesn't actually need an io.pipe, that should make the test simpler, and simpler to write a fuzzer

greatroar:
New test and fuzzer now calls NewClientPipe.

greatroar:
The test failure is the same as one master (fixed in #387).

client_test.go
```diff
@@ -208,3 +212,24 @@ func TestUseFstatChecked(t *testing.T) {
...
+
+func (*sink) Close() error                { return nil }
+func (*sink) Write(p []byte) (int, error) { return len(p), nil }
+
+func TestBrokenServer(t *testing.T) {
+	// Simulate a server that sends a single packet with too small a length,
+	// regardless of what the client sends.
+
+	var packet [4]byte
+	w := &sink{}
```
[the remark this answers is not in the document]
greatroar: Changing NewClientPipe would, strictly speaking, be a backward incompatible change. Also, it suggests that the caller should close the pipe, leading to double Close errors in client code.

client_test.go
```diff
@@ -208,3 +212,24 @@ func TestUseFstatChecked(t *testing.T) {
...
+
+	for length := uint32(0); length < minMsgLength; length++ {
+		binary.BigEndian.PutUint32(packet[:], length)
+
+		r := ioutil.NopCloser(bytes.NewReader(packet[:]))
```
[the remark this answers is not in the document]
greatroar: My mistake, I thought the first arg needed to be a Closer too.

fuzz.go
```diff
@@ -0,0 +1,29 @@
...
+func (*sink) Write(p []byte) (int, error) { return len(p), nil }
+
+var devnull = &sink{}
+
+// To run: go-fuzz-build && go-fuzz
+func Fuzz(data []byte) int {
+	r := ioutil.NopCloser(bytes.NewReader(data))
+	c, err := NewClientPipe(r, devnull)
+	if err == nil {
+		c.Close()
```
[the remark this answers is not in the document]
greatroar: Yes. go-fuzz is not the most user-friendly tool, but it's quite powerful. There's a [draft proposal](https://go.googlesource.com/proposal/+/master/design/draft-fuzzing.md) for fuzz testing in the stdlib, but for now, go-fuzz is the best we've got.

Re: only testing recvVersion, @davecheney didn't want to test a private API. This does exercise the main piece of code that the client and server share, but it doesn't get to see all the inlined packet unmarshaling in the client. Getting the fuzzer to test that has proven very difficult, since you need to simulate an entire conversation.

client_test.go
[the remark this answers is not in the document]
puellanivis: No, it would not be a backwards-incompatible change. All `io.WriteCloser` are also `io.Writer`, and with `if closer, ok := writer.(io.Closer); ok { closer.Close() }` it also retains the same behavior. This also is not unusual design: https://golang.org/pkg/crypto/cipher/#StreamWriter already exhibits this same design for `io.Writer`, and https://golang.org/pkg/net/http/#Client.Post uses it for `io.Reader`.

Since this is not a reciever method, there are no interface mismatch concerns, so really the only “incompatible” thing about it is if someone explicitly uses the type: `func(io.Reader, io.WriteCloser, ...sftp.ClientOption) (*sftp.Client, error)`,  but avoiding the type in declarations (`var newClientPipe = sftp.NewClientPipe`) should be proper hygiene, so that the type is auto-updated, leaving really only a situation where the function itself it is either a parameter to another function, or a field in a struct. Both of those seem quite unlikely, and unnecessary.

fuzz.go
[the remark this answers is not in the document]
puellanivis: :+1: sounds good

packet.go
```diff
@@ -154,6 +155,10 @@ func recvPacket(r io.Reader, alloc *allocator, orderID uint32) (uint8, []byte, e
 		debug("recv packet %d bytes too long", length)
 		return 0, nil, errLongPacket
 	}
+	if length < minMsgLength {
```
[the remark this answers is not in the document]
puellanivis: RIght now, it would only have the one use case, but it might be a good helper function to have around regardless, for better clarity of purpose. Like rather than just having an `b = append(b, packetType)` we can have `b = marshalByte(b, packetType)` that gives semantic information about purpose.

But _that_ is a whole big can of bigger worms than just dealing with this. I think I’ll be behind any decision you come to here, the magic isn’t _that_ concerning… there is for instance, no sftp packet I am aware of that is just `packetLength + packeType` even though strictly that wouldn’t necessary be technically incompatible.

client_test.go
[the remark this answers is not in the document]
puellanivis: It should also add, this is a convenience for testing more than a good API fix, and it’s not without _any_ risk, it just should be low risk. But we should perhaps tackle it all on its own.

packet.go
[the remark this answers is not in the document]
greatroar: Done. I did have to install a check for EOF below, to make sure that gets translated to errShortPacket.

packet.go
```diff
@@ -154,11 +154,18 @@ func recvPacket(r io.Reader, alloc *allocator, orderID uint32) (uint8, []byte, e
...
+		debug("recv packet of %d bytes too short", length)
+		return 0, nil, errShortPacket
+	}
 	if alloc == nil {
 		b = make([]byte, length)
 	}
 	if _, err := io.ReadFull(r, b[:length]); err != nil {
 		debug("recv packet %d bytes: err %v", length, err)
+		if err == io.EOF || err == io.ErrUnexpectedEOF {
+			err = errShortPacket
```
puellanivis, line 167: I mean, technically, in these cases, I would expect `io.ErrUnexpectedEOF`.

We’re trying to read a packet of a `length > 0` but before we can read it all, we encounter an EOF. Perfect case for an `io.ErrUnexpectedEOF`. It‘s not a too short packet (where we read the whole length but an internal value has a length that is longer than the packet) it‘s a proper length packet, but like… EOF before reading it all. That‘s unexpected.

(Sorry to keep “waffling” like this. Code polishing is good, but sometimes you end up going a different direction, then the wind turns, and oh no, now you’re stuck going the wrong way.)

greatroar: I think io.ErrUnexpectedEOF is conflating two error conditions. If client code calls io.ReadFull on an sftp.File, it cannot distinguish a protocol error/server crash from the premature end of an actual file on the remote. Maybe errShortPacket isn't the right value, but there should be a stronger hint that something went wrong in transmission.

Come to think of it, the same may hold true for the ReadFull above, which can return both kinds of EOF...

Or maybe I'm misreading something and I should check again in the morning.

client_test.go
```diff
@@ -208,3 +210,20 @@ func TestUseFstatChecked(t *testing.T) {
...
+func TestBrokenServer(t *testing.T) {
+	// Simulate a server that sends a single packet with zero length,
+	// then closes the connection. This used to cause a panic in the client.
+
+	packet := []byte{0, 0, 0, 0}
+	w := &sink{}
+
+	r := bytes.NewReader(packet[:])
+	_, err := NewClientPipe(r, w)
+	assert.Equal(t, errShortPacket, err)
```
davecheney, line 228: I don’t think it’s necessary to check the type of the error, only that the operation did not succeed.

fuzz.go
```diff
@@ -0,0 +1,25 @@
...
+
+// To run: go-fuzz-build && go-fuzz
+func Fuzz(data []byte) int {
+	c, err := NewClientPipe(bytes.NewReader(data), devnull)
+	if err == nil {
+		c.Close()
+		return 1
+	}
+	if c != nil {
+		panic("error, but non-nil *Client")
```
davecheney, line 22: Please drop this check, there is no utility in fuzzing if new client pipe returns a client instance in the case of an error.

packet_test.go
```diff
@@ -213,6 +215,18 @@ func TestRecvPacket(t *testing.T) {
...
 
+func TestRecvPacketTooShort(t *testing.T) {
+	for _, packet := range [][]byte{
+		{0, 0, 0, 0},
+		{0, 0, 0, 3},
+		{0, 0, 0, 4, sshFxpInit},
+		{0, 0, 0, 200, sshFxpData, 1, 1, 1, 1, 1, 1},
+	} {
+		_, _, err := recvPacket(bytes.NewReader(packet), nil, 0)
+		assert.Equal(t, errShortPacket, err)
```
davecheney, line 226: You don’t need to bring in a library to write err == errShortRead.

davecheney (changes requested):
Thank you for working on this

client_test.go
greatroar: See the discussion above about what error should be returned here.

davecheney: Thanks, i'll defer to the other maintainers, but I would prefer not to take another dependency for this check

client_test.go
```diff
@@ -208,3 +210,20 @@ func TestUseFstatChecked(t *testing.T) {
...
+func (*sink) Write(p []byte) (int, error) { return len(p), nil }
+
+func TestBrokenServer(t *testing.T) {
+	// Simulate a server that sends a single packet with zero length,
+	// then closes the connection. This used to cause a panic in the client.
+
+	packet := []byte{0, 0, 0, 0}
+	w := &sink{}
+
+	r := bytes.NewReader(packet[:])
```
davecheney, line 226: no need to replace here, packet is already a slice

davecheney (changes requested):
Thank you for working on this

I don't see the utility of testing recvPacket for short packets. I would like to see a test against the external API exposed by NewClientPipe. If recvPacket crashes because the input is smaller than expected then this is the fault of its caller. It should be fixed at that point.

packet_test.go
greatroar: Ok, removed.

fuzz.go
greatroar: Done.

client_test.go
greatroar: Holdover from a previous version, removed now.

greatroar:
> I would like to see a test against the external API exposed by NewClientPipe.

Done, and also added a test that crashes the server on current master.

client_test.go
greatroar: stretchr/testify is already a dependency, it just wasn't imported in this file. But I've removed it for now.

davecheney (approved):
LGTM, thank you. I'll leave this for the other maintainers.

packet.go
greatroar: I'm having a very hard time simulating this in a test, but I did notice the following. When I print the error return from io.Copy in TestServerRoughDisconnect,

https://github.com/pkg/sftp/blob/3c22ebff33f246c81ec85ef7fedb04c2bb2da018/client_integration_test.go#L1836

... that prints nil, because io.Copy does not consider EOF an error. Will it help the review for this issue if I remove this check and open a separate issue/PR for this?

client_test.go
puellanivis: I tend to aire on the side of not checking specific errors as well. It is too easy then for the error itself to become a part of your contract, and then it becomes hard to change without breaking tests.

puellanivis:
Probably fine from me as well, but I’d like to get https://github.com/pkg/sftp/pull/387 merged first. Then a rebase would be required here.

packet.go
greatroar: I'll open a separate issue.

client_test.go
greatroar: I've now changed the test to check for err == nil || err == io.EOF.

greatroar:
Rebased on #387.

server_test.go
```diff
@@ -365,3 +366,31 @@ func TestStatNonExistent(t *testing.T) {
...
+		append(validInit, brokenOpen...),
+	} {
+		srv, err := NewServer(struct {
+			io.Reader
+			io.WriteCloser
+		}{
+			bytes.NewReader(clientInput),
+			&sink{},
+		})
+		assert.NoError(t, err)
```
puellanivis, line 390: In the event that there is an error here, `srv == nil`, which means the following line will panic from a nil-pointer deref.

So, this needs to be `require` not `assert`.

greatroar: Done!

client_test.go
```diff
@@ -208,3 +209,22 @@ func TestUseFstatChecked(t *testing.T) {
...
+func (*sink) Close() error                { return nil }
+func (*sink) Write(p []byte) (int, error) { return len(p), nil }
+
+func TestClientZeroLengthPacket(t *testing.T) {
+	// Packet length zero (never valid). This used to crash the client.
+	packet := []byte{0, 0, 0, 0}
+
+	r := bytes.NewReader(packet)
+	c, err := NewClientPipe(r, &sink{})
+	if err == nil || err == io.EOF {
```
puellanivis, line 224: I‘m not so sure how useful it is to check that we didn’t get an `io.EOF` back.

greatroar: Well, it ties in with #390 and makes the test robust in the face of further changes to recvPacket. It's not essential and I can remove it if you want.

puellanivis: I think here, we should just be fine that we get some sort of error, we really don’t care what here.

greatroar: Ok, changed to err == nil.

PR #391, opened 2020-11-04
Detect remote disconnect in File methods

### LINKED ISSUE #390: Remote disconnect not detected
Errors from recvPacket are returned to callers directly from Client methods, including the io.EOF and io.ErrUnexpectedEOF from two calls to io.ReadFull. This means that some errors are not detected at all: in TestServerRoughDisconnect,

https://github.com/pkg/sftp/blob/3c22ebff33f246c81ec85ef7fedb04c2bb2da018/client_integration_test.go#L1822-L1836

the io.Copy in fact returns nil, because it interprets io.EOF as "whole file copied", while actually the server was killed. The same occurs when an SSH child process dies. In the latter case, callers probably Wait on the SSH process and will get a "signal: terminated" error, but they can't know if the last operation on the pipe was successfully completed before SSH exited.

I see two ways of dealing with this. Both need a custom error, say errDisconnect.

**Option 1**: Return errDisconnect instead of EOF (both kinds) when a ReadFull fails in recvPacket. In the client, that signals an actual unexpected behavior, because it's normally the client that disconnects. To keep current Server and RequestServer behavior, their Serve methods then need to translate that error back to io.EOF.

@puellanivis suggested using io.ErrUnexpectedEOF when I first raised this issue in #389, but then a client doing io.ReadFull on an sftp.File would have no way to distinguish an actual premature EOF in a remote file from a disconnect.

**Option 2**: keep recvPacket largely as is, but let Client.recvVersion and clientConn.recv translate any EOFs to errDisconnect. This is probably cleaner.

The only desirable change to recvPacket would, I think, be to translate an io.EOF from the second ReadFull,

https://github.com/pkg/sftp/blob/3c22ebff33f246c81ec85ef7fedb04c2bb2da018/packet.go#L160

into an ErrUnexpectedEOF. This EOF occurs in the middle of a packet, which is a protocol error. Only the Servers would notice, though.

greatroar:
Fixes #390. Uses @puellanivis's patch, adds a test and adds assertions to existing tests.

### DIFF
--- a/client_integration_test.go
+++ b/client_integration_test.go
@@ -1872,7 +1872,8 @@ func TestServerRoughDisconnect(t *testing.T) {
 		cmd.Process.Kill()
 	}()
 
-	io.Copy(ioutil.Discard, f)
+	_, err = io.Copy(ioutil.Discard, f)
+	assert.Error(t, err)
 }
 
 // sftp/issue/181, abrupt server hangup would result in client hangs.
@@ -1930,7 +1931,8 @@ func TestServerRoughDisconnect3(t *testing.T) {
 		cmd.Process.Kill()
 	}()
 
-	io.Copy(rf, lf)
+	_, err = io.Copy(rf, lf)
+	assert.Error(t, err)
 }
 
 // sftp/issue/234 - also affected Write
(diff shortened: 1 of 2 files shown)


puellanivis (approved):

drakkan:
LGTM, thank you!

client_integration_test.go
```diff
@@ -1962,19 +1964,50 @@ func TestServerRoughDisconnect4(t *testing.T) {
...
+	}
+	defer f.Close()
+	go func() {
+		time.Sleep(100 * time.Millisecond)
+		cmd.Process.Kill()
+	}()
+
+	_, err = io.ReadFull(f, make([]byte, 10))
+	assert.Error(t, err)
+	assert.NotEqual(t, io.ErrUnexpectedEOF, err)
```
puellanivis, line 1999: 🤦‍♀️ We were never testing for `io.EOF`. When I went into v2 to ensure this returns a disconnect error, and not an `io.EOF`… it returns an `io.EOF`. 😐 (Adding a `panic(err)` before the asserts.)
```
=== RUN   TestServerRoughDisconnectEOF
--- FAIL: TestServerRoughDisconnectEOF (0.01s)
panic: EOF [recovered, repanicked]

goroutine 2163 [running]:
testing.tRunner.func1.2({0x102702f20, 0x1027ca530})
        /usr/local/go/src/testing/testing.go:1974 +0x1a0
testing.tRunner.func1()
        /usr/local/go/src/testing/testing.go:1977 +0x318
panic({0x102702f20?, 0x1027ca530?})
        /usr/local/go/src/runtime/panic.go:860 +0x12c
github.com/pkg/sftp.TestServerRoughDisconnectEOF(0x606f1d50ab48)
        /src/github.com/pkg/tmp/sftp/client_integration_test.go:2359 +0x178
testing.tRunner(0x606f1d50ab48, 0x10276a6c8)
        /usr/local/go/src/testing/testing.go:2036 +0xc4
created by testing.(*T).Run in goroutine 1
        /usr/local/go/src/testing/testing.go:2101 +0x3a8
exit status 2
FAIL    github.com/pkg/sftp     1.588s
```

I was spinning my gears looking for what was wrong, and it turns out the client really is receiving a `FX_STATUS_EOF` packet. 🙃

puellanivis: https://go.dev/play/p/OQ05dmLiumZ 🤦‍♀️ 
```go
	n, err := io.ReadFull(f, make([]byte, 10))
	if err != nil {
		fmt.Println(n, err)
	}
```
This prints `0 EOF`. 🤷‍♀️

PR #392, opened 2020-11-05
request-server: don't return EOF if there is an unexpected error

drakkan:
io.EOF means [no error](https://github.com/pkg/sftp/blob/master/examples/request-server/main.go#L124) so convert to ErrUnexpectedEOF if we detect a transfer error

### DIFF
--- a/request-server_test.go
+++ b/request-server_test.go
@@ -41,7 +41,7 @@ func clientRequestServerPair(t *testing.T) *csPair {
 	ready := make(chan bool)
 	os.Remove(sock) // either this or signal handling
 	pair := &csPair{
-		svrResult: make(chan error),
+		svrResult: make(chan error, 1),
 	}
 	var server *RequestServer
 	go func() {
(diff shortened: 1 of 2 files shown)


davecheney (changes requested):
Please add a test. Thank you

drakkan:
> Please add a test. Thank you

Done, thank you

request-server_test.go
```diff
@@ -39,6 +40,9 @@ func clientRequestServerPair(t *testing.T) *csPair {
 	skipIfPlan9(t)
 	ready := make(chan bool)
 	os.Remove(sock) // either this or signal handling
+	pair := &csPair{
+		svrResult: make(chan error),
```
puellanivis, line 44: Can you make this with a buffer of 1? That way even if we do not read from this channel, there is not a blocked goroutine left hanging?

drakkan: Done, thank you

puellanivis (approved):
I cannot find anything else to even nitpick about. :laughing:

drakkan:
> I cannot find anything else to even nitpick about.

Thank you!

PR #393, opened 2020-11-16
request server: fix handles leak in error case

drakkan:
For open/opendir requests we create a new handle and close it
when the client sends an SSH_FXP_CLOSE but if we return an error the
client will never send the close packet so we have to close the handle
ourselves.

You can easily see the bug running the modified test cases without the patch.

The open handles will be closed when the client closes the whole connection, but if we have a client connected for a long time we accumulate handles every time we return an error for open/opendir.

This bug is more evident after applying #392 since the `EOF` will be converted to `ErrUnexpectedEOF` if open requests are found when the client disconnects

### DIFF
--- a/request-server.go
+++ b/request-server.go
@@ -198,12 +198,20 @@ func (rs *RequestServer) packetWorker(
 			rpkt = cleanPacketPath(pkt)
 		case *sshFxpOpendirPacket:
 			request := requestFromPacket(ctx, pkt)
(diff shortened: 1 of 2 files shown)


puellanivis (approved):

drakkan:
@puellanivis thank you for approving. What do you think about #392? After fixing the open connection tracking in this pull request #392 allows to return the correct exit status to the clients

PR #394, opened 2020-11-17
client: add WriteAt

drakkan:
this is very similar to how ReadAt is already implemented.

Since we use WriteAt inside Write current test cases already cover
this method


drakkan:
@puellanivis I just noticed you added `WriteAt` in your `more-optimization` branch. I tested that branch and it seems a great work, thank you! So closing this pull request

puellanivis:
Yeah, I need to get on that branch. :laughing:

PR #402, opened 2020-12-08
CI: switch to GitHub Actions

drakkan:
GitHub actions have some missing features compared to Travis but they are much faster.

Please take a look at the build times here:

https://github.com/drakkan/sftp/actions

the build starts as soon as we push a new commit, travis seems to wait more than hour before starting the build (at least today)


drakkan:
Probably travis will work as before if we switch to travis.com but before switching please read here:

https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing

gerrith3:
@drakkan  Read this one as well:  https://blog.travis-ci.com/oss-announcement - and let me know if you want help escalating any additional request for credits.

drakkan:
GitHub actions have some missing features compared to travis, for example:

- allow failure is different from travis, this could be an issue for the go tip build, see [here](https://github.com/actions/toolkit/issues/399)
- only amd64 is supported, to build for other archs you have to use qemu or similar tricks. For example see [here](https://github.com/drakkan/sftpgo/tree/master/.github/workflows) I use xgo for cross building and qemu for arm64 and ppc64le docker images.

I have no particular preferences, Travis is fine for me, I just want a CI that works and currently doesn't.
@puellanivis @eikenb @davecheney, other maintainers, what do you think about? Thank you

puellanivis:
:( I hate to hear that it’s not supporting architectures other than amd64… but I suppose that it’s the worst thing ever.

drakkan:
> :( I hate to hear that it’s not supporting architectures other than amd64… but I suppose that it’s the worst thing ever.

If we decide to switch to GitHub actions I can try to run the test cases on arm64 and ppc64le using an action such as this [one](https://github.com/uraimo/run-on-arch-action) or something similar. 

Let's wait a few more days to see if other maintainers want to help choose.

drakkan:
> :( I hate to hear that it’s not supporting architectures other than amd64… but I suppose that it’s the worst thing ever.

I made some attempts to run test cases on arm64 and ppc64le too, [here](https://github.com/drakkan/sftp/blob/master/.github/workflows/CI.yml) is the updated workflow and [here](https://github.com/drakkan/sftp/actions/runs/411583172) you can view the results. 

There are issues with the race detector tests:

- on arm64 I have this error `FATAL: ThreadSanitizer: unsupported VMA range`, I have the same error on real hw so I think race detector is not supported on arm64
- on ppc64le the race detector works on real hw but it timeouts on qemu, see [here](https://github.com/golang/go/issues/42080)

I'm not sure if push these changes here too, thank you

.github/workflows/CI.yml
```diff
@@ -27,15 +27,40 @@ jobs:
...
+        name: Run tests
+        with:
+          arch: ${{ matrix.arch }}
+          distro: ${{ matrix.distro }}
+          shell: /bin/bash
+
+          install: |
+            apt-get update -q -y
+            apt-get install -q -y curl gcc make openssh-sftp-server
+            GO_VERSION=$(curl https://golang.org/VERSION?m=text)
```
puellanivis, line 60: So, we run the integration tests for 1.14 and 1.15 for x86_64, and then run the other architectures only on the most recent version? (Which would jump automatically at 1.16?)

drakkan: Yes, correct

davecheney:
Travis is dead, I don’t see any value in not switching to GitHub actions and feel confident that the behaviour of this library is not architecture dependant so shouldn’t suffer from being tested on one architecture

drakkan:
> Travis is dead, I don’t see any value in not switching to GitHub actions and feel confident that the behaviour of this library is not architecture dependant so shouldn’t suffer from being tested on one architecture

I agree, I think it would be more useful to try running CI on Windows but this is a separate issue.

do we want to switch to GitHub actions and only support amd64? @puellanivis  do you agree?

puellanivis:
I am ambivalent about switching. I have hardly noticed the functionality on the repo anyways. 😆 Since travis is dying, let’s go ahead and move.

drakkan:
> I am ambivalent about switching. I have hardly noticed the functionality on the repo anyways.  Since travis is dying, let’s go ahead and move.

Ok, I'll remove arm and ppc64 and arm support from this commit and I'll push the change later today or tomorrow, thank you

PR #410, opened 2021-03-06
Export RealPath for potential client use

puellanivis:
Real simple, it’s exactly what the title says, nothing more.

### DIFF
--- a/client.go
+++ b/client.go
@@ -789,7 +789,7 @@ func (c *Client) PosixRename(oldname, newname string) error {
 	}
 }
 
-func (c *Client) realpath(path string) (string, error) {
+func (c *Client) RealPath(path string) (string, error) {
 	id := c.nextID()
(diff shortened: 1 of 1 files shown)


puellanivis:
As mentioned I went through the path handling code. It’s all generally server-side stuff. Nothing necessary to be done on the client side.

The standard even says this request packet can be called with essentially any arbitrary path.

drakkan:
Great! I suggest to add a comment to avoid warnings like this one `exported method Client.RealPath should have comment or be unexported`.

drakkan:
server side the method just calls [cleanPacketPath](https://github.com/pkg/sftp/blob/master/request-server.go#L266) so specific test cases should not be needed

puellanivis:
Feel free to merge, and then cut a release at your leisure.

PR #411, opened 2021-03-10
client fix potential crash if we receive a short packet

drakkan:
This was discovered using  continuous fuzzing, see [here](https://github.com/google/oss-fuzz/pull/5324)

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -228,3 +228,14 @@ func TestClientZeroLengthPacket(t *testing.T) {
 		c.Close()
 	}
 }
+
+func TestClientShortPacket(t *testing.T) {
+	// init packet too short.
+	packet := []byte{0, 0, 0, 1, 2}
+
+	r := bytes.NewReader(packet)
+	_, err := NewClientPipe(r, &sink{})
+	if !errors.Is(err, errShortPacket) {
+		t.Fatalf("expected error: %v, got: %v", errShortPacket, err)
+	}
+}
--- a/client.go
+++ b/client.go
@@ -273,7 +273,10 @@ func (c *Client) recvVersion() error {
 		return &unexpectedPacketErr{sshFxpVersion, typ}
 	}
 
-	version, data := unmarshalUint32(data)
+	version, data, err := unmarshalUint32Safe(data)
+	if err != nil {
+		return err
+	}
 	if version != sftpProtocolVersion {
 		return &unexpectedVersionErr{sftpProtocolVersion, version}
 	}


client_test.go
```diff
@@ -228,3 +229,12 @@ func TestClientZeroLengthPacket(t *testing.T) {
...
 	}
 }
+
+func TestClientShortPacket(t *testing.T) {
+	// init packet too short.
+	packet := []byte{0, 0, 0, 1, 2}
+
+	r := bytes.NewReader(packet)
+	_, err := NewClientPipe(r, &sink{})
+	require.Equal(t, errShortPacket, err)
```
puellanivis, line 239: I don’t think we need to use `require.Equal` here, and in particular this usage should probably be “considered harmful”. Since this is essentially equivalent to `if err != errShortPacket { … }`.

While the `testify` packages do offer `require.EqualError` to make an additional assertion that it must be an error, it’s working through string compares, which is equally bad.

Instead, the simplest dumbest use is now overwhelmingly the best recommended practice:
```
if errors.is(err, errShortPacket) {
  …
}
```

client_test.go
```diff
@@ -228,3 +229,12 @@ func TestClientZeroLengthPacket(t *testing.T) {
...
 	}
 }
+
+func TestClientShortPacket(t *testing.T) {
+	// init packet too short.
+	packet := []byte{0, 0, 0, 1, 2}
+
+	r := bytes.NewReader(packet)
+	_, err := NewClientPipe(r, &sink{})
+	require.True(t, errors.Is(err, errShortPacket))
```
puellanivis, line 239: The output of this `testify` “convenience function” is not actually at all useful:
```
--- FAIL: TestClientShortPacket (#.##s)
     client_test.go:239
                          Error Trace: client_test.go:239
                          Error:       Should be true
                          Test:        TestClientShortPacket
```

client_test.go
```diff
@@ -228,3 +228,14 @@ func TestClientZeroLengthPacket(t *testing.T) {
...
 }
+
+func TestClientShortPacket(t *testing.T) {
+	// init packet too short.
+	packet := []byte{0, 0, 0, 1, 2}
+
+	r := bytes.NewReader(packet)
+	_, err := NewClientPipe(r, &sink{})
+	if !errors.Is(err, errShortPacket) {
+		t.Fatalf("expected error: %v, got: %v", errShortPacket, err)
```
greatroar, line 239: require.ErrorIs?

drakkan: should be ok, now thank you

greatroar:
There's actually a lot more of these unsafe unmarshalUint32 calls, in ReadDir, openDir, Stat, Lstat, ReadLink, open, RealPath, readChunkAt, writeChunkAt, unmarshalStatus. The fuzzer won't find them, because it connects and immediately disconnects, without doing any actual work.

Rather than fix all of these by doing unmarshalUint32Safe, maybe recvPacket should take a minSize argument and return errShortPacket if it gets less than minSize bytes?

client_test.go
puellanivis: 🤷‍♀️ the benefits of using of the `require` package here was always really low.

I know everyone wants to use them, because everyone else does, but you always have to weigh cost/benefits. We’re saving just two lines of code, one of which is just a `}`, and the other is offset by including the `require` package in the first place.

And this `require.ErrorIs` just obscures the `errors.Is()` call from regexps, etc as well.

I’ve had to do significant cleanups on the testing codebase before, because there were a ton of `assert` instead of `require`, where if we just wrote things out clearly, we never would have had that issue. `t.Fatal("unexpected error", err)` is easy to see what it does and how, while `require.NoError(t, err)` vs `assert.NoError(t, err)` does not make it clear which one should be properly used.

If a set of test code is already using that package, then 🤷‍♀️ I’m not going to break convention and start using it. But there’s no particularly compelling reason why we need to drag it newly into every single test file just to save one or two lines.

puellanivis:
> There's actually a lot more of these unsafe unmarshalUint32 calls, in ReadDir, openDir, Stat, Lstat, ReadLink, open, RealPath, readChunkAt, writeChunkAt, unmarshalStatus. The fuzzer won't find them, because it connects and immediately disconnects, without doing any actual work.

In this case, let’s not worry too much about it. I don’t think we could easily push a `minSize` argument into the recvPacket. And properly addressing this issue might benefit enormously from stepping back, looking at what we’re doing, and using a new paradigm.

puellanivis (approved):

PR #413, opened 2021-03-10
Check for '/', not os.IsPathSeparator, in MkdirAll

greatroar:
The SFTP protocol uses '/' as the path separator, always, and the behavior should not depend on the client platform.

### DIFF
--- a/client.go
+++ b/client.go
@@ -864,12 +864,12 @@ func (c *Client) MkdirAll(path string) error {
 
 	// Slow path: make sure parent exists and then call Mkdir for path.
 	i := len(path)
-	for i > 0 && os.IsPathSeparator(path[i-1]) { // Skip trailing path separator.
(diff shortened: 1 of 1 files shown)


drakkan:
LGTM

client.go
```diff
@@ -864,12 +864,12 @@ func (c *Client) MkdirAll(path string) error {
 
 	// Slow path: make sure parent exists and then call Mkdir for path.
 	i := len(path)
-	for i > 0 && os.IsPathSeparator(path[i-1]) { // Skip trailing path separator.
+	for i > 0 && path[i-1] == '/' { // Skip trailing path separator.
 		i--
 	}
```
puellanivis, line 869: Hm… we could probably just make use of `path.Split(…)` instead, rather than doing this raw byte search ourselves. This is probably fine as is though.

puellanivis (approved):

Issue #335, opened 2020-02-12
client: Is there any way to create a file on the server with specific permissions?


toby-allsopp:
It looks like neither `Client.Create` or `Clilent.Open` provide a way to specify the file mode (e.g. 0640) when creating a file. I see that I can use `Chmod` once the file exists, but that leaves a window open where the file is accessible using the default permissions. Is there some way around this?

eikenb:
Currently no. The protocol supports sending file attributes along with the Open packet, but the client implementation doesn't utilize that presently.

I don't see anything preventing it from being added, though it would need to be a new method so as not to break backwards compatibility. CreateWithMode() or some such.

I'd be happy to review PRs for this or we could convert this ticket into a feature request.

toby-allsopp:
We've postponed the feature I was working on that required this feature, so I won't have time to work on it in the near future. If/when we pick it up again I'd be happy to create a PR but until then it's probably best to mark it as a feature request. Thanks!

puellanivis:
Hm, we probably should have made https://pkg.go.dev/github.com/pkg/sftp?tab=doc#Client.OpenFile support perm parallel to `os.OpenFile`. 🤷‍♀ Too late now, though.

But I would recommend expanding `Client.OpenFileWithPerm` or `WithMode` instead, or some other way to specify more options that exclusively from the default flags in `Client.Create`. Or perhaps using `Client.CreateWithMode(filename, flag, perm)` which does `flag |= os.O_CREATE` or something?

Kagami:
I had a similar problem with this package: it creates files with 664 permissions and directories with 775, even though umask is 022.

I found a fix though: replace `ForceCommand internal-sftp` with `ForceCommand internal-sftp -u 022` in `/etc/ssh/sshd_config` and restart sshd.

puellanivis:
Where was the umask? On the client? Because the server code runs on a completely different server, and does not automatically respect umask from the client.

Kagami:
You're right, it turned out that target user had `umask 002` (due to `USERGROUPS_ENAB yes` in `/etc/login.defs`).

I was confused by sftp executable because it preserves permissions.

puellanivis:
No problems. There’s a lot to keep track of when using this package.

PR #414, opened 2021-03-11
Support os.Mode{Setuid,Setgid,Sticky} in Chmod

greatroar:
Previously, these bits were ignored by Chmod, which sent the numerical value of the mode argument as-is to the server. As a result, callers had to supply POSIX values for setuid, setgid and sticky to Chmod.

The new version supports both the POSIX values and the Go values, at the expense of some code duplication.

Also added a note to the docs to clarify that the umask is not subtracted from the mode, and why that is. The only portable way to get the umask is to set it, then reset it, but that's racy. On Linux, we could parse /proc/self/status, but that doesn't work portably and will fail where /proc is not available (some Docker containers, notably). Ref #335.

### DIFF
--- a/client.go
+++ b/client.go
@@ -560,8 +560,12 @@ func (c *Client) Chown(path string, uid, gid int) error {
 }
 
 // Chmod changes the permissions of the named file.
+//
+// Chmod does not apply a umask, because even retrieving the umask is not
+// possible in a portable way without causing a race condition. Callers
+// should mask off umask bits, if desired.
 func (c *Client) Chmod(path string, mode os.FileMode) error {
-	return c.setstat(path, sshFileXferAttrPermissions, uint32(mode))
+	return c.setstat(path, sshFileXferAttrPermissions, toChmodPerm(mode))
 }
 
 // Truncate sets the size of the named file. Although it may be safely assumed
@@ -1719,6 +1723,8 @@ func (f *File) Chown(uid, gid int) error {
 }
 
 // Chmod changes the permissions of the current file.
+//
+// See Client.Chmod for details.
 func (f *File) Chmod(mode os.FileMode) error {
 	return f.c.Chmod(f.path, mode)
 }
@@ -1829,4 +1835,26 @@ func flags(f int) uint32 {
 		out |= sshFxfExcl
 	}
 	return out
+}
+
+// toChmodPerm converts Go permission bits to POSIX permission bits.
+//
+// This differs from fromFileMode in that we preserve the POSIX versions of
+// setuid, setgid and sticky in m, because we've historically supported those
+// bits, and we mask off any non-permission bits.
+func toChmodPerm(m os.FileMode) (perm uint32) {
+	const mask = os.ModePerm | s_ISUID | s_ISGID | s_ISVTX
+	perm = uint32(m & mask)
+
+	if m&os.ModeSetuid != 0 {
+		perm |= s_ISUID
+	}
+	if m&os.ModeSetgid != 0 {
+		perm |= s_ISGID
+	}
+	if m&os.ModeSticky != 0 {
+		perm |= s_ISVTX
+	}
+
+	return perm
 }
--- a/client_integration_test.go
+++ b/client_integration_test.go
@@ -873,6 +873,59 @@ func TestClientChmodReadonly(t *testing.T) {
 	}
 }
 
+func TestClientSetuid(t *testing.T) {
+	skipIfWindows(t) // No UNIX permissions.
+	if *testServerImpl {
+		t.Skipf("skipping with -testserver")
+	}
+
+	sftp, cmd := testClient(t, READWRITE, NODELAY)
+	defer cmd.Wait()
+	defer sftp.Close()
+
+	f, err := ioutil.TempFile("", "sftptest-setuid")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.Remove(f.Name())
+	f.Close()
+
+	const allPerm = os.ModePerm | os.ModeSetuid | os.ModeSetgid | os.ModeSticky |
+		s_ISUID | s_ISGID | s_ISVTX
+
+	for _, c := range []struct {
+		goPerm    os.FileMode
+		posixPerm uint32
+	}{
+		{os.ModeSetuid, s_ISUID},
+		{os.ModeSetgid, s_ISGID},
+		{os.ModeSticky, s_ISVTX},
+		{os.ModeSetuid | os.ModeSticky, s_ISUID | s_ISVTX},
+	} {
+		goPerm := 0700 | c.goPerm
+		posixPerm := 0700 | c.posixPerm
+
+		err = sftp.Chmod(f.Name(), goPerm)
+		require.NoError(t, err)
+
+		info, err := sftp.Stat(f.Name())
+		require.NoError(t, err)
+		require.Equal(t, goPerm, info.Mode()&allPerm)
+
+		err = sftp.Chmod(f.Name(), 0700) // Reset funny bits.
+		require.NoError(t, err)
+
+		// For historical reasons, we also support literal POSIX mode bits in
+		// Chmod. Stat should still translate these to Go os.FileMode bits.
+		err = sftp.Chmod(f.Name(), os.FileMode(posixPerm))
+		require.NoError(t, err)
+
+		info, err = sftp.Stat(f.Name())
+		require.NoError(t, err)
+		require.Equal(t, goPerm, info.Mode()&allPerm)
+	}
+}
+
 func TestClientChown(t *testing.T) {
 	skipIfWindows(t) // No UNIX permissions.
 	sftp, cmd := testClient(t, READWRITE, NODELAY)
--- a/stat_plan9.go
+++ b/stat_plan9.go
@@ -94,3 +94,11 @@ func fromFileMode(mode os.FileMode) uint32 {
 
 	return ret
 }
+
+// Plan 9 doesn't have setuid, setgid or sticky, but a Plan 9 client should
+// be able to send these bits to a POSIX server.
+const (
+	s_ISUID = 04000
+	s_ISGID = 02000
+	S_ISVTX = 01000
+)
--- a/stat_posix.go
+++ b/stat_posix.go
@@ -114,3 +114,9 @@ func fromFileMode(mode os.FileMode) uint32 {
 
 	return ret
 }
+
+const (
+	s_ISUID = syscall.S_ISUID
+	s_ISGID = syscall.S_ISGID
+	s_ISVTX = syscall.S_ISVTX
+)


greatroar:
Tests fail on Travis, while they work locally. May be a capability issue (?)

greatroar:
Got it. The test works with sftp, but not with -testserver.

client_integration_test.go
```diff
@@ -873,6 +873,44 @@ func TestClientChmodReadonly(t *testing.T) {
...
+	f.Close()
+
+	const perm = 0700 | os.ModeSetuid
+
+	err = sftp.Chmod(f.Name(), perm)
+	require.NoError(t, err)
+
+	info, err := sftp.Stat(f.Name())
+	require.NoError(t, err)
+	require.Equal(t, perm, info.Mode()&perm)
```
puellanivis, line 900: This test is insufficient. `info.Mode()&perm` would not catch spare additional bits in `info.Mode()` So if our implementation were bugged and applied `04755` to the file, while this should fail, here it would not.

client.go
```diff
@@ -1830,3 +1836,33 @@ func flags(f int) uint32 {
...
+	const (
+		posixSetuid = 04000
+		posixSetgid = 02000
+		posixSticky = 01000
+		posixBits   = 07777
+	)
+
+	posix := m & posixBits
+
+	switch {
```
puellanivis, line 1858: This switch means only one of these cases will be followed. If I pass in `m = os.ModeSetuid | os.ModeSetgid | os.ModeSticky` it will do the wrong thing: only apply `Setuid`.

client.go
```diff
@@ -1830,3 +1836,33 @@ func flags(f int) uint32 {
...
+
+// posixPerm converts Go permission bits to POSIX permission bits,
+// masking off file type bits.
+//
+// This differs from fromFileMode in that it preserves the POSIX versions of
+// setuid, setgid and sticky in m, because we've historically supported those
+// bits.
+func posixPerm(m os.FileMode) uint32 {
+	// POSIX mode bits, as listed at
+	// https://tools.ietf.org/html/draft-ietf-secsh-filexfer-13#section-7.6
```
puellanivis, line 1848: I would rather us reference the root authority for these values: `[IEEE.1003-1.1996]` rather than an indirect standard, which we specifically call out as not implementing.

client.go
```diff
@@ -1830,3 +1836,33 @@ func flags(f int) uint32 {
...
 	return out
 }
+
+// posixPerm converts Go permission bits to POSIX permission bits,
+// masking off file type bits.
+//
+// This differs from fromFileMode in that it preserves the POSIX versions of
+// setuid, setgid and sticky in m, because we've historically supported those
+// bits.
+func posixPerm(m os.FileMode) uint32 {
```
puellanivis, line 1846: I think we could change `fromFileMode` to incorporate this “support `S_ISUID` and `os.ModeSetuid`” in order to get the baseline permissions at the start, then it could go on to the next step of adding the additional `S_IFMT` values in.

It looks like there are bugs in `fromFileMode` anyways, because it could potentially OR in various `S_IFMT` values potentially all together, when only one of them should ever be valid at any one time. (i.e. since they all share the same bit mask, accidentally composing `S_IFREG | S_IFDIR` gets you `S_IFLNK` _not_ what was intended. In this case, it should probably return an `error`.)

client.go
```diff
@@ -1830,3 +1836,33 @@ func flags(f int) uint32 {
...
+	// POSIX mode bits, as listed at
+	// https://tools.ietf.org/html/draft-ietf-secsh-filexfer-13#section-7.6
+	const (
+		posixSetuid = 04000
```
puellanivis, line 1850: This code should probably instead go into `stat_posix.go` and `stat_plan9.go` so that we can use the appropriate `syscall` values the same as `fromFileMode` and handle plan9 appropriately.

client_integration_test.go
```diff
@@ -873,6 +873,44 @@ func TestClientChmodReadonly(t *testing.T) {
...
+	defer sftp.Close()
+
+	f, err := ioutil.TempFile("", "sftptest-setuid")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.Remove(f.Name())
+	f.Close()
+
+	const perm = 0700 | os.ModeSetuid
```
puellanivis, line 893: There are no tests for `Setguid` or `Sticky`, and no tests for more than one of those flags being set at the same time.

client.go
greatroar: The problem with that approach is that we get the bits from the platform the client is running on, rather than the bits that the server expects.

puellanivis: I have checked all these values. Except for `S_IFMT` these values are all the same across all implementations. (They have been after all codifed by POSIX after all.)

client.go
greatroar: Fixed and tested.

client.go
greatroar: I've looked and cited the POSIX spec. Turns out this draft was subtly misrepresenting it.

client_integration_test.go
greatroar: Fixed.

client.go
greatroar: fromFileMode on Plan 9 also doesn't handle the bits 07000 at all, nor does syscall on Plan 9 have the S_ISUID etc. constants. For the server, that's ok since Plan 9 doesn't have setuid, but it means a Plan 9 client talking to a Unix server couldn't set these bits.

Adding them in P9 fromFileMode runs the risk of returning some erroneous bits from Plan 9 servers, if it happens to use 07000 for a different purpose...

client_integration_test.go
```diff
@@ -873,6 +873,65 @@ func TestClientChmodReadonly(t *testing.T) {
...
+
+	f, err := ioutil.TempFile("", "sftptest-setuid")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.Remove(f.Name())
+	f.Close()
+
+	const (
+		unixSetuid = 04000
```
puellanivis, line 894: DRY is not really about avoiding code duplication, it is about avoiding two values/processes that are the same from having inconsistent implementations.

Implementing these two values separately means that there could be inconsistencies between them. This is why I was pushing for the use of the `syscall.S_ISUID` values: we should ideally not be defining these values unless we absolutely have to. (e.g. `S_IFMT` had to be imported locally, because the windows implementation has a bad value. None of _these_ flags are different though, because POSIX has already for a long time established strict values for them.)

client.go
```diff
@@ -1830,3 +1836,41 @@ func flags(f int) uint32 {
 	}
 	return out
 }
+
+// unixPerm converts Go permission bits to Unix permission bits,
+// masking off file type bits.
+//
+// The SFTP protocol draft doesn't specify the exact values of the bits.
+// POSIX is fuzzy in this regard: it specifies bits for the chmod command [1],
+// but not for C applications [2]. We follow the chmod bits, which are the
```
puellanivis, line 1845: I feel like the [2] reference here is essentially irrelevant, the internal details of how the C structures store data is immaterial to the values accepted and used by `chmod`. Any POSIX implementation will need to convert these strictly defined chmod values to any internal implementation regardless of the internal representation being unspecified.

Perhaps naming this `toChownPerms(…)` might be the best choice? This should make it clear that we are converting to chmod-specific permissions, rather than just a nebulous “POSIX permissions”.

client.go
puellanivis: The benefits of putting them into their specific files, is that the `stat_posix.go` file can use the definied `syscall` values, while the `stat_plan9.go` file can define these otherwise-POSIX values individually.

This means the code is _intentional_ rather than assuming that plan9 is unlimitedly the same.

client.go
```diff
@@ -1830,3 +1836,41 @@ func flags(f int) uint32 {
...
+// bits that OpenSSH's sftp-server understands when running on Unix, in the
+// hopes that other implementations simulate its behavior.
+//
+// This differs from fromFileMode in that it preserves the POSIX versions of
+// setuid, setgid and sticky in m, because we've historically supported those
+// bits.
+//
+// [1] https://pubs.opengroup.org/onlinepubs/009604499/utilities/chmod.html
+// [2] https://pubs.opengroup.org/onlinepubs/009604499/basedefs/sys/stat.h.html
+func unixPerm(m os.FileMode) uint32 {
```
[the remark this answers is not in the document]
greatroar: Calling them POSIX bits is misleading, though. The actual bit values aren't really specified by POSIX, as the comment above explains. They're more of a Unix tradition.

client.go
greatroar: The thing is that OpenSSH's sftp-server sends the actual C bits in response to a stat request: https://github.com/openssh/openssh-portable/blob/82fef71e20ffef425b932bec26f5bc46aa1ed41c/sftp-common.c#L78

Whether this is intentional is not explained in the 2001 protocol draft.

client.go
greatroar: The client should do exactly the same on all platforms, though. The syscall package is for talking to your local kernel. This code is talking to a remote.

client_integration_test.go
greatroar: I repeated the constants here because they should have these exact values, regardless of implementation (via syscall or explicitly). It looks ugly, but it's written this way precisely to catch potential inconsistencies.

client.go
greatroar: To put it differently, I'm regarding the value to be sent as part of the *protocol*, and the protocol is conceptually different from the syscall interface. I'll admit the end result is practically the same, except that it also prevents having to write plan9-specific code that I cannot test.

client.go
greatroar: The [old SFTP draft](https://tools.ietf.org/html/draft-ietf-secsh-filexfer-02#section-5) refers explicitly to the POSIX C API, without noticing that it doesn't specify the bit values. Or maybe it in the 1996 edition...

In any case, if you say "stop whining and use sycall", I'll comply. I'll rename the function toChownPerms in any case.

client.go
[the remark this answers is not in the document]
puellanivis: Rest assured, they did get codified to the same values later: https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_stat.h.html

client_integration_test.go
puellanivis: Again, they are codified in POSIX now, and have been set to the correct values… except `S_IFMT` but we resolved that one ourselves.

client.go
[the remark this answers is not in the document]
greatroar: :facepalm: Always check the right version of the spec.

greatroar:
Alright, since the actual values are standardized, there's little reason not to get them from syscall, so I updated the code. Sorry for the noise.

client.go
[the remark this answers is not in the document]
puellanivis: I mean, the SFTP doc absolutely references the versions of the specs that you link to here, but POSIX went ahead and said, “everyone is using these values, because it just makes sense to use the exact same values as the `chmod` command-line tool.” So, they went ahead and just turned what everyone was already doing into formal codification.

I really only know about this because of the `S_IFMT` shenanigans back a bit ago. 😆

puellanivis (approved):

PR #416, opened 2021-03-12
Clean up matching/globbing code

greatroar:
scanChunk and matchChunk were unused leftovers from the old Match implementation, removed in 4f3e725e885c021085d2fb8a9cc26e30ea1a992f.

isPathSeparator contained a useless cast to string.

The documentation (copied from path/filepath) referred to a Separator constant, which is not present in the package.

### DIFF
--- a/match.go
+++ b/match.go
@@ -133,6 +133,5 @@ func Join(elem ...string) string {
 // hasMeta reports whether path contains any of the magic characters
 // recognized by Match.
 func hasMeta(path string) bool {
-	// TODO(niemeyer): Should other magic characters be added here?
-	return strings.ContainsAny(path, "*?[")
+	return strings.ContainsAny(path, "\\*?[")
 }
--- a/match.go
+++ b/match.go
@@ -3,198 +3,39 @@ package sftp
 import (
 	"path"
 	"strings"
-	"unicode/utf8"
 )
 
 // ErrBadPattern indicates a globbing pattern was malformed.
 var ErrBadPattern = path.ErrBadPattern
 
-// Unix separator
-const separator = "/"
-
-// Match reports whether name matches the shell file name pattern.
-// The pattern syntax is:
+// Match reports whether name matches the shell pattern.
 //
-//	pattern:
-//		{ term }
-//	term:
-//		'*'         matches any sequence of non-Separator characters
-//		'?'         matches any single non-Separator character
-//		'[' [ '^' ] { character-range } ']'
-//		            character class (must be non-empty)
-//		c           matches character c (c != '*', '?', '\\', '[')
-//		'\\' c      matches character c
-//
-//	character-range:
-//		c           matches character c (c != '\\', '-', ']')
-//		'\\' c      matches character c
-//		lo '-' hi   matches character c for lo <= c <= hi
-//
-// Match requires pattern to match all of name, not just a substring.
-// The only possible returned error is ErrBadPattern, when pattern
-// is malformed.
-//
-//
+// This is an alias for path.Match from the standard library,
+// offered so that callers need not import the path package.
+// For details, see https://golang.org/pkg/path/#Match.
 func Match(pattern, name string) (matched bool, err error) {
 	return path.Match(pattern, name)
 }
 
 // detect if byte(char) is path separator
 func isPathSeparator(c byte) bool {
-	return string(c) == "/"
+	return c == '/'
 }
 
-// scanChunk gets the next segment of pattern, which is a non-star string
-// possibly preceded by a star.
-func scanChunk(pattern string) (star bool, chunk, rest string) {
-	for len(pattern) > 0 && pattern[0] == '*' {
-		pattern = pattern[1:]
-		star = true
-	}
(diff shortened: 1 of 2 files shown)


match.go
```diff
@@ -18,8 +18,8 @@ const separator = "/"
 //	pattern:
 //		{ term }
 //	term:
-//		'*'         matches any sequence of non-Separator characters
-//		'?'         matches any single non-Separator character
+//		'*'         matches any sequence of non-"/" characters
+//		'?'         matches any single non-"/" character
```
puellanivis, line 22: We can probably match up with https://golang.org/pkg/path/#Match here. (That is no need for quotes.)

match.go
```diff
@@ -9,7 +9,7 @@ import (
 // ErrBadPattern indicates a globbing pattern was malformed.
 var ErrBadPattern = path.ErrBadPattern
 
-// Unix separator
+// The SFTP protocol always uses slash as the path separator.
 const separator = "/"
```
puellanivis, line 13: We might be able to even just remove this entirely? 🤔  The only remaining use is in `cleanGlobPath`.

match.go
```diff
@@ -280,9 +166,8 @@ func (c *Client) glob(dir, pattern string, matches []string) (m []string, e erro
 	return
 }
 
-// Join joins any number of path elements into a single path, adding
-// a Separator if necessary.
-// all empty strings are ignored.
+// Join joins any number of path elements into a single path, separating
```
puellanivis, line 169: Like with `Match` I think we could import the godoc from `path.Join` directly.

match.go
```diff
@@ -178,23 +68,19 @@ func getEsc(chunk string) (r rune, nchunk string, err error) {
...
+// If there is no slash in path, Split returns an empty dir
 // and file set to path.
 // The returned values have the property that path = dir+file.
-func Split(path string) (dir, file string) {
-	i := len(path) - 1
-	for i >= 0 && !isPathSeparator(path[i]) {
-		i--
-	}
-	return path[:i+1], path[i+1:]
+func Split(p string) (dir, file string) {
```
puellanivis, line 76: Like with `Match` I think we could import the godoc from `path.Split` directly.

puellanivis (commented):
👍 good cleanup, maybe we can do a bit more as well?

match.go
greatroar: This one is different because I had to rename the argument to prevent a clash with the path package. (Either that, or change the import, which would make the diff bigger.)

match.go
greatroar: Done. However... we could also replace this by: "Join is a synonym for path.Join. Deprecated: use path.Join instead." (And similarly for Match, Split.) I'm not sure how much sense it makes to keep these synonyms around, given that even the rest of the package uses path functions directly. Only Glob is different.

puellanivis: I think the functions were imported so that users could reliably use the ones in the package, rather than having to import Yet Another Package, which can even sometimes run into issues by accidentally mixing/matching one for the other.

It would be easy for someone to accidentally pick up `filepath` instead of `path` and not notice until someone on Windows tries to use their code.

So, I don’t think it’s necessary to deprecate these, or direct people to other options. But saying something like “this is a convenience wrapper around path.Join, so that you don’t have to import that library with this one. For details see that package.”

match.go
puellanivis: If the text matches up but the only difference is `path p` then it’s fine. 👍 it was just hard to be sure about them without rigorous back-and-forth examination.

match.go
greatroar: Ok. Instead of, or in addition to, the descriptions from path?

puellanivis: Any choice is pretty valid here.

greatroar: I've trimmed the descriptions down to their first sentence, followed by a reference to stdlib documentation at golang.org.

greatroar:
I just removed getEsc as well. It was only called by matchChunk. And I caught a bug in Glob, for which I added a fix and a test.

match.go
```diff
@@ -3,198 +3,39 @@ package sftp
...
-//
-// Match requires pattern to match all of name, not just a substring.
-// The only possible returned error is ErrBadPattern, when pattern
-// is malformed.
-//
+// Match reports whether name matches the shell pattern.
 //
+// This is an alias for path.Match from the standard library, offered so
+// that callers need not import the path package. For details, see
+// https://golang.org/pkg/path/#Match.
```
puellanivis, line 15: [nitpick] Not in any way an actual issue, but I tend to recommend https://rhodesmill.org/brandon/2012/one-sentence-per-line/

I mention this really only because in this particular case splitting this on semantic boundaries gets nearly the same line breaks, but split up the text more coherently:
```
// This is an alias for path.Match from the standard library,
// offered so that callers need not import the path package.
// For details, see https://golang.org/pkg/path/#Match.
```

If you don’t want to fix this nitpick, just say so, and I will merge anyways.

greatroar: Done.

PR #420, opened 2021-03-16
Implement File.Chmod via setfstat

greatroar:
File.Chown is left as a to-do.

### DIFF
--- a/client_integration_test.go
+++ b/client_integration_test.go
@@ -853,6 +853,15 @@ func TestClientChmod(t *testing.T) {
 	} else if stat.Mode()&os.ModePerm != 0531 {
 		t.Fatalf("invalid perm %o\n", stat.Mode())
 	}
+
+	sf, err := sftp.Open(f.Name())
+	require.NoError(t, err)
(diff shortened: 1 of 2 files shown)


puellanivis:
Probably will be super compatible with my refactor, so we can probably squeeze this in easily.

client_integration_test.go
```diff
@@ -853,6 +853,15 @@ func TestClientChmod(t *testing.T) {
 	} else if stat.Mode()&os.ModePerm != 0531 {
 		t.Fatalf("invalid perm %o\n", stat.Mode())
 	}
+
+	sf, err := sftp.Open(f.Name())
+	require.NoError(t, err)
+	defer sf.Close()
```
puellanivis, line 859: Let’s move the `sf.Chmod()` to before this `defer sf.Close()` and then remove the `defer`.

If we add further tests to the end of this function, the `defer` will end up waiting all the way until after that other code, so scoping it more correctly here is probably a good idea.

greatroar: Done.

PR #421, opened 2021-03-17
Improve benchmarks and errors

puellanivis:
Ahead of some other improvements, I wanted to narrow the benchmarks to only relevant code, and this allow necessitated fixing some tests to go along with it.

Fixes:
* `errors.Errorf("constant string")` have been replaced with `errors.New("constant string")`
* `errors.Errorf("constant string: %v", err)` have been replaced with `errors.Wrap(err, "constant string")`
* `sftp.clientConn.conn.sendPacketTest` has been replaced with a `timeBombWriter`, this removes unnecessary test code from `conn`
* all `err != errFakeNet` have been replaced with `!errors.Is(err, errFakeNet)`, this works because we are now properly wrapping errors.
* Error messages starting with a capital letter have been fixed to use a lower-case character.
* `%v` formatting codes that can be more type-safe have been made to be more type specific.
* errors from `runSftpClient`’s `cmd.Wait` have been put into  an `execErr` type that wraps the underlying error.

`packet_test.go` has a lot of changes:
* test tables have been moved into their individual functions, rather than globals
* more marshaling and unmarshaling tests have been added to verify full byte ordering
* where a byte value in a byte-slice is semantically an ASCII character, use `'c'` to make the value clear.
* marshal/unmarshal string can contain NUL bytes, which can always be a sticky point, so add some tests for that
* byte-slices that are marshaled output have semantic line-breaks on each field to make the semantics of the binary data clearer
* `TestRecvPacket` now tests a the short-packet behavior, and the long-packet behavior.
* benchmarks on marshaling now marshal to `ioutil.Discard` as we do not want to benchmark `bytes.Buffer`, we want to benchmark our marshaling code.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,5 +20,8 @@ COUNT ?= 1
 BENCHMARK_PATTERN ?= "."
 
 benchmark:
-	go test -integration -run=NONE -bench=$(BENCHMARK_PATTERN) -benchmem -memprofile memprofile.out -count=$(COUNT)
+	go test -integration -run=NONE -bench=$(BENCHMARK_PATTERN) -benchmem -count=$(COUNT)
+
+benchmark_w_memprofile:
+	go test -integration -run=NONE -bench=$(BENCHMARK_PATTERN) -benchmem -count=$(COUNT) -memprofile memprofile.out
 	go tool pprof -svg -output=memprofile.svg memprofile.out
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,5 @@
+.PHONY: integration integration_w_race benchmark
+
 integration:
 	go test -integration -v ./...
 	go test -testserver -v ./...
@@ -14,4 +16,9 @@ integration_w_race:
 	go test -race -testserver -allocator -v ./...
 	go test -race -integration -allocator -testserver -v ./...
 
+COUNT ?= 1
+BENCHMARK_PATTERN ?= "."
 
+benchmark:
+	go test -integration -run=NONE -bench=$(BENCHMARK_PATTERN) -benchmem -memprofile memprofile.out -count=$(COUNT)
+	go tool pprof -svg -output=memprofile.svg memprofile.out
--- a/client_integration_test.go
+++ b/client_integration_test.go
@@ -6,7 +6,6 @@ package sftp
 import (
 	"bytes"
 	"crypto/sha1"
-	"encoding"
 	"errors"
 	"io"
 	"io/ioutil"
@@ -1490,26 +1489,48 @@ func TestClientReadFrom(t *testing.T) {
 var errFakeNet = errors.New("Fake network issue")
 
 func TestClientReadFromDeadlock(t *testing.T) {
-	clientWriteDeadlock(t, 1, func(f *File) {
-		b := make([]byte, 32768*4)
-		content := bytes.NewReader(b)
-		_, err := f.ReadFrom(content)
-		if err != errFakeNet {
-			t.Fatal("Didn't recieve correct error:", err)
-		}
-	})
+	for i := 0; i < 5; i++ {
+		clientWriteDeadlock(t, i, func(f *File) {
+			b := make([]byte, 32768*4)
+			content := bytes.NewReader(b)
+			_, err := f.ReadFrom(content)
+			if !errors.Is(err, errFakeNet) {
+				t.Fatal("Didn't recieve correct error:", err)
+			}
+		})
+	}
 }
 
 // Write has exact same problem
 func TestClientWriteDeadlock(t *testing.T) {
-	clientWriteDeadlock(t, 1, func(f *File) {
-		b := make([]byte, 32768*4)
+	for i := 0; i < 5; i++ {
+		clientWriteDeadlock(t, i, func(f *File) {
+			b := make([]byte, 32768*4)
 
-		_, err := f.Write(b)
-		if err != errFakeNet {
-			t.Fatal("Didn't recieve correct error:", err)
-		}
-	})
+			_, err := f.Write(b)
+			if !errors.Is(err, errFakeNet) {
+				t.Fatal("Didn't recieve correct error:", err)
+			}
+		})
+	}
+}
+
+type timeBombWriter struct {
+	count int
+	w     io.WriteCloser
+}
+
+func (w *timeBombWriter) Write(b []byte) (int, error) {
+	if w.count < 1 {
+		return 0, errFakeNet
+	}
+
+	w.count--
+	return w.w.Write(b)
+}
+
+func (w *timeBombWriter) Close() error {
+	return w.w.Close()
 }
 
 // shared body for both previous tests
@@ -1534,20 +1555,13 @@ func clientWriteDeadlock(t *testing.T, N int, badfunc func(*File)) {
 	}
 	defer w.Close()
 
-	// Override sendPacket with failing version
-	// Replicates network error/drop part way through (after 1 good packet)
-	count := 0
-	sendPacketTest := func(w io.Writer, m encoding.BinaryMarshaler) error {
-		count++
-		if count > N {
-			return errFakeNet
-		}
-		return sendPacket(w, m)
+	// Override the clienConn Writer with a failing version
+	// Replicates network error/drop part way through (after N good writes)
+	wrap := sftp.clientConn.conn.WriteCloser
+	sftp.clientConn.conn.WriteCloser = &timeBombWriter{
+		count: N,
+		w:     wrap,
 	}
-	sftp.clientConn.conn.sendPacketTest = sendPacketTest
-	defer func() {
-		sftp.clientConn.conn.sendPacketTest = nil
-	}()
 
 	// this locked (before the fix)
 	badfunc(w)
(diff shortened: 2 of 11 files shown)


packet_test.go
```diff
@@ -297,49 +459,49 @@ func TestSSHFxpOpenPackethasPflags(t *testing.T) {
 	}
 }
 
-func BenchmarkMarshalInit(b *testing.B) {
+func benchMarshal(b *testing.B, packet encoding.BinaryMarshaler) {
```
puellanivis, line 462: Should we reuse the `writeToBuffer` here to also include copying the marshalled data into a buffer, while also not testing the `bytes.Buffer` object?

:thinking: I can see arguments either way… but I think we don’t want to. Copying the data out of the marshaled binary data is essentially irrelevant to the benchmarks we really want to be testing here.

In either case, we should ideally `b.ResetTimer()` here so as to not include the setup to the benchmark. (It’s a pretty trivial amount of work compared to the fast loop below, but principles.)

packet_test.go
```diff
@@ -130,85 +184,193 @@ var unmarshalStringTests = []struct {
...
-	p    encoding.BinaryMarshaler
-	want []byte
-}{
-	{&sshFxInitPacket{
-		Version: 3,
-		Extensions: []extensionPair{
-			{"posix-rename@openssh.com", "1"},
-		},
-	}, []byte{0x0, 0x0, 0x0, 0x26, 0x1, 0x0, 0x0, 0x0, 0x3, 0x0, 0x0, 0x0, 0x18, 0x70, 0x6f, 0x73, 0x69, 0x78, 0x2d, 0x72, 0x65, 0x6e, 0x61, 0x6d, 0x65, 0x40, 0x6f, 0x70, 0x65, 0x6e, 0x73, 0x73, 0x68, 0x2e, 0x63, 0x6f, 0x6d, 0x0, 0x0, 0x0, 0x1, 0x31}},
+type nopCloserBuffer struct {
```
puellanivis, line 220: This is a vestigial type from internal refactors. It should be removed.

packet_test.go
```diff
@@ -130,85 +184,193 @@ var unmarshalStringTests = []struct {
...
+			},
 		},
-	}), sshFxpInit, []byte{0x0, 0x0, 0x0, 0x3, 0x0, 0x0, 0x0, 0x18, 0x70, 0x6f, 0x73, 0x69, 0x78, 0x2d, 0x72, 0x65, 0x6e, 0x61, 0x6d, 0x65, 0x40, 0x6f, 0x70, 0x65, 0x6e, 0x73, 0x73, 0x68, 0x2e, 0x63, 0x6f, 0x6d, 0x0, 0x0, 0x0, 0x1, 0x31}},
-}
+		{
+			b: []byte{
+				0x0, 0x0, 0x0, 0x0,
+			},
+			wantErr: errShortPacket,
+		},
```
puellanivis, line 351: Let’s add another test here for `errLongPacket`, we can use `0xff` for all four of the bytes, that should trip an `errLongPacket` for as long as `errLongPacket` has any meaning.

client_integration_test.go
```diff
@@ -2466,7 +2498,9 @@ func benchmarkWriteTo(b *testing.B, bufsize int, delay time.Duration) {
 	b.ResetTimer()
 	b.SetBytes(int64(size))
 
-	buf := new(bytes.Buffer)
+	buf := &writeToBuffer{
+		b: make([]byte, 0, size),
+	}
```
puellanivis, line 2503: This should go before the `b.ResetTimer()` so that it is not included in the memory benchmark stats.

client_integration_test.go
```diff
@@ -2444,6 +2454,28 @@ func BenchmarkReadFrom4MiBDelay150Msec(b *testing.B) {
 	benchmarkReadFrom(b, 4*1024*1024, 150*time.Millisecond)
 }
 
+// writeToBuffer implements the relevant parts of bytes.Buffer,
+// but does not release its internal buffer when Reset.
+//
+// Release its internal memory when Reset is good for avoiding memory leaks,
+// but not great for memory benchmarks, as this fills up a lot of irrelevant allocations.
+type writeToBuffer struct {
```
puellanivis, line 2462: :thinking: WTF? If I look at `bytes.Buffer.Reset` it keeps its backing buffer the same as we do here.

Yet, without this, I was seeing significant memory allocation in `bytes.Buffer.Grow()` :thinking:

puellanivis: Yep, looks like it was growing from somewhere else entirely.

packet_test.go
puellanivis: So, the `copy(…)` command is going to take the same amount of time regardless of the implementation, even if split into two `bytes.Buffer.Write` calls. Adding that overhead into the benchmark actually overwhelms the time taken to prep the byte-slices to be written, which is really what we want to be benchmarking here.

puellanivis:
Benchmarks between this and the current main branch are discontinuous, i.e. the tests have changed enough that any change in values is (mostly†) purely noise.

†: There is a small change here that has a surprising impact on performance. Removing the `if sendTestPacket != nil { … }` is pretty significant considering its part of every single packet sent. And since this was only ever for test code, reworking the tests to remove it is a straight and simple win.

drakkan:
> Benchmarks between this and the current main branch are discontinuous, i.e. the tests have changed enough that any change in values is (mostly†) purely noise.
> 
> †: There is a small change here that has a surprising impact on performance. Removing the `if sendTestPacket != nil { … }` is pretty significant considering its part of every single packet sent. And since this was only ever for test code, reworking the tests to remove it is a straight and simple win.

I have noticed the same thing in other projects in the past, it is precisely for this reason that in # 407 I wrote:

> If we want to change them we should do it in a way that does not require to checking if an interface is implemented for each packet (for example for writes)

Makefile
```diff
@@ -14,4 +16,9 @@ integration_w_race:
 	go test -race -testserver -allocator -v ./...
 	go test -race -integration -allocator -testserver -v ./...
 
+COUNT ?= 1
+BENCHMARK_PATTERN ?= "."
 
+benchmark:
+	go test -integration -run=NONE -bench=$(BENCHMARK_PATTERN) -benchmem -memprofile memprofile.out -count=$(COUNT)
+	go tool pprof -svg -output=memprofile.svg memprofile.out
```
greatroar, line 24: Isn't -memprofile expensive? I always benchmark without profiling if I can avoid it.

greatroar:
> Removing the `if sendTestPacket != nil { … }` is pretty significant considering its part of every single packet sent.

Just out of curiosity, do you have a figure on that? I just took a diff of the asm, and the only differences are a memory access that should go to the L1 cache and a predictable jump.

sftp.go
```diff
@@ -239,7 +239,7 @@ func getSupportedExtensionByName(extensionName string) (sshExtensionPair, error)
 			return supportedExtension, nil
 		}
 	}
-	return sshExtensionPair{}, fmt.Errorf("Unsupported extension: %v", extensionName)
+	return sshExtensionPair{}, fmt.Errorf("unsupported extension: %s", extensionName)
```
greatroar, line 242: This is a fmt.Errorf, while the rest of the file uses errors.Errorf.

Makefile
puellanivis: For the full sweet of benchmarks, it was 4m 43.640s with memprofile, and 4m 42.839s without memprofile.

Less than a second on a nearly 5 minute run time isn’t so bad?

sftp.go
puellanivis: Good point. Mostly, I didn’t want to disturb things too far, but this is probably a good candidate for switching.

puellanivis:
> Just out of curiosity, do you have a figure on that? I just took a diff of the asm, and the only differences are a memory access that should go to the L1 cache and a predictable jump.

Generating benchmarks now…

Makefile
puellanivis: I’ll switch the `Makefile` to have both options though, same as the above `_w_race` but `_w_memprofile`.

client_integration_test.go
```diff
@@ -2463,11 +2473,11 @@ func benchmarkWriteTo(b *testing.B, bufsize int, delay time.Duration) {
 	f.Write(data)
 	f.Close()
 
+	buf := new(bytes.Buffer)
```
greatroar, line 2476: bytes.NewBuffer(make([]byte, 0, size)) would remove some variance, as the first run no longer needs to allocate.

puellanivis: Indeed, I thought of that. I guess since I’m fixing small details elsewhere I should cover this one too, right? 😆 

👍 on keeping me on the straight and narrow!

puellanivis:
old is with the `if`, new is without.
```
name                          old time/op    new time/op    delta
Read1k-12                       52.1ms ± 3%    51.2ms ± 1%  -1.76%  (p=0.027 n=10+8)
Read16k-12                      6.75ms ± 2%    6.54ms ± 1%  -3.12%  (p=0.000 n=8+9)
Read32k-12                      7.08ms ±32%    5.10ms ±10%    ~     (p=0.105 n=10+10)
Read128k-12                     3.49ms ± 4%    3.64ms ± 4%  +4.37%  (p=0.001 n=10+10)
Read512k-12                     3.32ms ± 2%    3.40ms ± 1%  +2.40%  (p=0.000 n=9+8)
Read1MiB-12                     3.70ms ± 2%    3.76ms ± 3%  +1.59%  (p=0.011 n=10+10)
Read4MiB-12                     4.34ms ± 1%    4.41ms ± 3%    ~     (p=0.079 n=9+10)
Read4MiBDelay10Msec-12          69.5ms ± 1%    69.6ms ± 1%    ~     (p=0.436 n=10+10)
Read4MiBDelay50Msec-12           311ms ± 0%     310ms ± 0%    ~     (p=0.218 n=10+10)
Read4MiBDelay150Msec-12          912ms ± 0%     911ms ± 0%  -0.06%  (p=0.005 n=7+10)
Write1k-12                      98.1ms ± 7%    98.8ms ± 6%    ~     (p=0.579 n=10+10)
Write16k-12                     12.2ms ± 6%    12.1ms ± 3%    ~     (p=0.631 n=10+10)
Write32k-12                     9.03ms ± 5%    9.20ms ±10%    ~     (p=0.661 n=9+10)
Write128k-12                    9.05ms ± 7%    9.16ms ± 8%    ~     (p=0.549 n=9+10)
Write512k-12                    9.30ms ± 5%    9.51ms ± 9%    ~     (p=0.156 n=9+10)
Write1MiB-12                    9.30ms ± 6%    9.17ms ± 5%    ~     (p=0.604 n=10+9)
Write4MiB-12                    9.41ms ± 7%    9.45ms ±12%    ~     (p=0.739 n=10+10)
Write4MiBDelay10Msec-12          3.37s ± 0%     3.37s ± 0%    ~     (p=0.549 n=10+9)
Write4MiBDelay50Msec-12          16.3s ± 0%     16.3s ± 0%    ~     (p=0.400 n=10+9)
Write4MiBDelay150Msec-12         48.6s ± 0%     48.6s ± 0%    ~     (p=0.393 n=10+10)
```

I‘ve cut out most of the metrics, because they’re largely the same. Actually, things are looking generally so fast that the difference that are there look to be mostly just noise.

I would be more inclined to trust the smaller buffer ones, as they have more time/op (more transfers means more of this one tiny code change having any effect). But then also maybe confirmation bias.

I do think the most interesting thing here is the massive variance at the 32k buffer for both the read and write. Even though it’s a massive 2ms to the new one, the variance is so massive there’s no way to say for sure it’s because of this code change.

So, nothing spectacular, but I still think “nocode is best code” is still controlling anyways. Nothing massive, but still a better choice anyways. (More interesting, removing this opens up the possibility of moving the mutex later to just the writing, meaning the packet marshal times can potentially overlap. That might be where I saw the big boost that I spoke of above.)

puellanivis:
> If we want to change them we should do it in a way that does not require to checking if an interface is implemented for each packet (for example for writes)

:thinking: A good idea. Have all of them implement `marshalPacket` rather than `MarshalBinary` and then drop the test. Good idea. I will probably work it into my “rewrite all the unmarshal code”. Since I’m there futzing with all the packet marshal/unmarshal code already.

puellanivis:
With no objections, let’s merge this?

PR #422, opened 2021-03-20
Minor cleanup

greatroar:
Minor stuff that should not get in the way of major refactors.

### DIFF
--- a/request-errors.go
+++ b/request-errors.go
@@ -37,18 +37,18 @@ func (e fxerr) Error() string {
 	case ErrSSHFxEOF:
 		return "EOF"
 	case ErrSSHFxNoSuchFile:
-		return "No Such File"
+		return "no such file"
 	case ErrSSHFxPermissionDenied:
-		return "Permission Denied"
+		return "permission denied"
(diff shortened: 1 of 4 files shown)


sftp.go
```diff
@@ -58,7 +58,7 @@ const (
 	sshFxNoMedia                 = 13
 	sshFxNoSpaceOnFilesystem     = 14
 	sshFxQuotaExceeded           = 15
-	sshFxUnlnownPrincipal        = 16
+	sshFxUnknownPrincipal        = 16
 	sshFxLockConflict            = 17
 	sshFxDitNotEmpty             = 18
```
puellanivis, line 63:
```suggestion
	sshFxDirNotEmpty             = 18
```

request-errors.go
```diff
@@ -33,22 +33,22 @@ const (
 func (e fxerr) Error() string {
 	switch e {
 	case ErrSSHFxOk:
-		return "OK"
+		return "ok"
```
puellanivis, line 36: `OK` like `EOF` should be all caps.

greatroar: OK! :)

PR #423, opened 2021-03-21
Refactor/encoding ssh filexfer

puellanivis:
This makes an `encodings/ssh/filexfer` sub-package that defines the secsh-filexfer protocol encoding entirely independent of any implementation of functionality.

As this will be a public API, I would like merciless review of names, and design.

My biggest question: should we jump straight to `ReadDir` and `FStat` rather than `Readdir` and `Fstat`? Or should we try to overly strictly match the snakecase from the spec? I‘ve kind of taken the route of “ignore the spec for Go-friendly” for a number of fields already.


encoding/ssh/filexfer/buffer.go
```diff
@@ -0,0 +1,268 @@
...
+
+// Buffer wraps up the various encoding details of the SSH format.
+//
+// Data types are encoded as per section 4 from https://tools.ietf.org/html/draft-ietf-secsh-architecture-09#page-8
+type Buffer struct {
+	b   []byte
+	off int
+}
+
+// NewBuffer creates and initializes a new Buffer using buf as its initial contents.
```
greatroar, line 22: The argument is called b.

And the function doesn't just use it for initial contents. It takes ownership of b.

encoding/ssh/filexfer/buffer.go
```diff
@@ -0,0 +1,268 @@
...
+	ErrLongPacket  = errors.New("packet too long")
+)
+
+// Buffer wraps up the various encoding details of the SSH format.
+//
+// Data types are encoded as per section 4 from https://tools.ietf.org/html/draft-ietf-secsh-architecture-09#page-8
+type Buffer struct {
+	b   []byte
+	off int
+}
```
greatroar, line 20: Why not

    type Buffer bytes.Buffer

? That way, we get a host of convenience methods for free.

encoding/ssh/filexfer/buffer.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+
+	binary.BigEndian.PutUint32(b.b, uint32(size))
+}
+
+// MarshalBinary returns b as the binary encoding of b.
+func (b *Buffer) MarshalBinary() ([]byte, error) {
+	n := len(b.b)
+
+	// clamp cap() == len(), so any appends will reallocate.
```
greatroar, line 255: Appends will reallocate, but changes in the returned slice are still reflected in the Buffer.

encoding/ssh/filexfer/buffer.go
puellanivis: Because otherwise we have to read out values from the `bytes.Buffer` before we can use them. This makes the `ConsumeUintXY` functions more complicated.

encoding/ssh/filexfer/buffer.go
puellanivis: Line was taken from `bytes.NewBuffer` https://golang.org/pkg/bytes/#NewBuffer

greatroar:
Thanks again for working on this.

Have you considered giving the MarshalInto methods value receivers? Now, to marshal an extended attribute, we do

```go
attr := &ExtendedAttribute{Type: typ, Data: data}
attr.MarshalInto(buf)
```

which allocates the ExtendedAttribute on the heap if MarshalInto is not inlinable. With a value receiver, the whole packet goes on the stack, always, if we do

```go
ExtendedAttribute{Type: typ, Data: data}.MarshalInto(buf)
```

Given that Go manages the stack as a big dynamic array, this can be a big win.

encoding/ssh/filexfer/buffer.go
puellanivis: I would rather not allocate a buffer to return. It’s often sufficient. The `bytes.Buffer.Bytes()` function does the same. https://golang.org/src/bytes/buffer.go?s=2118:2149#L44

I did however, miss that this should be started from `b.b[b.off:]`. :laughing:

encoding/ssh/filexfer/buffer.go
greatroar: Which says in the next sentence that the Buffer takes ownership :)

encoding/ssh/filexfer/buffer.go
greatroar: ConsumeUint32 actually gets shorter:

```go
func ConsumeUint32(b *bytes.Buffer) (uint32, error) {
	if b.Len() < 4 {
		return 0, ErrShortPacket
	}

	p := b.Next(4)
	return binary.BigEndian.Uint32(p), nil
}
```

greatroar: The Append* methods may get more expensive, though. The inliner doesn't like bytes.Buffer.WriteByte.

encoding/ssh/filexfer/buffer.go
greatroar: If it starts at b.off, it's no longer MarshalBinary. It's Bytes.

encoding/ssh/filexfer/packets.go
```diff
@@ -0,0 +1,227 @@
...
+	b, err := readPacket(r, b)
+	if err != nil {
+		return err
+	}
+
+	return p.UnmarshalBinary(b)
+}
+
+// RequestPacket decodes a full request packet from the internal Data based on the Type.
+func (p *RawPacket) RequestPacket() (*RequestPacket, error) {
```
greatroar, line 146: Given that a RequestPacket is only three words, this method could return by value. That way, RequestPacket.Reset is no longer needed because it doesn't have to be pooled.

encoding/ssh/filexfer/packets.go
```diff
@@ -0,0 +1,227 @@
...
+		return err
+	}
+
+	p.Type = PacketType(typ)
+
+	if p.RequestID, err = buf.ConsumeUint32(); err != nil {
+		return err
+	}
+
+	p.Data = *buf
```
greatroar, line 97: This too takes ownership of the underlying buffer.

encoding/ssh/filexfer/buffer.go
puellanivis: Indeed. I’d also rather not have to deal with the possibly of someone Unreading anything either as well.

This gets the simplicity of everything we nee and only what we need.

encoding/ssh/filexfer/buffer.go
puellanivis: I mean, the two are functionally the same thing. They’re just named differently to match interfaces. Come to think of it though, “return b.Bytes()` is a good idea for the implementation as that was the intended version.

encoding/ssh/filexfer/packets.go
puellanivis: Just because you’re passing something by value does not mean that it does not also still get dragged into the heap. A lot of issues that I have been running into now are from things that _I_ can clearly see should be on the stack escaping to the heap because the compiler cannot be sure.

encoding/ssh/filexfer/packets.go
puellanivis: Like the other function, I’ll document this.

encoding/ssh/filexfer/buffer.go
puellanivis: 👍 I shall import more of the godoc. 👍

encoding/ssh/filexfer/buffer.go
```diff
@@ -0,0 +1,268 @@
...
+func (b *Buffer) ConsumeString() (string, error) {
+	v, err := b.ConsumeByteSlice()
+	if err != nil {
+		return "", err
+	}
+
+	return string(v), nil
+}
+
+// AppendString appends a single string of UTF-8 encoded text into the Buffer.
```
greatroar, line 236: Nothing in the code says UTF-8, nor should it. Go doesn't care what we put in our strings and the filexfer draft only specifies the encoding of error messages.

encoding/ssh/filexfer/buffer.go
```diff
@@ -0,0 +1,268 @@
...
+}
+
+// AppendUint8 appends a single byte into the Buffer.
+func (b *Buffer) AppendUint8(v uint8) {
+	b.b = append(b.b, v)
+}
+
+// ConsumeBool consumes a single byte from the Buffer, and returns true if that byte is non-zero.
+// If Buffer does not have enough data, it will return ErrShortPacket.
+func (b *Buffer) ConsumeBool() (bool, error) {
```
greatroar, line 86: This method is never used. I don't think filexfer uses booleans at all (?).

encoding/ssh/filexfer/filexfer.go
```diff
@@ -0,0 +1,12 @@
+package filexfer
+
+// Packet defines the behavior of an SFTP packet.
+type Packet interface {
+	MarshalPacket(reqid uint32) (header, payload []byte, err error)
+	UnmarshalPacketBody(buf *Buffer) error
+}
+
+// ComposePacket converts returns from MarshalPacket into an equivalent call to MarshalBinary.
+func ComposePacket(header, payload []byte, err error) ([]byte, error) {
```
greatroar, line 10: When would this function be used? It looks like it could be private, at least for now, but I haven't considered any server code.

greatroar:
Re:

> should we jump straight to ReadDir and FStat rather than Readdir and Fstat?

Fstat mirrors os.Lstat, so it's idiomatic. I think Read[Dd]ir should match whatever you do with Read[Ll]ink.

encoding/ssh/filexfer/packets.go
greatroar: Presumably, that happens only when you assign to an interface value (?). Those are pointers in disguise, even when you put a non-pointer value in them.

greatroar:
I think this code still mixes the concerns of allocation and the wire format. There are also several packet types that have multiple ways of marshaling and unmarshaling. If all code used Buffer as a central type and type+id were moved into it, I think the design could be cleaner and more efficient to use:

```go
type Buffer struct {
    Type      PacketType
    RequestID uint32
    // actual buffer stuff
}

func NewBuffer() *Buffer // For senders. Sets up "hidden" space for length+type+id.

func ReadPacket(io.Reader, *Buffer) error // For receivers. Parses off length+type+id.

func (b *Buffer) AppendUint32(uint32)
func (b *Buffer) ConsumeString() (string, error) // etc.

func (b *Buffer) Send(w io.Writer, payload []byte) error {
    b.AppendUint32(uint32(len(payload)))
     // actual Write calls
    b.Reset()
    return err
}
```

Then, marshaling code can be written without using interfaces, without NewMarshalBuffer and without size computation:

```go
type OpenPacket struct {
    Filename string
    PFlags   uint32
    Attrs    Attributes
}

// Could be a method too, but I don't really see the benefit.
func MarshalOpenPacket(b *Buffer, filename string, pflags uint32, attrs Attributes) {
    b.AppendString(filename)
    b.AppendUint32(pflags)
    MarshalAttributes(b, attrs)
}

func (p *OpenPacket) UnmarshalPacketBody(*Buffer) error // As right now.
```

A Client can now manage all of its memory using a single sync.Pool (global or field) of Buffers. Servers could presumably do the same.

greatroar:
As a side concern, none of the UnmarshalPacketBody functions that I've seen check whether they reach the end of the packet. This could be added to all of them, but maybe a function wrapper could do this?

```go
type Packet interface {
    MarshalPacket(reqid uint32) (header, payload []byte, err error)

    // UnmarshalPacketBody should not be called directly. Use the function with the same name instead.
    UnmarshalPacketBody(buf *Buffer) error
}

func UnmarshalPacketBody(buf *Buffer, p Packet) error {
    err := p.UnmarshalPacketBody(buf)
    if err == nil && buf.Len() != 0 {
        err = errors.New("sftp: trailing garbage after packet body")
    }
    return err
}

File: encoding/ssh/filexfer/packets.go (hunk 4 above)
Comment (puellanivis) [reply encoding/ssh/filexfer/packets.go:146]: Also, oddly, just because something is passed by pointer does not mean that the compiler has actually put it on the heap. Many might be surprised that `b := make([]byte, 4)` can allocated on the stack, or that a `r := &RequestPacket{}` can also be on the stack.

Meanwhile, even something like `fmt.Printf("constant format string: %d", i)` can have its arguments escape to heap. 🙃 It’s been a very _very_ annoying go around with this stuff, and confusing choices made by the compiler. I have generally found that any virtualized function call will force its receiver to escape to the heap even if its receiver is a value rather than pointer type, and the only way to prevent such a transfer to heap is if the whole chain can be inlined and the virtualized function can be devirtualized. At that point, it doesn’t matter if it is by value or receiver, it does not automatically escape to heap.

Also, any buffers that are subject to a `r.Read(…)` also escape to heap, even if we can patently see that it doesn’t really have to. But because the contents are handed off to a virtualized subfunction that could potentially retain the buffer, it has to be moved to the heap. 😭 

My point here is: switching types just in order to prevent escaping to heap is way more complicated than can be resolved just by switching from a pointer to value type. That said, I have looked at the potentials of having the `MarshalPacket` types be value receivers, as this seems to increase the likelihood of them devirtualizing… but I’m also pretty sure that I cannot prevent this, as we very likely will have to call some other virtualized method on the type, causing it to escape to heap even if it is constructed by value.

File: encoding/ssh/filexfer/buffer.go (hunk 6 above)
Comment (puellanivis) [reply encoding/ssh/filexfer/buffer.go:236]: The Go standard actually does say that the string type is “implicitly UTF-8 encoded”. Specifically, it is defined behavior that a `range` over a string will interpret that string as UTF-8 runes, and those runes will be iterated through, not the actual bytes. Therefore, I took the choice that anything intended to fully arbitrary data should be handled (as per Go Best Practice) a byte slice, while respecting that strings should be treated as very-likely-to-be-turned-into-UTF-8.

While the filexfer-02 standard does not explicitly say that most strings are UTF-8 encoded, starting from filexfer-03 many of the string fields _are_ defined to be UTF-8 only (unless otherwise negotiated), the only ones that stay as arbitrary byte encodings are on the FXP_DATA, and FXP_WRITE packets, as well as all of the handles.

I think I will however update the documentation by saying that due to how Go works, strings are necessarily always implicitly UTF-8, and “as per usual” care must be taken when using them for arbitrary data or other encodings.

File: encoding/ssh/filexfer/buffer.go (hunk 7 above)
Comment (puellanivis) [reply encoding/ssh/filexfer/buffer.go:86]: While `filexfer-02` does not use this type, it does appear in `filexfer-06`. While the protocol of this version of the library is at `filexfer-02`, if someone wanted to implement a later version of the protocol, I expect that most of these types and values will be directly imported in, rather than reimplementing everything. Thus, a few of these types have been written to support not-just-the-code-here, but any potential version of the `filexfer` format.

File: encoding/ssh/filexfer/filexfer.go (hunk 8 above)
Comment (puellanivis) [reply encoding/ssh/filexfer/filexfer.go:10]: The situation is for other versions of the library being implemented in a sort of “overlay” format. So `encoding/ssh/filexfer/v4` could make use of this function as well to quickly convert its `MarshalPacket` into a `MarshalBinary` for its own `SSH_FXP_NAME` type (which is binary incompatible with v3’s version), or like `encoding/ssh/filexfer/v6` with the `SSH_FXP_LINK` and `SSH_FXP_BLOCK` and `SSH_FXP_UNBLOCK`. 

This library is intended to be flexible enough to implement all versions of the `filexfer` protocol, not just strictly the v3. Though, v3 is the default and main protocol implementation.

**Comment by puellanivis** [comment]
Most of the `NewBuffer` usage in the library does not actually escape from the stack, and so they get allocated out of the stack. Putting them into a `Pool` is thus actually counter-productive. Though, I see where you’re coming from. They could then potentially save an allocated buffer, and then we could reuse that internal buffer. I see some merit there, where we could have the packets also implement `MarshalInto()`, but then there is confusion from things that are not packets, like `Attributes` and `ExtensionAttribute`. Though, that doesn’t mean we couldn’t use `MarshalPacketInto` and `MarshalPacketBodyFrom`…

But it turns out most of the packets types are called pretty infrequently, and do not have to be optimized so heavily that they reuse buffers. The overhead of the allocations from a `OpenPacket` is on the whole quite minimal. So, we really only need concern ourselves with `WritePacket`, `DataPacket` and `StatusPacket`, which compose the vast majority of communications. It turns out that even the allocation of the headers here is actually pretty trivial compared to the use of the payloads underneath them. That’s probably why we haven’t bothered “fixing” the `sendPacket` code so far. It’s such a minimal pressure on memory compared to the `recvPacket`.

That’s why I focused attention on `ReadFrom(io.Reader, []byte)` allowing a backing buffer input, to permit reuse. This is the primary memory hog/chokepoint right now. Running a benchmark comparison between an `InitPacket.MarshalInto` vs `InitPacket.MarshalBinary` the lack of allocating a buffer saves 1 alloc of 48 B, and runs in 17 ns instead of 71~81 ns. But now that I think about it, realistically, its going to have to pull that buffer out of a pool, right? Well, conveniently we already have the `bufPool` so making that change was easy. Now it saves 1 alloc of 48 B, and runs in ~64 ns instead of ~81 ns. I mean, obviously it’s going to have some gains there, but the question is if those gains are at all useful. We can chase optimizations forever, what we want to do is have data and profiling showing where real world bottlenecks are before just blinding making code harder to read, understand, and use.

Though, now that I’ve typed all of the above, and made a whole big argument about it. Rewriting the `MarshalPacket` calls to just take a `[]byte` to use as an optional backing buffer for the header actually isn’t actually all that bad of an idea, and I’m going to incorporate it.

**Comment by puellanivis** [comment]
I understand the concerns of checking to make sure that we’ve consumed the whole packet, but I’m not sure that’s altogether too necessary. There’s not really much to be gained by testing it, and if clients are sending spurious data that we would be ignoring anyways, I don’t see any reason to crash the server based on it.

File: encoding/ssh/filexfer/buffer.go (hunk 2 above)
Comment (puellanivis) [reply encoding/ssh/filexfer/buffer.go:20]: Oh, also, `type Buffer bytes.Buffer` is an opaque definition that does not import the functions of `bytes.Buffer`. In order to expose the functions of the underlying struct you have to encapsulate it: `type Buffer struct{ bytes.Buffer }`. But over-exposure of exported methods is also frustrating. Anyone could `b.Write(…)` whatever they wanted outside of spec encoding rules, and I’d really rather avoid misuse of the buffer.

**Comment by puellanivis** [comment]
I’ve taken a close look at making the `MarshalPacket` value receiver so they can potentially be devirtualized if the functions are inlined, and unfortunately, doing pretty much anything other than just `MarshalPacket` itself makes that function too complex to be inlined.

That includes even just getting a unique request id with `atomic.AddUint32(&count, 1)`.

So, it looks like in the Real World, that devirtualization is simply never going to happen, and I think the solution here is going to have to be the use of a pool of request packets for places where these allocations can happen in a tight loop, such as for the `SSH_FXP_WRITE`, `SSH_FXP_DATA` and `SSH_FXP_STATUS` packets.

As for the `ExtendedAttribute`, `ExtensionPair`, and `NameEntry` calls using values, my principle concern is in the performance of the `append(…)` functions that can programmatically build list of these structs. Their utility in isolation from their lists is pretty low. So, in this case, we’re almost always already going to already have a pointer to them. And as I mentioned, just because something is a pointer receiver does not mean it cannot be allocated on the stack.

In fact, looking at the simple example, the escape analysis shows it doesn’t escape, and is allocated on the stack, even though it also reports that `MarshalInto` it is too complex to inline. So, really, the only time it is going to escape onto the stack, is where `MarshalInto` is being run as a virtualized call through an interface, but we don’t actually do that with any of them.

**Comment by greatroar** [comment]
> Rewriting the MarshalPacket calls to just take a []byte to use as an optional backing buffer for the header actually isn’t actually all that bad of an idea, and I’m going to incorporate it.

Please, stop! Let's take a step back from performance and look at the complexity of the API. There are bytes slices with unclear ownership, packet types with MarshalBinary (but not fullfilling the interface promise), packet types with MarshalPacket, packet types with both, two types that can ReadFrom.

I have a further thought out alternative design that can drastically reduce the number of methods and types. Untested code up ahead.

```go
type Buffer struct {
	b   []byte
	off int
}

// NewBuffer constructs a buffer with the given id,
// reusing the space allocated for b if not nil.
func NewBuffer(b *Buffer, requestID uint32) *Buffer {
	// TODO version and init packets.

	const headerSize = 9 // length+type+id

	if b == nil {
		b = &Buffer{}
	} else {
		b.Reset()
	}
	b.reserve(headerSize)

	b.AppendUint32(0) // Reserve space for length.
	b.AppendByte(0)   // Reserve space for type.
	b.AppendUint32(requestID)

	return b
}

func (b *Buffer) SetType(t PacketType) { b.b[4] = byte(t) }
```

This replaces NewBuffer, NewMarshalBuffer. The Buffer type also replaces RawPacket and RequestPacket. It doesn't get Bytes, Len, MarshalBinary, PutLength or UnmarshalBinary.

NewBuffer doesn't even need to know about defaultMaxPacketSize if Buffers are properly reused (although reserve might want to allocate some minimum number of bytes, say 1kB).

```go
// ReadPacket reads a full packet from r into b.
//
// The Consume methods on b start reading after the request id.
func ReadPacket(r io.Reader, b *Buffer) error {
	// TODO version and init packets.

	const headerSize = 9 // length+type+id

	b.Reset()
	b.reserve(headerSize)

	b.b = b.b[:headerSize]
	_, err := io.ReadFull(r, b.b)
	if err != nil {
		return err
	}

	length := binary.BigEndian.Uint32(b.b)
	if length == 0 {
		return errShortPacket
	}

	b.reserve(4 + int(length))
	b.b = b.b[:4+length]
	b.off = headerSize

	n, err := io.ReadFull(r, b.b[b.off:])
	b.b = b.b[:headerSize+n]
	return err
}

func (b *Buffer) Type() PacketType { return PacketType(b.b[4]) }

func (b *Buffer) RequestID() uint32 { return binary.BigEndian.Uint32(b.b[5:]) }
```

You can rename this Buffer.ReadFrom if you want. In any case, it no longer needs the byte slice argument so there is less to document/think about.

UnmarshalPacketBody would work exactly as before.

```go
// Send writes the binary serialization of b to w, followed by the payload.
// It modifies b but does not retain it or the payload.
func (b *Buffer) Send(w io.Writer, payload []byte) error {
	if payload != nil {
		b.AppendUint32(uint32(len(payload)))
	}
	binary.BigEndian.PutUint32(b.b, uint32(len(b.b)+len(payload)-4))

	_, err := w.Write(b.b)
	if err != nil {
		return err
	}

	_, err = w.Write(payload)
	return err
}
```

Here's the only byte slice left, and it's read-only. You can rename this Buffer.WriteTo.

The predefined packet types now all get a method with signature

```go
// Send writes p to w.
// It uses the non-nil Buffer b to perform serialization.
// The Buffer should have the request id set appropriately.
func (p T) Send(w io.Writer, b *Buffer) error
```

E.g.,

```go
func (p AttrsPacket) Send(w io.Writer, b *Buffer) error {
	b.SetType(PacketTypeAttrs)
	p.Attrs.MarshalInto(b)
	return b.Send(w, b)
}

func (p DataPacket) Send(w io.Writer, b *Buffer) error {
	b.SetType(PacketTypeData)
	return b.Send(w, p.data)
}
```

These replace MarshalBinary, MarshalInto, MarshalPacket. If a packet needs to be marshaled in full, just Send it to a bytes.Buffer. If MarshalBinary is still needed for whatever reason, an adapter can be written for any type that has Send.

A Client would normally get a Buffer from a Pool, NewBuffer to set the id, Send it off, reuse it to receive the response and Put it back in the Pool.

A Server would similarly get a Buffer from a Pool, receive into it, reuse it for the response using NewBuffer, Send it off and Put it on the Pool.

Summarizing:

* RawPacket and RequestPacket are folded into Buffer.
* Byte slices are all packaged up in Buffer. No more worrying about ownership semantics or about slices in sync.Pool when optimizing.
* There is one marshaling method for each packet type. It's short and doesn't need to worry about memory management.
* A single Pool gets rid of nearly all the allocations, small or tiny, with no additional effort. In particular, no length calculations are needed. That means no hidden GC overhead that you don't see in a sequential benchmark.

puellanivis:
I‘ve changed most of the names to match up with `ReadDir` except:
* `Setstat` or `SetStat`? This one is pretty borderline.
* `Mkdir/Rmdir` or `MkDir/RmDir` or maybe even `MakeDir/RemoveDir`? Our own `Client` uses `Mkdir/RemoveDirectory` which is probably not the best and most consistent choice to copy.

`MarshalPacket` now takes an optional backing byte-slice that it can use for marshaling if it has a reasonably capacity (every packet uses `cap() < 9` to ignore the backing buffer and allocate a new one.)

The design is that there are _four_ types of packets:
* Init/Version packets that do not have a `RequestID`. These should not fulfill the `Packet` interface, as they are extraordinary packets. They implement `MarshalBinary` and `UnmarshalBInary` instead to easily interface with binary encoding.
* RawPacket/RequestPacket are special packets outside of the standard, and provided as convenience wrappers. They can `ReadFrom`, `MarshalBinary/UnmarshalBinary` because they are self-contained data structures that can render their full entirety without any input. They also implement `Packet` because it makes sense that they can.
* The `openssh/` packets provide an example of how an ExtendedPacket works. They implement `MarshalBinary/UnmarshalBinary` to encode their packet-specific data, and provide an implementation of `Packet` by wrapping a call to `ExtendedPacket`.
* The rest of the packets only need to encode their specific packet data. Passing a `reqid` into them avoids having to call an `ID()` virtual method on them, which makes them escape to the heap. They implement `Packet` as they are packets, but they do not implement `MarshalBinary/UnmarshalBinary` because they do not have a RequestID to allow them to be self-contained. An earlier version had them carry their own `RequestID` but there’s really no reason why the caller at the `clientConn.sendPacket` endpoint really need to care about what RequestID it has. That should be coördinated at the connection level. Servers can deal with a `RequestPacket` that will have the `RequestID` and the rest of the packet.

There is no getting around the fact that there are four different types of packets here. Short of removing the `RawPacket/RequestPacket` and leaving us only with three types. But the `RequestPacket` in particular allows us to easily implement a server, while the `RawPacket` allows us to simplify the client response receiver.

puellanivis:
Replacing the byte-slice argument with a `Buffer` argument is not saving anything at all. Neither complexity or arguments.

Not everything can be replaced with `Send(…)`, particularly almost all of the `MarshalInto(…)` which you seem to have recognized.

> No more worrying about ownership semantics or about slices in sync.Pool when optimizing.

We have already solved this problem with the `bufPool`, which works significantly better than even a `sync.Pool` of `bytes.Buffer`.

> A single Pool gets rid of nearly all the allocations, small or tiny, with no additional effort. In particular, no length calculations are needed. That means no hidden GC overhead that you don't see in a sequential benchmark.

That’s great for the complex case, where we’re pooling all of these buffers. But what about the simple cases? Where we don’t care if they’re allocating, because we’re not concerned about allocating a few bytes here or there. Now in order to use the library, we have to concern ourselves with all our `Buffer` lifetime and ownerships. My solution offers a two-tier approach that can even choose between on a case-by-case basis: either you use `nil` and get the naive allocation, or you use a backing buffer. No special data type to go into it, just any old byte-slice. And these byte-slices eliminate the allocation reuse that you’re already complaining about. But they do offer a one-allocation naive solution on the side.

I think your concerns are based around a lot of what the codebase looks like _right now_, but I’ve already done a big refactor that hasn’t been included in this PR. It gets rid of a lot of waste, and overhead on the actual caller code, in exchange for complexity here. Simplicity is complex, and presenting a simpler to use API is way more important to me than simplifying this implementation itself.

If we want to just write another transparent layer and force every client to write out their own marshalling code, or track their own buffer inputs, I don’t think that’s a good idea.

encoding/ssh/filexfer/packets.go
puellanivis: I don’t think I addressed this in the other comment, but no, not just because it’s assigned into an interface value. If a struct is large enough, keeping it in the stack even once becomes a liability, and Go will instead allocate that value on the heap. Even though it’s a value.

Tracking what goes on the stack and what goes on the heap is simply not as simple as “it’s secretly a pointer” or not. Plenty of values that are interfaces slices, or even pointer values can avoid allocation on the heap.

greatroar:
> Not everything can be replaced with Send(…), particularly almost all of the MarshalInto(…) which you seem to have recognized.

MarshalInto and Send do different things, though. One sends a packet when it's ready. The other builds a piece of a packet.

> But what about the simple cases? Where we don’t care if they’re allocating, because we’re not concerned about allocating a few bytes here or there. Now in order to use the library, we have to concern ourselves with all our Buffer lifetime and ownerships.

Hardly. Just get a buffer. Add local reuse or caching as necessary. Let's assume I renamed ReadPacket RecvPacket and gave it a *Buffer return value. First version:

```go
// This is ugly, but can be wrapped up in a utility function
// for callers that are willing to pay interface overhead.
err := filexfer.OpenDirPacket{Path: p}.Send(conn, filexfer.NewBuffer(nil, id))
// Handle err.

b, err := filexfer.RecvPacket(conn, nil)
// Handle err, b, maybe using an exported variant of newPacketFromType.

functionThatDoesReaddir()
```

Second version:

```go
b := filexfer.NewBuffer(nil, id)
err := filexfer.OpenDirPacket{Path: p}.Send(conn, b)
// Handle err.

b, err = filexfer.RecvPacket(conn, b)
// Handle err, b.

functionThatDoesReaddir(b)
```

Third version would cache Buffers somewhere.

> I think your concerns are based around a lot of what the codebase looks like right now

Probably. I'm curious to see the next version.

encoding/ssh/filexfer/buffer.go
greatroar: Agreed. I'd mark this resolved, but GitHub doesn't allow me to.

encoding/ssh/filexfer/fx.go
```diff
@@ -0,0 +1,137 @@
...
+)
+
+// Status defines the SFTP error codes used in SSH_FXP_STATUS response packets.
+type Status uint32
+
+// Defines the various SSH_FX_* values.
+const (
+	// see draft-ietf-secsh-filexfer-02
+	// https://tools.ietf.org/html/draft-ietf-secsh-filexfer-02#section-7
+	StatusOK = Status(iota)
```
greatroar, line 14: godoc doesn't recognize this form. It does recognize

```go
StatusOK Status = iota
```

and puts the constants below the type definition. (Not sure what pkg.go.dev does, but it might be the same.)

encoding/ssh/filexfer/fxp.go
```diff
@@ -0,0 +1,124 @@
...
+
+	// https://tools.ietf.org/html/draft-ietf-secsh-filexfer-08#section-3.3
+	PacketTypeV6Block
+	PacketTypeV6Unblock
+)
+
+// Response packet types.
+const (
+	// https://tools.ietf.org/html/draft-ietf-secsh-filexfer-02#section-3
+	PacketTypeStatus = PacketType(iota + 101)
```
greatroar, line 45: ... and here

encoding/ssh/filexfer/fxp.go
```diff
@@ -0,0 +1,124 @@
...
+	PacketTypeHandle
+	PacketTypeData
+	PacketTypeName
+	PacketTypeAttrs
+)
+
+// Extended packet types.
+const (
+	// https://tools.ietf.org/html/draft-ietf-secsh-filexfer-02#section-3
+	PacketTypeExtended = PacketType(iota + 200)
```
greatroar, line 55: ... and here

encoding/ssh/filexfer/packets.go
greatroar: I can't reproduce that with Go 1.16:

```go
package main

import "fmt"

func main() {
	var x foo
	f(x)
}

type foo struct {
	a [1 << 20]byte
}

func f(x foo) {
	switch x.a[0] {
	}
	fmt.Println(x.a[0])
}
```

According to -gcflags=-m, f cannot be inlined. In the generated assembly, I see a copy of 131072 quadwords onto the stack and an increment of the stack pointer by a megabyte. The Println call then copies one byte to the heap :)

When I change the size of foo.a to 1<<30, compilation fails:

```
./main.go:4:6: stack frame too large (>1GB): 1024 MB locals + 0 MB args
./main.go:14:6: stack frame too large (>1GB): 0 MB locals + 1024 MB args
```

greatroar: Same with return values:

```go
//go:noinline
func f() [1<<20]byte {}
```

copies a megabyte to the stack, but with 1<<30 it fails to compile.

puellanivis:
I’ve already spent a whole week working on simplifying the whole code base, only to find that this sort of library was important enough to implement now, rather than later. I had a specific goal in mind, and I weighing 15 different priorities all at the same time.

Now that I’ve wasted all this time and effort understanding the specs of filexfer-02 through to filexfer-13, and the finer points of all of them, and how I could best logically represent the relevant data-types in idiomatic Go in a way that was extensible and useful enough to expand for V4, V5, and V6 support rather than hyper focusing on making _this_ and only this version of the package work as quickly as possible.

I’m not going to spend yet another week rewriting the package for a whole different paradigm that drops 14 of the 15 goals I was trying to juggle in the first place.

This has been entirely demoralizing, and I’m out for at least a month. I’ll enjoy this vacation time spending my spare time playing video games rather than putting my 25 years of coding experience on thankless tasks that just get attacked for details that I’ve already considered and weighed, and found to be fundamentally lacking. I get enough people ignoring my expertise at work… at least there, they pay me to put up with it.

drakkan:
@puellanivis first of all I want to confirm that I really appreciate your work on this project and your great reviews. 

I am sorry for this and for not having had enough time to review the proposed changes (I have had some personal problems in the last few weeks). In any case I would have needed some time to try the code locally and play with it before proposing any change. 

In general I think we should first make the code work and then optimize it.

Enjoy this vacation, I hope you will continue to send your great contributions to this project in the future

greatroar:
@puellanivis I did not intend any of my comments as attacks on you or your work. I should at least have stopped arguing earlier.

Apologies, enjoy your vacation, and thank you for all the effort you put into this package. I will stay clear of this project if you want.

drakkan:
@puellanivis I don't want to lose your work, I took the liberty of restoring your [branch](https://github.com/pkg/sftp/tree/filexfer), we can resume from here in the future, thanks

puellanivis:
One really should not write things while actively upset over an issue. I’ve had some time to cool off, and want to follow up on some things.

@greatroar It’s not necessary for you to “stay clear of this project”, but it was just very frustrating. I took many of your recommendations to heart, despite even at first arguing against them. I also incorporated some of them as best as it would fit in considering the many competing interests involved. But some requests were either asking me to conflate two responsibilities/purposes into a single item, or were criticisms about the integration of the very idea suggested. I’ve had time to think over the criticisms despite being so upset, and will likely address a few of them still. (i.e. making sure interface constraints are respected, even if that means cloning backing byte-slices, etc.)

I usually do not push a first, second, or even third draft out for review. There were a lot of different considerations that were made during these multiple drafts, and issues that arose while I was working on refactoring the main package. I can see how there might be a disconnect between seeing what is being designed here, and the integration into the package as a whole. But I’ve been trying to address far more than just “is a superficially hyper-efficient way of turning a bunch of uint32/uint64/string into a stream of bytes.” In fact, very little of the involved code is actually intended to be used by callers, and far more intended to be used by implementers of extensions/newer-versions. Now that I think about it, I could potentially split things even further to make how and where the responsibilities are different.

PR #425, opened 2021-03-29
[Bugfix] WriteTo write work channel should never be closed…

puellanivis:
An earlier draft of this had the channel close signal a valid EOF, but since, the code has changed so that this closed channel read should never happen. As such, if it does, we should be returning an `err != nil` not an `EOF`.


puellanivis:
Well, this didn’t solve the individual’s issue, it’s also still a bit of a corner-case bug. So, I will merge it, but we won’t push any new version just for this. It should be rarely hit anyways.

PR #429, opened 2021-04-13
Guard against accidental io.EOFs

puellanivis:
There are a number of situations where we should really be able to _know_ that the code path doesn’t contain an `io.EOF`. As a precaution, we put a guard statement in each of these paths that should prevent an unintentional `io.EOF` from falling through.

Instead, we return `ErrSSHFxConnectionLost` instead. I’m not sure this is precisely the correct return value we want to use instead of an `io.EOF`, but I cannot think of what other kind of an error we should return.

### DIFF
--- a/client.go
+++ b/client.go
@@ -944,7 +944,7 @@ func (f *File) Read(b []byte) (int, error) {
 // readChunkAt attempts to read the whole entire length of the buffer from the file starting at the offset.
 // It will continue progressively reading into the buffer until it fills the whole buffer, or an error occurs.
 func (f *File) readChunkAt(ch chan result, b []byte, off int64) (n int, err error) {
-	for err == nil && n < len(b) {
+	for n < len(b) {
 		id := f.c.nextID()
 		typ, data, err := f.c.sendPacket(ch, &sshFxpReadPacket{
 			ID:     id,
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -944,7 +944,7 @@ func (f *File) Read(b []byte) (int, error) {
 // readChunkAt attempts to read the whole entire length of the buffer from the file starting at the offset.
 // It will continue progressively reading into the buffer until it fills the whole buffer, or an error occurs.
 func (f *File) readChunkAt(ch chan result, b []byte, off int64) (n int, err error) {
-	for err == nil && n < len(b) {
+	for n < len(b) {
```
puellanivis, line 947: This `err == nil` was always true, because line 949 shadowed the variable. Since every error path already `return`s in the loop anyways, the check was entirely extraneous.

client.go
```diff
@@ -974,7 +980,7 @@ func (f *File) readChunkAt(ch chan result, b []byte, off int64) (n int, err erro
 		}
 	}
 
-	return
+	return len(b), nil
```
puellanivis, line 983: We make the return here more precise and narrow. Since we know that reaching here these are the two values these return values logically _must_ have, we say so explicitly.

puellanivis:
Integration test failure appeared to have been caused by a race condition between two StatVFS commands returning slightly different values.

puellanivis:
This not only didn’t solve the issue we were facing, but it’s also obsoleted by the other refact, so I’m going to close this PR.

Issue #424, opened 2021-03-29
Incomplete downloads


mrwonko:
Since upgrading to 1.13 the library will often report EOF before a file opened with `Client.Open()` is downloaded completely. For example a 634618 byte file has stopped downloading after just 65536 bytes, and a 2367599 byte file after 2031616.

As a workaround I've added a call to `File.Stat().Size()`, which returns the correct file size, and compare that with the downloaded size to check for success.

puellanivis:
Queries: Is it returning an `io.EOF` error, or a nil error? How are you transferring the file? `Client.Open()` then a `io.Copy`?

mrwonko:
Yes, `Client.Open()` followed by `io.Copy()`, which returns a nil error, implying that the reader itself returned `io.EOF`.

puellanivis:
Hm… this is proving pretty hard to pin down what might be going wrong with `WriteTo`. I see one path that could result in an `err == nil` case without an `io.EOF` but no one closes any of those channels anyways, so it seems unlikely to be that, and I don’t really see how we could be getting an `io.EOF` from anything other than an `SSH_FXP_STATUS`…

Clearly some number of following bytes are not being committed, this can be clearly seen, because the two mistaken file size examples are at 32768 byte boundaries…

mrwonko:
>  the two mistaken file size examples are at 32768 byte boundaries

Sounds right, other examples like 1343488, 1638400 and 786432 also fulfil that.

puellanivis:
I’m thinking we might want to try a cherry pick that covers the unexpected channel close case. I’m entirely unsure how that could ever trigger, but it _might_ be the problem, as that is the only other way we can get a `nil` error without an EOF, and like, it very clearly shouldn’t be a `nil` error… it was just like that as a hold-over from an earlier draft.

This could also explain the observed behavior, some number of 32 KiB writes work, then we receive on a closed channel, and stop all transfers.

drakkan:
@puellanivis thank you. So if I want try to replicate this issue it should happen only for unexpected transfer errors, right?

puellanivis:
@drakkan That’s the thing, this really shouldn’t happen for unexpected transfer errors either. Those should all percolate some non-EOF error up the line as well? 🤔

drakkan:
I tryed to replicate this issue with no luck using something like this:

```
  for i := 0; i < 1000; i++ {
		start := time.Now()
		f, err := sftpClient.Open("file.zip")
		if err != nil {
[...]

Issue #426, opened 2021-04-03
Performance regression from v1.12.0 to v1.13.0 (and master)


ncw:
While investigating https://github.com/rclone/rclone/issues/5197 I discovered that doing transfers to rsync.net had gone from 300 KB/s in v1.12.0 to 100KB/s in v1.13.0

The rsync.net servers run FreeBSD which is one unusual thing about them and the other is that they are a long way away from me (150ms) so have the usual problems with long fat TCP pipes. Rclone uses the `ReadFrom` interface in the sftp client so that the sftp library can increase the number of outstanding packets to help with this.

Here are some tests (my upload is capable of 2MB/s).

    make && rclone version && rclone copy -vv --stats 10s /tmp/499.91M rsyncnet: 2>&1 | grep 499.91M:

v1.12.0

```
 *                                       499.91M:  0% /499.910M, 312.432k/s, 27m8s
 *                                       499.91M:  1% /499.910M, 309.264k/s, 27m15s
 *                                       499.91M:  1% /499.910M, 281.361k/s, 29m49s
 *                                       499.91M:  2% /499.910M, 312.820k/s, 26m37s
 *                                       499.91M:  3% /499.910M, 347.230k/s, 23m48s
```

v1.13.0

```
 *                                       499.91M:  0% /499.910M, 81.331k/s, 1h44m44s
 *                                       499.91M:  0% /499.910M, 97.554k/s, 1h27m8s
 *                                       499.91M:  0% /499.910M, 98.655k/s, 1h25m59s
 *                                       499.91M:  0% /499.910M, 122.371k/s, 1h9m7s
 *                                       499.91M:  1% /499.910M, 122.173k/s, 1h9m4s
```

master

```
 *                                       499.91M:  0% /499.910M, 109.761k/s, 1h17m33s
 *                                       499.91M:  0% /499.910M, 102.153k/s, 1h23m11s
 *                                       499.91M:  0% /499.910M, 91.254k/s, 1h32m58s
 *                                       499.91M:  0% /499.910M, 85.305k/s, 1h39m18s
 *                                       499.91M:  0% /499.910M, 76.641k/s, 1h50m23s
```

I bisected this change and discovered this commit is probably the problem. I'm reasonably sure that is the problem commit, and it is certainly touching the code in question. Before this commit at 0be6950c0e91d5cb73a4d690270d3a5010ac9808 the performance is definitely OK and after it is is bad. However there is some variation after the commit so there may be more commits involved.

[...]

PR #430, opened 2021-04-16
Encoding SSH filexfer

puellanivis:
Let’s try this as round two.

Goals of this sub-package:
* provide a common and safe wire encoding/decoding abstraction
* provide a common packet-level abstraction (packets are the most natural abstraction layer to surface here)
* the wire encoding/decoding abstraction should provide a easy and naïve use-case API (i.e. optimizing reallocations should not be necessary to even use the package)
* packet abstractions should be cleanly separated from the wire encoding/decoding
* `filexfer-02` should be the default implementation (but not the only possible implementation)
* the `openssh` extensions that we currently use should be implemented, in part as an example of how to write extensions
* optimizing reallocations should be possible, but as noted above: way more importantly, it should not be required
* when possible, input reusable buffers should be simple go types (i.e. `[]byte`)

Non-goals:
* thoroughly optimizing the `filexfer-02` use case to the detriment of flexibility. (The reason for this subpackage is to build flexibility, not limit it.)
* optimizations of avoiding reallocations should not be handled in package, but should be the responsibility of the caller.
* avoiding nuances and gotchas of any particular type or strategy of reallocations are not this subpackage’s responsibility (e.g. `[]byte`s in `sync.Pool`s; the mechanics of why these issues exist could be fixed or resolved at any particular time. We do not want to be chasing the bleeding edge here.)


drakkan:
Hi,

looks good to me, I propose to merge it without promising backward compatibility for now, so when we start using this package we will be free to make any changes we need. What do you think about it?

puellanivis:
Seems reasonable, “this is a test run, and we might need to make changes as we convert our main package to use it” sort of thing.

We could maybe stuff it into an `internal/` subdirectory until we’re ready to promise backwards compatibility?

drakkan:
> Seems reasonable, “this is a test run, and we might need to make changes as we convert our main package to use it” sort of thing.
> 
> We could maybe stuff it into an `internal/` subdirectory until we’re ready to promise backwards compatibility?

internal is a good idea

drakkan:
Thank you

puellanivis:
Now begins the fun part… 🙈 rewriting nearly everything.

drakkan:
Maybe we should create a new branch for bugfixes, for example we could release 1.13.1 once #424 and #426 are fixed

puellanivis:
Sounds good. Although, I actually started this PR based on a bunch of work I had essentially already done rewriting a lot of the client stuff.

Though, I’m still a bit at a loss for how to rresolve #424 . The PRs that I opened are way more defensive than anything else, and the changes I already started on would require both to be reworked anyways… but also since it seems that neither of the PRs actually solve the issue I’m absolutely stumped about where to go from there… and facing a big rewrite as well, I’m not sure anything we do now to fix the issue would matter much post-rewrite. 🤔 

But if we have a lead on how to fix #424 before a rewrite, I’m all ears, but it’s just baffling honestly.

drakkan:
I tried to replicate # 424 several times with no luck. I have tried several things but I am unable to reproduce, it must be something really weird. If it gets fixed after rewriting, this will be fine too

Issue #433, opened 2021-04-22
UseConcurrentWrites copies 0 bytes on windows when using io.Copy


lespea:
So this is a strange one and I'm not sure what to do next to help debug... but here's what I've found so far:

* On Mac & Linux setting up the ssh connection and copying the bytes from a file handle using `io.Copy` copies the entire file without error
* On Windows, using `io.Copy` against
    * the file handle copies 0 bytes
    * a buffered reader copies 0 bytes
    * a bytes.Reader copies the bytes correctly
 * Directly calling `sshWriter.Write(bytes)` copies everything just fine

Disabling `useConcurrentWrites` and all methods work as expected.  No errors are ever returned during any copy.

Tested using 32bit go 1.16.3 and go 1.15.8, and 64 bit 1.16.3

drakkan:
Hi,

can you please share a small program to reproduce the issue? thank you

lespea:
Okay here is a sample program that demonstrates the issue.  I tried for a while to get a server setup but ran into issues so you'll just have to fill out the Dest/User (optional KeyPass) with values to a real ssh server that the key has access to.

```go
package main

import (
	"bufio"
	"github.com/pkg/sftp"
	"golang.org/x/crypto/ssh"
	"io"
	"io/ioutil"
	"log"
	"net"
	"os"
)

const (
	Dest     = ""
	User     = ""
	KeyPass  = ""
	KeyPath  = "testKey"
	TestFile = "main.go"
	NoPanics = true
)

type ReadType uint8

const (
	CopyFile ReadType = iota
	CopyBuf
	RawWrite
)

func (rt ReadType) String() string {
	switch rt {
	case CopyFile:
		return "CopyFile"
	case CopyBuf:
		return "CopyBuf"
	case RawWrite:
		return "RawWrite"
	default:
		log.Panicf("Invalid ReadType: %d", rt)
	}
	return ""
}

func cez64(n int64, err error) {
	ce(err)
	if n <= 0 {
		log.Panic("got 0 bytes but expected > 0")
	}
}

func cez(n int, err error) {
	ce(err)
	if n <= 0 {
		log.Panic("got 0 bytes but expected > 0")
	}
}

func ce(err error) {
	if err != nil {
		panic(err)
	}
}

func cef(f func() error) {
	ce(f())
}

func getConf() *ssh.ClientConfig {
	keyBytes, err := ioutil.ReadFile(KeyPath)
	ce(err)

	var signer ssh.Signer
[...]

PR #435, opened 2021-04-22
bugfix: int64 math overflows possible on 32-bit architectures

puellanivis:
Investigating #433 I tracked it down to a situation where an overflow in 32-bit architectures can result in a calculation of `concurrency = 0`. With zero concurrency… the `wg.Wait()` returns immediately and no work is done.

The solution is to check if concurrency is less than one, and if it is, then use max concurrency, as we overflowed an int32, so it should use max concurrency.

### DIFF
--- a/client.go
+++ b/client.go
@@ -1058,7 +1058,7 @@ func (f *File) ReadAt(b []byte, off int64) (int, error) {
 	errCh := make(chan rErr)
 
 	concurrency := len(b)/f.c.maxPacket + 1
-	if concurrency > f.c.maxConcurrentRequests {
+	if concurrency > f.c.maxConcurrentRequests || concurrency < 1 {
 		concurrency = f.c.maxConcurrentRequests
 	}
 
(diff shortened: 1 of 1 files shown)


puellanivis:
https://play.golang.org/p/7y0QNA-vLNO

puellanivis:
:thinking: Actually, I’m going to add a bit more of a thing.

client.go
```diff
@@ -1181,7 +1181,7 @@ func (f *File) WriteTo(w io.Writer) (written int64, err error) {
 	}
 
 	concurrency := int(fileSize/uint64(f.c.maxPacket) + 1) // a bad guess, but better than no guess
```
puellanivis, line 1183: Let’s actually use a math expression that cannot overflow, and then clamp to `f.c.maxPacket` first on that.

client.go
```diff
@@ -1650,6 +1652,10 @@ func (f *File) ReadFrom(r io.Reader) (int64, error) {
 			}
 		}
 
+		if remain < 0 {
+			remain = math.MaxInt64
+		}
```
puellanivis, line 1657: This was added from discussion in #426 where we want to default to maximum concurrency when we have a `Size` reported as less than 0.

drakkan:
Thank you

PR #437, opened 2021-04-27
request server: add an optional RealpathFileLister interface

drakkan:
This allow to customize the responses for SSH_FXP_REALPATH requests
and so implementing features like a start directory.

I'm not sure if this is the best way to add this feature, maybe we could consider a server option too

### DIFF
--- a/request-example.go
+++ b/request-example.go
@@ -469,7 +469,7 @@ func (fs *root) Realpath(p string) string {
 	if fs.startDirectory == "" || fs.startDirectory == "/" {
 		return cleanPath(p)
 	}
-	return cleanPathWithBase(p, fs.startDirectory)
+	return cleanPathWithBase(fs.startDirectory, p)
 }
 
 // In memory file-system-y thing that the Hanlders live on
--- a/request-example.go
+++ b/request-example.go
@@ -464,10 +464,19 @@ func (fs *root) Lstat(r *Request) (ListerAt, error) {
 	return listerat{file}, nil
 }
 
+// implements RealpathFileLister interface
+func (fs *root) Realpath(p string) string {
+	if fs.startDirectory == "" || fs.startDirectory == "/" {
+		return cleanPath(p)
+	}
+	return cleanPathWithBase(p, fs.startDirectory)
+}
+
 // In memory file-system-y thing that the Hanlders live on
 type root struct {
-	rootFile *memFile
-	mockErr  error
+	rootFile       *memFile
+	mockErr        error
+	startDirectory string
 
 	mu    sync.Mutex
 	files map[string]*memFile
--- a/request-interfaces.go
+++ b/request-interfaces.go
@@ -86,13 +86,13 @@ type LstatFileLister interface {
 	Lstat(*Request) (ListerAt, error)
 }
 
-// RealpathFileLister is a FileLister that implements the Realpath method.
+// RealPathFileLister is a FileLister that implements the Realpath method.
 // We use "/" as start directory for relative paths, implementing this
 // interface you can customize the start directory.
 // You have to return an absolute POSIX path.
-type RealpathFileLister interface {
+type RealPathFileLister interface {
 	FileLister
-	Realpath(string) string
+	RealPath(string) string
 }
 
 // ListerAt does for file lists what io.ReaderAt does for files.
--- a/request-interfaces.go
+++ b/request-interfaces.go
@@ -86,6 +86,15 @@ type LstatFileLister interface {
 	Lstat(*Request) (ListerAt, error)
 }
 
+// RealpathFileLister is a FileLister that implements the Realpath method.
+// We use "/" as start directory for relative paths, implementing this
+// interface you can customize the start directory.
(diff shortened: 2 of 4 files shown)


puellanivis:
I’m kind of hesitant to add yet more behaviors exposed through these interfaces… I’m not sure what a better solution would look like, but I feel like this path is unsustainable and overly complicated.

request-interfaces.go
```diff
@@ -86,6 +86,15 @@ type LstatFileLister interface {
 	Lstat(*Request) (ListerAt, error)
 }
 
+// RealpathFileLister is a FileLister that implements the Realpath method.
+// We use "/" as start directory for relative paths, implementing this
+// interface you can customize the start directory.
+// You have to return an absolute POSIX path.
+type RealpathFileLister interface {
+	FileLister
+	Realpath(string) string
```
puellanivis, line 95: In the client we’re using `RealPath`, we might want to keep the two consistent between each other.

request-server.go
```diff
@@ -279,9 +284,13 @@ func cleanPacketPath(pkt *sshFxpRealpathPacket) responsePacket {
 
 // Makes sure we have a clean POSIX (/) absolute path to work with
 func cleanPath(p string) string {
+	return cleanPathWithBase(p, "/")
+}
+
+func cleanPathWithBase(p, base string) string {
```
puellanivis, line 290: The ordering of these arguments seems odd. We call `path.Join(base, p)`, and base is a prefix, but in this argument list it comes second?

request-server_test.go
```diff
@@ -807,6 +808,25 @@ func TestUncleanDisconnect(t *testing.T) {
...
 
+func TestRealPath(t *testing.T) {
+	root := &root{
+		rootFile:       &memFile{name: "/", modtime: time.Now(), isdir: true},
+		files:          make(map[string]*memFile),
+		startDirectory: "/apath",
+	}
+
+	p := root.Realpath(".")
+	require.Equal(t, root.startDirectory, p)
```
puellanivis, line 819: In the cases where these tests don’t succeed it’s not like anything is going to explode or panic if we continue execution, so `require` might not be entirely necessary.

request-example.go
```diff
@@ -464,10 +464,19 @@ func (fs *root) Lstat(r *Request) (ListerAt, error) {
 	return listerat{file}, nil
 }
 
+// implements RealpathFileLister interface
+func (fs *root) Realpath(p string) string {
+	if fs.startDirectory == "" || fs.startDirectory == "/" {
+		return cleanPath(p)
+	}
+	return cleanPathWithBase(p, fs.startDirectory)
```
puellanivis, line 472: It’s kind of weird that the only usage of this function occurs a file separate from its definition.

Usually, I prefer to see functions declared as close as possible to their usage, so that spacial locality is kept tight. But I do see `cleanPath` is used in a few different spots.

drakkan:
Thank you for the review, I tryed to fix the code based on your comments. Other libraries that I use within SFTPGo have a similar structure, for example:

- https://pkg.go.dev/golang.org/x/net/webdav#FileSystem
- https://pkg.go.dev/github.com/fclairamb/ftpserverlib#ClientDriver

In this case we could also add a server option that allows to set the start directory and use these values instead of `/` in `cleanPath`. Do you have better ideas?

puellanivis:
Yeah, following up on my concern about the `ReadPathLister` this is direction that we’re wanting to go towards: extensible interfaces. I just had a knee-jerk reaction to it. 😆

drakkan:
Thank you

drakkan:
Hi @puellanivis,

I'm trying to add a start directory feature to SFTPGo, see https://github.com/drakkan/sftpgo/issues/705.

I implemented the `RealPathLister` interface and with smart clients like `sftp` CLI or FileZilla this is enough. These clients request the current directory when they connect and then send absolute paths. 

I started to write the test cases, take a look here:

```
currentDir, err := client.Getwd()
assert.NoError(t, err)
assert.Equal(t, startDir, currentDir)

entries, err := client.ReadDir(".")
```

`client.Getwd` and `client.RealPath` are ok, but if I do `client.ReadDir(".")` I get the `/` contents: our client sends a relative path and it is converted server side as absolute in our `cleanPath` method. I think this happen only for the request server.

I think it would be better to avoid path normalization server side and delegate them to the request server implementation but it would be a backward incompatible change at this point. 

Add a check for relative path and call the `RealPathLister` could do the trick but it seems quite ugly and I don't want to add something like this.
Do you have better suggestions? Thanks

puellanivis:
🤔 Hm. It seems the standard never really required all paths to be absolute. Although, it specifies only that relative paths be from a rooted point (the user’s home directory), and provides no way to change that directory once set. The fact that the relative directory support does not have any way to traverse directories is precisely the reason why requiring all paths be absolute has worked for so long, client have to do that relative path tracking almost entirely themselves already, so why not also do it for the user directory as well.

I think is a case where we’re slightly out of spec, and should be doing different behavior, but also… the assumptions made lead to a design that is going to be difficult to “fix” now.

PR #439, opened 2021-05-22
Export a ReadFromWithConcurrency function that permits ensuring concurrency usage.

puellanivis:
Surfaces the `readFromConcurrent` into a `ReadFromWithConcurrency` alternative that allows the caller to be able to ensure that concurrency will be used.

### DIFF
--- a/client.go
+++ b/client.go
@@ -1500,8 +1500,13 @@ func (f *File) WriteAt(b []byte, off int64) (written int, err error) {
 	return len(b), nil
 }
 
-// readFromConcurrent implements ReaderFrom, but works concurrently rather than sequentially.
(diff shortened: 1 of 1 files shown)


puellanivis:
🤦‍♀️ forgot to run a full set of tests prior to pushing the PR. 😆 No one is perfect.

puellanivis:
```
sftp$ GOARCH=386 make integration
```

This one fails, but it only fails because there is no UID/GID matching outside of `CGO_ENABLED=1`, this is known from the client refactor to use `filexfer` and should be resolved after that.

drakkan:
LGTM, thank you

PR #445, opened 2021-06-29
[bugfix] Windows paths don’t work with Server

puellanivis:
Change uses of direct paths in `Server` to use `toLocalPath` to deconvert encoded absolute paths.

For Windows:
* We convert to slashes, then if it does not start with a slash, we add one at the start. So, when decoding paths, we have to check if after `FromSlash` removing the leading backslashes produces an absolute path. If it does, then we return that instead.
* We switch to using `filepath.Clean` first before `filepath.ToSlash` so that clients work consistently. ( `\\UNC\Path` paths work on clients that do not `FXP_REALPATH` sanitize first, while for those that do, `path.Clean` would otherwise strip out the double slash start.)

For Plan9:
* Some absolute paths (those that start with `#`) get prefixed with a slash in `cleanPath`, using `filepath.Clean` first should avoid accidentally dropping the `#`, and then we make sure a leading `/` will get removed if added from the “add a starting slash” encoding .
* Plan9 wasn’t compiling because `s_ISVTX` was typoed as `S_ISVTX`

All other OSes:
* `toLocalPath` just returns the path. There’s nothing further to be done for it.

### DIFF
--- a/request-server.go
+++ b/request-server.go
@@ -288,9 +288,9 @@ func cleanPath(p string) string {
 }
 
 func cleanPathWithBase(base, p string) string {
-	p = filepath.ToSlash(p)
+	p = filepath.ToSlash(filepath.Clean(p))
 	if !path.IsAbs(p) {
 		return path.Join(base, p)
 	}
-	return path.Clean(p)
+	return p
 }
--- a/request-plan9.go
+++ b/request-plan9.go
@@ -2,7 +2,11 @@
 
 package sftp
 
-import "syscall"
+import (
+	"path"
+	"path/filepath"
+	"syscall"
+)
 
 func fakeFileInfoSys() interface{} {
 	return &syscall.Dir{}
@@ -10,4 +14,21 @@ func fakeFileInfoSys() interface{} {
 
 func testOsSys(sys interface{}) error {
 	return nil
+}
+
+func toLocalPath(p string) string {
+	lp := filepath.FromSlash(p)
+
+	if path.IsAbs(p) {
+		tmp := lp[1:]
+
+		if filepath.IsAbs(tmp) {
+			// If the FromSlash without any starting slashes is absolute,
+			// then we have a filepath encoded with a prefix '/'.
+			// e.g. "/#s/boot" to "#s/boot"
+			return tmp
+		}
+	}
+
+	return lp
 }
--- a/request-unix.go
+++ b/request-unix.go
@@ -21,3 +21,7 @@ func testOsSys(sys interface{}) error {
 	}
 	return nil
 }
+
+func toLocalPath(p string) string {
+	return p
+}
--- a/request_windows.go
+++ b/request_windows.go
@@ -1,6 +1,10 @@
 package sftp
 
-import "syscall"
+import (
+	"path"
+	"path/filepath"
+	"syscall"
+)
 
 func fakeFileInfoSys() interface{} {
 	return syscall.Win32FileAttributeData{}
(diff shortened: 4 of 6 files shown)


request-server.go
```diff
@@ -288,9 +288,9 @@ func cleanPath(p string) string {
 }
 
 func cleanPathWithBase(base, p string) string {
-	p = filepath.ToSlash(p)
+	p = filepath.ToSlash(filepath.Clean(p))
```
puellanivis, line 291: The order here really doesn’t matter except for Windows, and only for UNC paths. “Let’s just not support UNC paths“ doesn’t work, because OpenSSH’s client does not sanitize the path first, so UNC paths would still work with that client anyways. :sob:

puellanivis:
```
linux-machine $ sftp -P 31337 windows-machine
sftp> ls //wsl$/openSUSE-Leap-15.3
//wsl$/openSUSE-Leap-15.3/bin           //wsl$/openSUSE-Leap-15.3/dev           //wsl$/openSUSE-Leap-15.3/etc           
//wsl$/openSUSE-Leap-15.3/home          //wsl$/openSUSE-Leap-15.3/init          //wsl$/openSUSE-Leap-15.3/lib           
//wsl$/openSUSE-Leap-15.3/lib64         //wsl$/openSUSE-Leap-15.3/lost+found    //wsl$/openSUSE-Leap-15.3/mnt           
//wsl$/openSUSE-Leap-15.3/opt           //wsl$/openSUSE-Leap-15.3/proc          //wsl$/openSUSE-Leap-15.3/root          
//wsl$/openSUSE-Leap-15.3/run           //wsl$/openSUSE-Leap-15.3/sbin          //wsl$/openSUSE-Leap-15.3/selinux       
//wsl$/openSUSE-Leap-15.3/srv           //wsl$/openSUSE-Leap-15.3/sys           //wsl$/openSUSE-Leap-15.3/tmp           
//wsl$/openSUSE-Leap-15.3/usr           //wsl$/openSUSE-Leap-15.3/var
```
:woman_shrugging: alright.

request-server.go
```diff
@@ -288,9 +288,9 @@ func cleanPath(p string) string {
...
 	if !path.IsAbs(p) {
 		return path.Join(base, p)
 	}
-	return path.Clean(p)
+	return p
```
drakkan, line 295: I did some basic tests and all worked, anyway since `path.Join` calls Clean internally, I'll leave path.Clean(p) here too

puellanivis: Hm… the point here in removing this `path.Clean()` is that PuTTY sftp sanitizes names with `FXP_REALPATH` and this removes the double slash at the start of a `//UNC/Path` this means that while on my Linux machine, I could do `ls //wsl$/openSUSE-Leap-15.3` attempting to do the same on my PuTTY sftp client, resulted in the path being cleaned in the `FXP_REALPATH`, and then failing with `cannot open: /wsl$/openSUSE-Leap-15.3`

drakkan: So we could have the same issue if the path is not absolute and we hit `return path.Join(base, p)`

puellanivis: True, but this should only apply in the case where the base being added is a UNC address. So, for example `cleanPathWithBase("//wsl$/openSUSE-Leap-15.3", aPath)`

I think in this case, we’re kind of a bit stuck on “having a non-default base paths is not particularly robust and a little bit of a hack”. 🤔 For this particular case, I’m a bit more willing to let the inconsistent behavior slide.

drakkan: Allowing non cleaned paths could break some usages, let's merge this patch and see if someone complaints. I'll try to do some tests myself using wsl (I never tested it) as soon as I have some free time, it could take a while

drakkan (approved):

request-server.go
puellanivis: Right, this is why I think keeping the `path.Join` which will auto-clean as well is the better choice for now, even though it would break UNC base paths. Getting some sort of logic to correctly handle all the nuances of properly rooting all reasonable base/path combos would be 😩 .

It’s at least pretty nice that we can do UNC paths with a “local” datastore with WSL, if I had to setup a remote SMB server to test any of this, I might very well tear my hair out. 😆

APinfinite (commented):
request-plan9.go
request-server.go

APinfinite (commented):
stat_plan9.go

PR #449, opened 2021-07-01
bugfix: there may be a nil pointer dereference

alanskind:
I wonder if it was designed to be that way or if this is a bug. Thanks a lot!

### DIFF
--- a/pool.go
+++ b/pool.go
@@ -17,7 +17,7 @@ func newBufPool(depth, bufLen int) *bufPool {
 func (p *bufPool) Get() []byte {
 	if p == nil {
 		// functional default: no reuse.
-		return make([]byte, p.blen)
+		return nil
 	}
 
 	for {


pool.go
```diff
@@ -17,7 +17,7 @@ func newBufPool(depth, bufLen int) *bufPool {
 func (p *bufPool) Get() []byte {
 	if p == nil {
 		// functional default: no reuse.
-		return make([]byte, p.blen)
+		return nil
```
puellanivis, line 20: Hm… this was put here as a safeguard, but it never runs, because it was a “just in case”. It definitely is a nil-pointer deref, but changing this to `return nil` won’t actually fix any problems, because this `Get()` needs to return a byte-slice that has an actual length. Returning `nil` would break things worse than a nil-pointer deref: i.e. it breaks things silently.

I think the only answer here is unfortunately to do something like explicitly panic. 🤔

alanskind: Thanks for the explanation!

puellanivis:
Closing in favor of https://github.com/pkg/sftp/pull/450 which first removed the condition, but then added a better safe-guard that asserts the stronger requirement about the behavior of the bufPool that `p.blen` must be a postive integer.

PR #452, opened 2021-07-09
[bugfix] FileZilla parses the directory listing in longname

puellanivis:
Not even joking:
```
        wchar_t chr = permissionToken[0];
        …
        if (chr == 'd' || chr == 'l') {
                entry.flags |= CDirentry::flag_dir;
        }
```

### DIFF
--- a/server_unix.go
+++ b/server_unix.go
@@ -6,21 +6,33 @@ package sftp
 import (
 	"fmt"
 	"os"
-	"path"
 	"syscall"
 	"time"
 )
 
-func runLsStatt(dirent os.FileInfo, statt *syscall.Stat_t) string {
+// ls -l style output for a file, which is in the 'long output' section of a readdir response packet
+// this is a very simple (lazy) implementation, just enough to look almost like openssh in a few basic cases
+func runLs(dirname string, dirent os.FileInfo) string {
 	// example from openssh sftp server:
 	// crw-rw-rw-    1 root     wheel           0 Jul 31 20:52 ttyvd
 	// format:
 	// {directory / char device / etc}{rwxrwxrwx}  {number of links} owner group size month day [time (this year) | year (otherwise)] name
 
 	typeword := runLsTypeWord(dirent)
-	numLinks := statt.Nlink
-	uid := statt.Uid
-	gid := statt.Gid
+
+	var numLinks uint64 = 1
+	if dirent.IsDir() {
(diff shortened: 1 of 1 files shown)


drakkan:
Hi,

while this patch should work for the server implementation, it likely does not help for the request server or for the fileinfo returned from the client implementation. If you get directory contents from an sftp client and then serve the returned os.Fileinfo directly over sftp (using a request server in my case) the issue happen too. We can merge this patch in the meantime and fix these other issue in future

puellanivis:
I tested this specifically with `sftpgo` with an `sshfs` backend to an `openssh` server, with Filezilla as the endpoint client, and it works.

Both server implementations use the `runLs` command internally for responding to `SSH_FXP_READDIR` requests: https://github.com/pkg/sftp/blob/master/request.go#L419

The only places where we setup an `sshFxpNameAttr` that doesn’t have a `LongName` assignment is in tests, and the only places where we do not use `runLs` is on responses to `SSH_FXP_REALPATH` and `SSH_FXP_READLINK`.

drakkan:
> I tested this specifically with `sftpgo` with an `sshfs` backend to an `openssh` server, with Filezilla as the endpoint client, and it works.
> 
> Both server implementations use the `runLs` command internally for responding to `SSH_FXP_READDIR` requests: https://github.com/pkg/sftp/blob/master/request.go#L419
> 
> The only places where we setup an `sshFxpNameAttr` that doesn’t have a `LongName` assignment is in tests, and the only places where we do not use `runLs` is on responses to `SSH_FXP_REALPATH` and `SSH_FXP_READLINK`.

ops, great, thank you! Let me remove my workarounds within sftpgo (it worked without this patch too) and I'll confirm in few minutes. Thanks!!!

puellanivis:
After you confirm, we should be good to merge. I tested the compile with `aix/ppc64` (the weirdest `Nlink` value) and it worked, so everything that was already working, should still be working fine.

drakkan:
wow, it works. Thanks and sorry for my laziness, I could have solved this issue a long time ago instead of adding workarounds inside SFTPGo

drakkan (approved):

puellanivis:
🤷‍♀️ It was a really weird set of symptoms, so no need to be sorry. You had stuff to do, and tracking down weird bugs isn’t always a high priorty thing. 🤣

PR #455, opened 2021-08-02
Cleanup Request mutex usage

puellanivis:
As I was working on the `filexfer` PR again, I noticed some use of mutex-protected values outside of holding a mutex, and then not only do all the copies point to the same mutex, but also, because this mutex has to be set, the use of `&sftp.Request{}` will always generate an invalid structure that will panic.

I realize the pointer to mutex was there so that `*r2 = *r` would work and not generate an error during `go vet` that a mutex is being copied, but this design unfortunately creates even more issues than it solved. I tried a bunch of different ways to keep the pointer, but still have `&sftp.Request{}` not immediately be incurably invalid, but any solutions to this _also_ require a mutex, which is the very problem we’re trying to avoid. We’re then stuck with having to copy values around the mutex. :slightly_frowning_face: A redesign might be necessary to avoid all mutex usage, I am unsure.

Along with cleaning up the mutex usage, I linted a lot of `request.go`, breaking some lines up, paragraphing, and removing an unused parameter to `runLs()`, so we can remove some unused path cleanup work, which then removes some imports in both `server.go` and `request.go`.

### DIFF
--- a/request.go
+++ b/request.go
@@ -503,18 +503,20 @@ func filelist(h FileLister, r *Request, pkt requestPacket) responsePacket {
 			return statusFromError(pkt.id(), err)
 		}
 
-		ret := &sshFxpNamePacket{
-			ID: pkt.id(),
-		}
+		var nameAttrs []sshFxpNameAttr
 
 		for _, fi := range finfo {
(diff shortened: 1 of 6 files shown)


drakkan (approved):
LGTM, thank you

PR #465, opened 2021-09-03
fix bug:RemoveDirectory return SSH_FX_FAILURE.

aschoolboy:
fix bug:if the directory contains files or directories, RemoveDirectory will return SSH_FX_FAILURE.

### DIFF
--- a/client.go
+++ b/client.go
@@ -8,6 +8,7 @@ import (
 	"math"
 	"os"
 	"path"
+	"path/filepath"
 	"sync"
 	"sync/atomic"
 	"syscall"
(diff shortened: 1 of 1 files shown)


puellanivis:
> … An error will be returned if no directory
> with the specified path exists, or if the specified directory is not
> empty, or if the path specified a file system object other than a
> directory.  The server responds to this request with a SSH_FXP_STATUS
> message.

https://datatracker.ietf.org/doc/html/draft-ietf-secsh-filexfer-02

Code is already working as intended and according to the standard.

Issue #466, opened 2021-09-04
statfs hangup, regression in commit 861a8ea (Feb 22, 2021, v1.13.0)


AkihiroSuda:
https://github.com/lima-vm/sshocker hangs up when it is compiled with commit 861a8ea (Feb 22, 2021, v1.13.0) and later.

Reproduction step:

```bash
git clone https://github.com/lima-vm/sshocker.git
cd sshocker
git checkout v0.2.0
go get github.com/pkg/sftp@861a8eaf5c37b4ad5b2f1d0fc61004101bbdcb49
make
make install
mkdir /tmp/foo
sshocker -v /tmp/foo:/home/user/mnt:ro example.com df
```
(the remote host `example.com` is Ubuntu 21.04 with sshfs 2.10.0, the local host is macOS 11)

The `df` command executed on the remote host hangs up because `statfs(2)` hangs up.

This is a regression in commit 861a8ea (Feb 22, 2021, v1.13.0).
The previous commit 1d73fd92d5dd1779f4123b6063d9bbb052600d4f works as expected.

puellanivis:
Poking around at things, it does not appear that the `statfs` call is actually locking up. But the call definitely is still locking up, so I’m poking at things now.

puellanivis:
Found the issue: https://github.com/pkg/sftp/commit/861a8eaf5c37b4ad5b2f1d0fc61004101bbdcb49#diff-cc2782728612ca8e00dada25397c6e9cd54e8952a3ed46e98b114fc4bc5293deL17

I will be pushing a PR soon to address the issue.

puellanivis:
@AkihiroSuda could you try out this branch: https://github.com/pkg/sftp/pull/467 and see if it resolves the issue? If it does, we’ll merge it and cut a patch release.

AkihiroSuda:
Yes, #467 works, thank you

puellanivis:
Closing as resolved in `v1.13.3`

PR #467, opened 2021-09-05
Missing ID passthru on statfs

puellanivis:
Issue https://github.com/pkg/sftp/issues/466 raised an issue of a lockup that was ascribed to `statfs` hanging, however, the problem is that `Server` was responding with `ID = 0` as it wasn’t being populated. It looked like the issue was `statfs` because it was the last thing that happened before the combination of binaries started sleeping, but it had actually returned.

This error caused `sshfs` to freakout, because it had not sent a request with `ID = 0` (which is technically an invalid ID anyways) and print `request 0 not found`. But as `sshfs` was still waiting for the response it just threw out, everything desynced and started waiting forever.


AkihiroSuda (approved):
Thanks

PR #471, opened 2021-09-27
CI: run test cases also on 32-bit arch

drakkan:
this should help to detect errors on 32-bit archs


puellanivis (approved):

puellanivis:
Hm, looks like `-race` cannot run with `GOARCH=386`.

PR #478, opened 2021-10-30
Translate EACCES to PERMISSION_DENIED.

tommie:
This is returned by (POSIX) open(2) if there is a permission issue,
e.g. opening a read-only file for writing.


puellanivis:
Tested against all `GOOS` for `GOARCH=amd64` `GOARCH=386` as well as `js/wasm` and `aix/ppc64` all either pass tests or compile. 👍

PR #480, opened 2021-11-11
Stop ReadFromWithConcurrency submitting out of order writes

ncw:
This is for discussion. This is causing a real problem in rclone https://github.com/rclone/rclone/issues/5806 but there may be a better way of fixing.

----

Before this change, ReadFromWithConcurrency would very often submit writes to the server out of order due to small timing differences when lots of blocks were queued and ready to send.

This causes more work for the server as seeking about the file isn't free. Writing chunks of a file out of order can also cause fragmentation and consequently loss of performance on that file.

This patch causes the write to the channel to be handled synchronously but the acks are received asynchronously allowing for many write packets to be in flight as before.

### DIFF
--- a/client.go
+++ b/client.go
@@ -1424,16 +1424,21 @@ func (f *File) Write(b []byte) (int, error) {
 	return n, err
 }
 
-func (f *File) writeChunkAt(ch chan result, b []byte, off int64) (int, error) {
-	typ, data, err := f.c.sendPacket(ch, &sshFxpWritePacket{
+func (f *File) startWriteChunkAt(ch chan result, b []byte, off int64) {
+	f.c.dispatchRequest(ch, &sshFxpWritePacket{
 		ID:     f.c.nextID(),
 		Handle: f.handle,
 		Offset: uint64(off),
 		Length: uint32(len(b)),
 		Data:   b,
 	})
+}
+
+func (f *File) finishWriteChunkAt(ch chan result) error {
+	s := <-ch
+	typ, data, err := s.typ, s.data, s.err
 	if err != nil {
-		return 0, err
+		return err
 	}
 
 	switch typ {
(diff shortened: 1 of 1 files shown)


ncw:
I speed tested this over a long wide pipe and I couldn't see any difference in speed before and after the patch, both about 40 MiB/s falling all the way back to 900KiB/s with concurrent writes disabled as you'd expect.

ncw:
Thinking about this some more

- this idea could be used in other reading and writing parts of the library
- can replies to sftp messages come back out of order? If not then starting `concurrency` go routines isn't necessary, only one is needed

drakkan:
Hi,

thanks for the contribution, it seems fine to me, however this code was written by @puellanivis, at the moment she has some personal issues, please wait some time for her to have a chance to have a look to the code, thanks for your patience

puellanivis:
Indeed, sequential ordering of requests is probably a good idea, and I think I was writing it into the client updates.

There are some things I think that should be addressed, but as @drakkan mentioned, I'm currently out. (In hospital) and haven't been able to really do anything yet. This also happened as I was moving into a new apartment, and now missed the appointment for connecting my internet. 😫

ncw:
Thanks for taking a look and happy to make changes when you are up to it. No hurry though and look after yourself.

puellanivis:
Oh, I did want to mention, I did similar work already in PR 443, it might help to review that PR, and compare to what you're doing here. That'll end up being one of the first things I do myself.

ncw:
Any thoughts on this PR?

puellanivis (commented):
I had recommended reviewing PR-443 and paralleling what was done there, as it would be the first thing I would do. Doing so helped me spot a bug in the existing code, that we can fix as part of this PR.

Also, a number of little “trivial” things could have been caught ahead of review that would nevertheless would make the code inconsistent between two highly similar tasks.

puellanivis:
Functionality was merged in with #482

Issue #479, opened 2021-11-02
sftp.Client.Rename()  giving permission denied, but works with other sftp clients


iamneal:
I have written some code that takes:
- user credentials
- a sftp server to connect to
- a filename to process
 
My code takes this data all this data, then:
- makes an sftp.Client to the server using the user credentials
- reads the file from the server
- moves the file on the server to an archive so it isn't processed twice, or deletes it if the move fails

my code works fine for  reading the file from the server, but when it attempts to do a `client.Rename(filename, path.Join("Archive", "filename"))` it is giving me: 

```
sftp: "Failed to rename file /EXAMPLE_FILE 2021-11-02.csv" (SSH_FX_FAILURE)
```

similarly when I attempt to do `client.Remove(filename)` I am getting:

```
sftp: "User <redacted> does not have directory make, remove or rename permissions." (SSH_FX_PERMISSION_DENIED)
```

I thought maybe the sftp server didn't give our user enough permissions to move/delete files, but the rename worked  using the same user creds when using both command line sftp OpenSSH_8.1p1, and Filezilla

The user permissions on the server allow us to read, put, and rename files in its userspace. It cannot mkdir, but the archive already exists. I can't give much more detail on the server architecture, because I don't have access, but if there  is any other info I can provide, I'll try

my code look pretty much like this:

```go
func processFile(file string, client *sftp.Client) error {
  f, err := client.Open(file)
  if err != nil {
    return err
  }
  readFile(f)

  f.Close()

  err = client.Rename(file, archiveFilename)
  if err != nil {
    logError(err)
    err = client.Remove(file)
    if err != nil {
      logError(err)
    }
  }
  return nil
}
```

I appreciate any help on this, thanks

puellanivis:
I've been in and out of hospital all week, and have been without internet due to a recent move. Just wanted to give an FYI on this thread. I've been thinking about it, but haven't had much I can poke at.

puellanivis:
I’ve been percolating on this for awhile. The only thing I can really think of is that the filenames are not being given as absolute? It’s hard to say without for instance example text from say OpenSSH_8.1p1. Seeing the actual commands issued, and the output resulting from it (obviously redacting secret information).

Understanding the full scope of what you’re trying to do as an example of `sftp` CLI commands can point to where assumptions may have been mistaken.

iamneal:
Ok so update on this, I figured out the cause.  

[...]

PR #482, opened 2021-12-05
Sequentially issue write requests, process results concurrently

puellanivis:
Supersedes #480

Originally, it was assumed that we could issue write requests in any arbitrary order, and while this holds and is correct, it can cause locality issues for servers. Issuing write requests sequentially means the server can efficiently handle the writes in order without needing to seek around the file so much.

So, this takes care to issue requests sequentially, and then handles the results concurrently as they come back.

Also, this changes `writeAtWithConcurrency` to issue sequential writes as well, not just `ReadFromWithConcurrency`

### DIFF
--- a/client.go
+++ b/client.go
@@ -1461,10 +1461,19 @@ func (f *File) writeAtConcurrent(b []byte, off int64) (int, error) {
 	cancel := make(chan struct{})
 
 	type work struct {
-		b   []byte
+		id  uint32
+		res chan result
+
 		off int64
 	}
 	workCh := make(chan work)
+
+	concurrency := len(b)/f.c.maxPacket + 1
+	if concurrency > f.c.maxConcurrentRequests || concurrency < 1 {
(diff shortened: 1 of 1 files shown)


drakkan:
Thanks @puellanivis ! Is the note about [concurrent writes](https://pkg.go.dev/github.com/pkg/sftp#UseConcurrentWrites) still valid after this PR? Do we still need to truncate the file after an error?

puellanivis:
It’s definitely much less likely, but I think the standard defines that while things can potentially be handled by the server out-of-order, the responses returned have to be essentially identical to doing them in order. So, it’s probably unlikely that this concern would remain.

drakkan:
I did a quick test and the patch seems ok for me, @ncw can you please provide your feedback? Thanks

drakkan (approved):

puellanivis:
I’m going to go ahead and merge this. I would like to also address #479 before we cut any sort of release, though.

PR #484, opened 2021-12-18
Add Client.OpenFileWithMode

### LINKED ISSUE #335: client: Is there any way to create a file on the server with specific permissions?

greatroar:
This allows creating a file with specific permissions. Fixes #335.

The name of the new method is the one suggested by @puellanivis over at #335.

It's tempting to immediately declare `os.OpenFile` deprecated, but there may be a reason to keep it indefinitely: the protocol apparently allows opening a file with or without a mode, letting the server determine the mode in the former case.

### DIFF
--- a/packet.go
+++ b/packet.go
@@ -662,7 +662,8 @@ type sshFxpOpenPacket struct {
 	ID     uint32
 	Path   string
 	Pflags uint32
-	Flags  uint32 // ignored
+	Flags  uint32
+	Perm   uint32
 }
 
 func (p *sshFxpOpenPacket) id() uint32 { return p.ID }
@@ -672,12 +673,19 @@ func (p *sshFxpOpenPacket) MarshalBinary() ([]byte, error) {
 		4 + len(p.Path) +
 		4 + 4
 
+	if p.Flags&sshFileXferAttrPermissions != 0 {
+		l += 4
+	}
+
 	b := make([]byte, 4, l)
 	b = append(b, sshFxpOpen)
 	b = marshalUint32(b, p.ID)
 	b = marshalString(b, p.Path)
 	b = marshalUint32(b, p.Pflags)
 	b = marshalUint32(b, p.Flags)
+	if p.Flags&sshFileXferAttrPermissions != 0 {
+		b = marshalUint32(b, p.Perm)
+	}
 
 	return b, nil
 }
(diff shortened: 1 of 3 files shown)


packet.go
```diff
@@ -662,7 +662,8 @@ type sshFxpOpenPacket struct {
 	ID     uint32
 	Path   string
 	Pflags uint32
-	Flags  uint32 // ignored
+	Flags  uint32
```
puellanivis, line 665: My primary concern here is that this `Flags` does not support all of the `SSH_ATTR` possibilities, but there is nothing to warn that it _only_ supports `SSH_FILEXFER_ATTR_PERMISSIONS`.

It’s also still quite a bit of a hack, and every hack we make can limit our future. Consider: we’ve already added an `OpenFile` and its signature doesn’t match `os.OpenFile`; we’re stuck with all of the errors spelled `ErrSshFxOk` even though they’ve all already been replaced by `ErrSSHFxOk`; By adding the `ReadFromWithConcurrency` we’re kind of stuck with exposing it, even though our new sequential issuing eliminates almost all of the issues that were theoretically possible with enabling concurrency; we’re on the precipice of handling client side packets with the new `filexfer` package that handle all features of all packets, and avoids lots of the “hacky” packet things here.

I’d really rather people just make due without a “correct” `OpenFile` until we have a chance to cut `v2` where we can fix the signature of `OpenFile` itself.

packet.go
```diff
@@ -690,8 +698,12 @@ func (p *sshFxpOpenPacket) UnmarshalBinary(b []byte) error {
 		return err
 	} else if p.Pflags, b, err = unmarshalUint32Safe(b); err != nil {
 		return err
-	} else if p.Flags, _, err = unmarshalUint32Safe(b); err != nil {
+	} else if p.Flags, b, err = unmarshalUint32Safe(b); err != nil {
 		return err
+	} else if p.Flags&sshFileXferAttrPermissions != 0 {
```
puellanivis, line 703: Where is the `server.go` and `request_server.go` code to handle this expanded feature set? This is one of the further complications to this sort of a PR: this is not simply a client-side concern, there is also a concern about how to handle the server side of the feature.

puellanivis (changes requested):
I have concerns about tacking on Yet Another Workaround.

packet.go
greatroar: Fair enough. Feel free to close.

packet.go
greatroar: Understood, but note that this is a broader issue, as other client implementations can also send open packets with attributes that the servers will ignore.

puellanivis: True.

PR #485, opened 2021-12-19
Add File.ReadFrom test on a SectionReader

greatroar:
This tests ReadFrom's determining concurrency from the estimated input
size. go test -cover -integration, before:

    coverage: 77.3% of statements

After:

    coverage: 79.2% of statements

### DIFF
--- a/client_integration_test.go
+++ b/client_integration_test.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"crypto/sha1"
 	"errors"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"math/rand"
@@ -164,7 +165,7 @@ func netPipe(t testing.TB) (io.ReadWriteCloser, io.ReadWriteCloser) {
 	return c1, r.Conn
 }
 
-func testClientGoSvr(t testing.TB, readonly bool, delay time.Duration) (*Client, *exec.Cmd) {
+func testClientGoSvr(t testing.TB, readonly bool, delay time.Duration, opts ...ClientOption) (*Client, *exec.Cmd) {
 	c1, c2 := netPipe(t)
 
 	options := []ServerOption{WithDebug(os.Stderr)}
@@ -183,7 +184,7 @@ func testClientGoSvr(t testing.TB, readonly bool, delay time.Duration) (*Client,
 		wr = newDelayedWriter(wr, delay)
 	}
 
-	client, err := NewClientPipe(c2, wr)
+	client, err := NewClientPipe(c2, wr, opts...)
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -194,13 +195,13 @@ func testClientGoSvr(t testing.TB, readonly bool, delay time.Duration) (*Client,
 
 // testClient returns a *Client connected to a locally running sftp-server
 // the *exec.Cmd returned must be defer Wait'd.
-func testClient(t testing.TB, readonly bool, delay time.Duration) (*Client, *exec.Cmd) {
+func testClient(t testing.TB, readonly bool, delay time.Duration, opts ...ClientOption) (*Client, *exec.Cmd) {
 	if !*testIntegration {
 		t.Skip("skipping integration test")
 	}
 
 	if *testServerImpl {
-		return testClientGoSvr(t, readonly, delay)
+		return testClientGoSvr(t, readonly, delay, opts...)
 	}
 
 	cmd := exec.Command(*testSftp, "-e", "-R", "-l", debuglevel) // log to stderr, read only
@@ -228,7 +229,7 @@ func testClient(t testing.TB, readonly bool, delay time.Duration) (*Client, *exe
 		t.Skipf("could not start sftp-server process: %v", err)
 	}
 
-	sftp, err := NewClientPipe(pr, pw)
+	sftp, err := NewClientPipe(pr, pw, opts...)
 	if err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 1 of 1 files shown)


client_integration_test.go
```diff
@@ -1489,6 +1489,49 @@ func TestClientReadFrom(t *testing.T) {
...
 
+func TestClientReadFromSectionReader(t *testing.T) {
+	const (
+		packetSize = 1024
+		bufsize    = 4 * packetSize
+		skip       = 100
+		filesize   = bufsize - skip
+	)
+
+	sftp, cmd := testClient(t, READWRITE, NODELAY,
```
puellanivis, line 1500: This function call is more readable when all on one line, and splitting the options out is awkward.
https://github.com/golang/go/wiki/CodeReviewComments#line-length

client_integration_test.go
```diff
@@ -1489,6 +1489,49 @@ func TestClientReadFrom(t *testing.T) {
...
+		MaxPacketChecked(packetSize),
+		UseConcurrentWrites(true))
+	defer cmd.Wait()
+	defer sftp.Close()
+
+	d, err := ioutil.TempDir("", "sftptest-readfrom-sectionreader")
+	if err != nil {
+		t.Fatal(err)
```
puellanivis, line 1508: I recommend a bit of additional information as to why the fatal is being called: `t.Fatal("unexpected error:", err)`

client_integration_test.go
```diff
@@ -1489,6 +1489,49 @@ func TestClientReadFrom(t *testing.T) {
...
+	}
+	defer os.RemoveAll(d)
+
+	buf := bytes.NewReader(make([]byte, bufsize))
+	r := io.NewSectionReader(buf, 0, bufsize)
+
+	// Skip some bytes so that r.Size is no longer accurate.
+	// ReadFrom uses it to determine concurrency,
+	// but should not rely on it to determine how many bytes to send.
+	r.Seek(skip, io.SeekStart)
```
puellanivis, line 1518: We should be able to do this with the `bytes.Reader` itself. https://pkg.go.dev/bytes@go1.17.5#Reader.Size

client_integration_test.go
```diff
@@ -1489,6 +1489,49 @@ func TestClientReadFrom(t *testing.T) {
...
+
+	f := path.Join(d, "writeTest")
+	w, err := sftp.Create(f)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer w.Close()
+
+	n, err := w.ReadFrom(r)
+	assert.EqualValues(t, bufsize-skip, n)
```
puellanivis, line 1528: `bufsize-skip` is the same value as `filesize`

Ooo… `EqualValues` is a nice function over `Equal` which often requires you to cast your constants to specific values.

client_integration_test.go
```diff
@@ -1489,6 +1489,49 @@ func TestClientReadFrom(t *testing.T) {
...
+
+	fi, err := os.Stat(f)
+	require.NoError(t, err)
```
puellanivis, line 1531: Let’s not mix `if err != nil { t.Fatal("unexpected error:", err) }` and `require.NoError(t, err)` in new code. (Though, I’m sure you might be copying existing code.)

I would prefer the former rather than the later. We’ve used a lot of the later, more because we are adapting existing use of `assert.NoError(t, err)`.

puellanivis:
Looks like quite excellent work!

client_integration_test.go
greatroar: Done.

client_integration_test.go
greatroar: You mean not use a SectionReader at all?

client_integration_test.go
greatroar: Done.

client_integration_test.go
greatroar: Changed to filesize.

client_integration_test.go
greatroar: Done.

client_integration_test.go
puellanivis: yeah, we should be able to do the same test with `bytes.Reader` and not need `SectionReader`

greatroar: That's less strong a test, though, since the type switch inside ReadFrom first checks if its Reader has a Len method, and bytes.Reader does:

https://github.com/pkg/sftp/blob/1849af66afae504778b257e286cd1a406ea14bc5/client.go#L1770-L1785

puellanivis: Ach so. So the SectionReader is to get a `Size, but not a `Len`?` 🤔 I think things would be clearer if we use a one off type to make it clear what exact functionality we’re testing against. That is, don’t require people to look up `SectionReader` to know exactly what functionality we’re testing against.

greatroar: Good idea. Done.

client_integration_test.go
```diff
@@ -1489,6 +1490,59 @@ func TestClientReadFrom(t *testing.T) {
...
+
+	for i, reported := range []int{-1, filesize - 100, filesize, filesize + 100} {
+		r := &wrongSizeReader{filesize, reported}
+
+		f := path.Join(d, fmt.Sprintf("writeTest%d", i))
+		w, err := sftp.Create(f)
+		if err != nil {
+			t.Fatal("unexpected error:", err)
+		}
+		defer w.Close()
```
puellanivis, line 1533: These `defer`s only trigger at the end of the function, not the end of each loop. It’s only four handles right now so it’s not so bad, but it’s a good idea to always be wary of any `defer` that appears in a loop.

client_integration_test.go
```diff
@@ -1489,6 +1490,59 @@ func TestClientReadFrom(t *testing.T) {
 	}
 }
 
+// A wrongSizeReader reads zeros and has an arbitrary Size.
```
puellanivis, line 1493: `wrongSizeReader` does not actually read zeros, because it does not modify the byte slice that is passed into `Read` at all.

We should do one of the following:
* Change the documentation to say that it is a no-op that pretends to read `actual` number of bytes, and that one can have it report any arbitrary value for `Size()` by setting `reported`. (That is, describe the actual contract the implementation given fufills.)
* Actually zero out the byte-slice passed into `Read`, in accordance with the contract this comment purposes.
* Have this wrap an `io.Reader` and then we don’t have to specify what the `Read` does at all.

(nitpick) Hm, also, one of the `wrongSizeReaders` that we make is not a wrong size. 😅

client_integration_test.go
```diff
@@ -1489,6 +1490,59 @@ func TestClientReadFrom(t *testing.T) {
...
+	defer sftp.Close()
+
+	d, err := ioutil.TempDir("", "sftptest-readfrom-size-mismatch")
+	if err != nil {
+		t.Fatal("cannot create temp dir:", err)
+	}
+	defer os.RemoveAll(d)
+
+	for i, reported := range []int{-1, filesize - 100, filesize, filesize + 100} {
+		r := &wrongSizeReader{filesize, reported}
```
puellanivis, line 1526: (style) Please use explicit field names. Using implicit field assignments is inherently brittle, as changing the order of the fields will changes the order the assignment values have to be in, thus making refactoring harder.

Using explicit field names means not only is it clear exactly which field is getting which value, but also the underlying struct fields can be reordered or even expanded at any time without needing to adjust this code.

client_integration_test.go
greatroar: Replaced by subtests. Btw., the test just above hold more files open than this one...

client_integration_test.go
greatroar: Replaced by your suggestion of wrapping an io.Reader. The code is shorter to boot.

client_integration_test.go
greatroar: Done.

puellanivis (approved):
Looks good.

(Waiting for anyone else to weigh in. If not, I’ll merge in a few days, at latest just after the new year.)

PR #486, opened 2022-01-05
feat: add server opt Chroot() to change root path

UnightSun:


drakkan:
Hi,

test cases don't pass. I didn't tested your implementation but after a quick look at the code I think chroot can be escaped via symlinks and this could or could not be desiderable.  Additionally things like `p = path.Join(svr.rootPath, p)` assume UNIX paths.

I'm basically +0 here, chroot can be easily achieved, outside the library, using the request server implementation

UnightSun:
You're right, I will check out request server, thank you for your time

server_test.go
```diff
@@ -249,3 +252,208 @@ func TestServerWithBrokenClient(t *testing.T) {
...
+			} else {
+				require.Nil(t, err)
+				assert.True(t, f.IsDir())
+			}
+		}
+		t.Run("mkdir", func(t *testing.T) {
+			relPath := "/dir/mkdir"
+			require.Nil(t, client.Mkdir(relPath))
+			// dir should be created
+			assertDir(filepath.Join(rootPath, relPath), true)
```
puellanivis, line 335: The `*testing.T` used by `assertDir` is captured by closure to be the one from the `t.Run("dir", …)`, but this `*testing.T` is from the `t.Run("mkdir", …)`. Calling functions on the parent `*testing.T` for a daughter `t.Run` is improper usage.

server_test.go
```diff
@@ -249,3 +252,208 @@ func TestServerWithBrokenClient(t *testing.T) {
...
+			f, err := os.Lstat(absPath)
+			if !exist {
+				require.True(t, os.IsNotExist(err))
+			} else {
+				require.Nil(t, err)
+				assert.True(t, f.IsDir())
+			}
+		}
+		t.Run("mkdir", func(t *testing.T) {
+			relPath := "/dir/mkdir"
```
puellanivis, line 332: This is technically not a relative path.

server_test.go
```diff
@@ -249,3 +252,208 @@ func TestServerWithBrokenClient(t *testing.T) {
...
+
+	client, server := clientServerPair(t, Chroot(rootPath))
+	defer client.Close()
+	defer server.Close()
+
+	t.Run("stat", func(t *testing.T) {
+		// prepare, create file and symlink for stat
+		require.Nil(t, os.MkdirAll(filepath.Join(rootPath, "/stat"), 0700))
+		regular := "/stat/regular"
+		symlink := "/stat/symlink"
```
puellanivis, line 275:
```
const regular, symlink = "/stat/regular", "/stat/symlink"
```

puellanivis (commented):
I see now the PR is closed, so I will just submit what I have drafted here.

PR #487, opened 2022-01-06
feat: add ChrootHandler() for specific root path

UnightSun:
1. As drakkan's suggestion, refactor to `RequestServer` and `Handlers`, take example by `InMemHandler`
2. Optimized the test code refer to puellanivis' advice.
3. Tested on windows 10 and WSL ubuntu.

### DIFF
--- a/request-chroot.go
+++ b/request-chroot.go
@@ -1,0 +1,341 @@
+// +build !plan9
+
+package sftp
+
+// This serves as an example of how to implement the request server handler as
+// well as a dummy backend for testing. It implements an in-memory backend that
+// works as a very simple filesystem with simple flat key-value lookup system.
+
+import (
+	"errors"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+	"syscall"
+)
+
+// ChrootHandler returns a Hanlders object with the specific root path.
+func ChrootHandler(rootPath string) Handlers {
+	chroot := &chroot{
+		rootPath: rootPath,
+	}
+	return Handlers{chroot, chroot, chroot, chroot}
+}
+
+type chroot struct {
+	rootPath string
+}
+
+func (fs *chroot) toOsFlags(flags uint32) (int, error) {
+	var osFlags int
+	if flags&sshFxfRead != 0 || flags&sshFxfWrite != 0 {
+		osFlags |= os.O_RDWR
+	} else if flags&sshFxfWrite != 0 {
+		osFlags |= os.O_WRONLY
+	} else if flags&sshFxfRead != 0 {
+		osFlags |= os.O_RDONLY
+	} else {
+		// how are they opening?
+		return 0, syscall.EINVAL
+	}
+	if flags&sshFxfCreat != 0 {
+		osFlags |= os.O_CREATE
+	}
+	if flags&sshFxfTrunc != 0 {
+		osFlags |= os.O_TRUNC
+	}
+	if flags&sshFxfExcl != 0 {
+		osFlags |= os.O_EXCL
+	}
+	return osFlags, nil
+}
+
+const OUT_PATH_PREFIX = ".." + string(filepath.Separator)
+
+func (fs *chroot) getRealPath(p string) (string, error) {
+	p = filepath.Join(fs.rootPath, p)
+	if fs.rootPath != "" && fs.rootPath != "/" {
+		rel, err := filepath.Rel(fs.rootPath, p)
+		if err != nil || strings.HasPrefix(rel, OUT_PATH_PREFIX) {
+			return "", os.ErrInvalid
+		}
+	}
+	return p, nil
+}
+
+func (fs *chroot) getRelativePath(p string) (string, error) {
+	if fs.rootPath != "" && fs.rootPath != "/" {
+		rel, err := filepath.Rel(fs.rootPath, p)
+		if err != nil || strings.HasPrefix(rel, OUT_PATH_PREFIX) {
+			return "", os.ErrInvalid
+		}
+		p = rel
+	}
+	return cleanPath(p), nil
+}
+
+func (fs *chroot) Fileread(r *Request) (io.ReaderAt, error) {
+	if r.Flags&sshFxfRead == 0 {
+		return nil, os.ErrInvalid
+	}
+	return fs.OpenFile(r)
+}
+
+func (fs *chroot) Filewrite(r *Request) (io.WriterAt, error) {
+	if r.Flags&sshFxfWrite == 0 {
+		return nil, os.ErrInvalid
+	}
+	return fs.OpenFile(r)
+}
+
+func (fs *chroot) OpenFile(r *Request) (WriterAtReaderAt, error) {
+	_ = r.WithContext(r.Context()) // initialize context for deadlock testing
+	return fs.openfile(r.Filepath, r.Flags)
+}
+
+func (fs *chroot) openfile(pathname string, flags uint32) (*os.File, error) {
+	osFlags, err := fs.toOsFlags(flags)
+	if err != nil {
+		return nil, err
+	}
+	realPath, err := fs.getRealPath(pathname)
+	if err != nil {
+		return nil, err
+	}
+	file, err := os.OpenFile(realPath, osFlags, 0644)
+	if err != nil {
+		return nil, err
+	}
+	stat, err := file.Stat()
+	if err != nil {
+		return nil, err
+	}
+	if stat.IsDir() {
+		return nil, os.ErrInvalid
+	}
+	return file, nil
+}
+
+func (fs *chroot) Filecmd(r *Request) error {
+	_ = r.WithContext(r.Context()) // initialize context for deadlock testing
+	switch r.Method {
+	case "Setstat":
+		file, err := fs.openfile(r.Filepath, sshFxfWrite)
+		if err != nil {
+			return err
+		}
+
+		if r.AttrFlags().Size {
+			return file.Truncate(int64(r.Attributes().Size))
+		}
+
+		return nil
+
+	case "Rename":
+		return fs.rename(r.Filepath, r.Target, false)
+
+	case "Rmdir":
+		return fs.rmdir(r.Filepath)
+
+	case "Remove":
+		// IEEE 1003.1 remove explicitly can unlink files and remove empty directories.
+		// We use instead here the semantics of unlink, which is allowed to be restricted against directories.
+		return fs.unlink(r.Filepath)
+
+	case "Mkdir":
+		return fs.mkdir(r.Filepath)
+
+	case "Link":
+		return fs.link(r.Filepath, r.Target)
+
+	case "Symlink":
+		// NOTE: r.Filepath is the target, and r.Target is the linkpath.
+		return fs.symlink(r.Filepath, r.Target)
+	}
+
+	return errors.New("unsupported")
+}
+
+func (fs *chroot) rename(oldpath, newpath string, posix bool) error {
+	oldpath, err := fs.getRealPath(oldpath)
+	if err != nil {
+		return err
+	}
+	newpath, err = fs.getRealPath(newpath)
+	if err != nil {
+		return err
+	}
+	// SFTP-v2: "It is an error if there already exists a file with the name specified by newpath."
+	// This varies from the POSIX specification, which allows limited replacement of target files.
+	if !posix {
+		_, err := os.Lstat(newpath)
+		if !os.IsNotExist(err) {
+			return os.ErrExist
+		}
+	}
+	return os.Rename(oldpath, newpath)
+}
+
+func (fs *chroot) PosixRename(r *Request) error {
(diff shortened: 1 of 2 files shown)


request-chroot_test.go
```diff
@@ -0,0 +1,428 @@
...
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+const outbound = "../outbound"
+
+func getChrootTuple(t *testing.T) (string, ClientWrapper) {
+	testRoot, err := ioutil.TempDir(os.TempDir(), "sftp")
```
puellanivis, line 24: We should be using a different directory name for each test, so that we can see at a glance which test failed to cleanup if it unintentionally gets left hanging.

request-chroot_test.go
```diff
@@ -0,0 +1,428 @@
...
+	)
+	rand.Seed(time.Now().UnixNano())
+	files := make(map[string]int)
+	bytes := make([]byte, MAX_SIZE)
+	require.NoError(t, os.MkdirAll(filepath.Join(rootPath, "readdir"), 0755))
+	// TODO: generate complicate file tree
+	for i := 0; i < FILE_CNT; i++ {
+		fileName := fmt.Sprintf("file_%d", i)
+		filePath := filepath.Join(rootPath, "readdir", fileName)
+		fileSize := rand.Intn(MAX_SIZE)
```
puellanivis, line 416: rather than `MAX_SIZE` we should be using `len(bytes)`.

request-chroot_test.go
```diff
@@ -0,0 +1,428 @@
...
+func TestChrootReadDir(t *testing.T) {
+	rootPath, client := getChrootTuple(t)
+	defer client.Close(t)
+	const (
+		FILE_CNT = 10
+		MAX_SIZE = 1024
+	)
+	rand.Seed(time.Now().UnixNano())
+	files := make(map[string]int)
+	bytes := make([]byte, MAX_SIZE)
```
puellanivis, line 410: Try and avoid using `bytes` as a variable name, as it will shadow the `bytes` package.

Habit is to use `b` or `buf`.

request-chroot_test.go
```diff
@@ -0,0 +1,428 @@
...
+}
+
+func TestChrootReadDir(t *testing.T) {
+	rootPath, client := getChrootTuple(t)
+	defer client.Close(t)
+	const (
+		FILE_CNT = 10
+		MAX_SIZE = 1024
+	)
+	rand.Seed(time.Now().UnixNano())
```
puellanivis, line 408: Avoid using things that effect a mutation of a global state in tests tests like this. (`rand.Seed()` operates on `rand.globalRand`)

Instead use a locally scoped PRNG: `r := rand.New(rand.NewSource(time.Now().UnixNano()))` https://pkg.go.dev/math/rand@go1.17.5#New

request-chroot_test.go
```diff
@@ -0,0 +1,428 @@
...
+	files := make(map[string]int)
```
puellanivis, line 409: Since this is a map of filenames to filesizes, I think this is more clearly named `fileSizes`. I woudl expect a map named `files` to be of type `map[string]*os.File`

request-chroot_test.go
```diff
@@ -0,0 +1,428 @@
...
+		fileName := fmt.Sprintf("file_%d", i)
+		filePath := filepath.Join(rootPath, "readdir", fileName)
+		fileSize := rand.Intn(MAX_SIZE)
+		require.NoError(t, ioutil.WriteFile(filePath, bytes[:fileSize], 0644))
+		files[fileName] = fileSize
+	}
+
+	stats, err := client.ReadDir("/readdir")
+	require.NoError(t, err)
+	res := make(map[string]int)
```
puellanivis, line 423: I think `res` here might have a more clear name about its purpose?

request-chroot_test.go
```diff
@@ -0,0 +1,428 @@
...
+		// Stat() should retrieve file info which symlink point to
+		assert.EqualValues(t, len(content), f.Size())
+		assert.True(t, f.Mode().IsRegular())
+	}
+
+	// outer path will be converted relate to root, and /{root}/outbound is not exist
+	_, err = client.Lstat(outbound)
+	assert.ErrorIs(t, err, os.ErrNotExist)
+
+	// but we allow server to use symlink point to an outer file
```
puellanivis, line 109: This is not how `chroot` should work:
```
sheleg ~$ sudo chroot /tmp/foobar /bin/sh
sh-4.4# ls -l
total 12
drwxr-xr-x 2 1000 100 4096 Jan  6 12:37 bin
drwxr-xr-x 2 1000 100 4096 Jan  6 12:38 lib64
lrwxrwxrwx 1 1000 100   11 Jan  6 12:31 outlink -> ../outbound
-rw-r--r-- 1 1000 100    8 Jan  6 12:32 regular
lrwxrwxrwx 1 1000 100    7 Jan  6 12:32 symlink -> regular
sh-4.4# cat regular
regular
sh-4.4# cat symlink
regular
sh-4.4# cat outlink
cat: outlink: No such file or directory
```

puellanivis (changes requested):
There’s a critical issue that symlinks breaking out of a `chroot` should _not_ be followed.

As well, the utility of this code is limited, as it has a single backing implementation on the existing filesystem, and thus cannot compose for example the `Chroot` functionality with `InMemHandler`.

As mentioned at the top of the file, we need to be mindful about polluting the root package with more example implementations. The `InMemHandler` should not be in the root package, but it’s unfortunately too late for that.

The implementation of both `InMemHandler` and `Chroot` are obscured from the user by a return of type `Handlers` which means they’re not even particularly useful examples.

request-chroot.go
```diff
@@ -0,0 +1,341 @@
+// +build !plan9
+
+package sftp
+
+// This serves as an example of how to implement the request server handler as
+// well as a dummy backend for testing. It implements an in-memory backend that
+// works as a very simple filesystem with simple flat key-value lookup system.
```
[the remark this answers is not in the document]
UnightSun: Maybe I can move this to example/chroot-request-server?

request-chroot.go
```diff
@@ -0,0 +1,341 @@
...
+	}
+	return osFlags, nil
+}
+
+const OUT_PATH_PREFIX = ".." + string(filepath.Separator)
+
+func (fs *chroot) getRealPath(p string) (string, error) {
+	p = filepath.Join(fs.rootPath, p)
+	if fs.rootPath != "" && fs.rootPath != "/" {
+		rel, err := filepath.Rel(fs.rootPath, p)
```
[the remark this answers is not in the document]
UnightSun: Sorry, but I didn't get the point, what's wrong here?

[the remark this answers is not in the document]
puellanivis: You shouldn’t need to check if the `fs.rootPath != "/"` since `filepath.Rel(fs.rootPath, p)` would not generate an error, and the `strings.HasPrefix(rel, "../")` should never show up, and even if `filepath.Rel` didn’t error on `/../a/b` and instead just return `../a/b` we would want to return `os.ErrInvalid` the same.

request-chroot.go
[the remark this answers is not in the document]
puellanivis: 🤷‍♀️ That would be a more appropriate place for it, so long as you think it should only ever be presented as an example, and thus not immediately useful code to anyone without a copy-paste.

dmlb2000:
This had been idle for almost a year now? What is the current state of this?

puellanivis:
The PR still pollutes the root level namespace, and does not prevent traversal of symlinks outside of the chroot. I suppose we could just close this PR as “will not merge without rewrite”.

nevun:
> The PR still pollutes the root level namespace, and does not prevent traversal of symlinks outside of the chroot. I suppose we could just close this PR as “will not merge without rewrite”.

Regarding the symlink issue it seems to me is works as it should?

In the chroot I created (as root in a normal shell) the following:
```
[root@comp upload3923341595]# ls -latr
total 4
drwx------. 3 user user  30 Jan 16 11:28 ..
-rw-r--r--. 1 user user 260 Jan 16 11:28 watermarked_files.yar
lrwxrwxrwx. 1 root root  11 Jan 16 11:28 passwd -> /etc/passwd
drwx------. 2 user user  71 Jan 16 11:39 .
```

Connected to the sftp chroot daemon:

```
sftp> ls -latr
-rw-r--r--    1 1002     1002          260 Jan 16 10:28 watermarked_files.yar
lrwxrwxrwx    1 0        0              11 Jan 16 10:28 passwd
sftp> get passwd
Fetching /passwd to passwd
stat remote: No such file or directory
sftp>

sftp> symlink /etc/nonexisting foo
sftp> symlink /etc/passwd bar
sftp> ls -latr
-rw-r--r--    1 1002     1002          260 Jan 16 10:28 watermarked_files.yar
lrwxrwxrwx    1 0        0              11 Jan 16 10:28 passwd
lrwxrwxrwx    1 1002     1002           42 Jan 16 10:51 foo
lrwxrwxrwx    1 1002     1002           37 Jan 16 10:51 bar
sftp> get foo
Fetching /foo to foo
stat remote: No such file or directory
sftp> get bar
Fetching /bar to bar
stat remote: No such file or directory
sftp>
```

As per your comment on the PR where stat() returns an error but lstat() does not: stat should return info on the symlink and not the file linked to so it seems fine to me.

But I am probably misunderstanding something..

puellanivis:
Could you have some sort of logging of what files it’s actually trying to use on the server side? I’m curious to know if the fetching is failing because of a convenient misalignment, or if it’s genuinely preventing access.

nevun:
Seems you were right.

I could not trigger the Stat() call from my sftp client but running the test showed with some logging turned on shows that the Stat implementation in the chroot code is incorrect. It should not allow Stat to do os.Stat(), it needs something like this that resolves the actual file and checks whether or not it is inside the chroot:

```
 func (fs *chroot) Filelist(r *Request) (ListerAt, error) {
        _ = r.WithContext(r.Context()) // initialize context for deadlock testing
        realPath, err := fs.getRealPath(r.Filepath)
        if err != nil {
                return nil, err
        }
        switch r.Method {
        case "List":
                files, err := os.ReadDir(realPath)
                if err != nil {
                        return nil, err
                }
                var res listerat = make(listerat, 0, len(files))
                for _, file := range files {
                        stat, err := file.Info()
                        if err != nil {
                                return nil, err
                        }
                        res = append(res, stat)
                }
                return res, nil

        case "Stat":
+               fmt.Printf("Stat: %s\n", realPath)
+               f, err := os.Lstat(realPath)
+               if f.Mode()&os.ModeSymlink == os.ModeSymlink {
+                       symlink, err := os.Readlink(realPath)
+                       if err != nil {
+                               fmt.Printf("Stat: could not resolve link %s: %v\n", realPath, err)
+                               return nil, err
+                       }
+                       res, err := fs.getRelativePath(symlink)
+                       if err != nil {
+                               fmt.Printf("Stat: symlink %s could not be looked up on chroot: %v\n", symlink, err)
+                               return nil, err
+                       }
+                       realPath = res
+               }
+
                file, err := os.Stat(realPath)
                if err != nil {
+                       fmt.Printf("Stat %s returning error\n", realPath)
                        return nil, err
                }
+               fmt.Printf("Stat %s returning %v\n", realPath, file)
                return listerat{file}, nil

        case "Readlink":
                symlink, err := os.Readlink(realPath)
                if err != nil {
                        return nil, err
                }
                res, err := fs.getRelativePath(symlink)
```

With that the TestChrootStat test fails as one would expect:

```
$ go test -run "TestChrootStat*"
[ .. snipped other test output .. ]
TestChrootStat: stat outlink
getRealPath /outlink
getRealPath with chroot: /tmp/sftp1226763008/sftp/outlink
getRealPath filePath.Rel -> outlink
getRealPath: strings.HasPrefix(outlink, ../) -> false
getRealPath returning /tmp/sftp1226763008/sftp/outlink
Stat: /tmp/sftp1226763008/sftp/outlink
getRelativePath: err: <nil> and strings.HasPrefix(../outbound, ../) == true
Stat: symlink /tmp/sftp1226763008/outbound could not be looked up on chroot: invalid argument
--- FAIL: TestChrootStat (0.00s)
    request-chroot_test.go:101:
        	Error Trace:	/user/kode/sftp/request-chroot_test.go:101
        	Error:      	Received unexpected error:
        	            	file does not exist
        	Test:       	TestChrootStat
    request-chroot_test.go:115:
        	Error Trace:	/user/kode/sftp/request-chroot_test.go:115
        	Error:      	Received unexpected error:
        	            	sftp: "invalid argument" (SSH_FX_FAILURE)
        	Test:       	TestChrootStat
FAIL
exit status 1
$
```
Sorry for the noise. I think your arguments for closing this PR stands.

puellanivis:
Thanks for the check! It’s always valuable to validate assumptions!

I’ll go ahead and close this, but if anyone wants to pick it up again knowing just how complex the problem is, and willing to deal with all the seriously nitpicky pedantic nonsense required, I’ll be happy to take another look.

Issue #489, opened 2022-01-19
ioutil.ReadAll never finishes downloading a file with disabled concurency


glebteterin:
I'm connecting to a server with `sftp.UseConcurrentReads(false)`, then open a file with Client.Open and read it to memory with ioutil.ReadAll. Due to the `readAtSequential` function (client.go) suppressing io.EOF error, the ioutil.ReadAll can't know when the file is fully read and won't finish.

I'm not sure I understand the reasoning for File.Read() doesn't return io.EOF, it seems like a bug introduced during refactoring. If so, would you accept a PR that fixes it?

puellanivis:
OOooooof… yeah, this looks like it was copied from `writeToSequential` which has an entirely different semantic here. We definitely shouldn’t be swallowing the `io.EOF` for `readAtSequential`.

I’ll be happy to accept a PR. 👍

puellanivis:
Fixed in https://github.com/pkg/sftp/pull/490

PR #490, opened 2022-01-19
Fix missing io.EOF when concurency is disabled #489

glebteterin:
Removed suppression of io.EOF in readAtSequential and updated tests


puellanivis:
Thanks!

Issue #481, opened 2021-11-24
Is file modification time preservation working properly?


vansante:
As a user of this very nice library I am getting some bug reports from some of my users that file modification timestamps are not getting preserved when copying/uploading a file over SFTP.

The modification time is supposed to be set with a `setstat` call.

I did some digging around and came to the following conclusions:

- It works for Filezilla
- It does not work when using the `sftp` CLI utility
- It does not work when using the `sshfs` CLI utility

I did some deeper digging with logging and noticed that when I upload the file with Filezilla the `setstat` call comes in _after_ the file has been opened, data was written and the file handle was closed again. However, for the non functioning CLI utils, the order is:

- File handle opened 
- `setstat` call comes in
- File data written
- File closed

Which of course has the effect of the modification timestamps in the `setstat` call being immediately overwritten.

At this point I tested the posix behaviour of this with the following simple test:
```golang
func TestModificationTime(*testing.T) {
	f, _ := os.Create("testfile.txt")
	os.Chtimes(f.Name(), time.Now(), time.Unix(824522342, 0))
	f.WriteAt([]byte("sefesfse"), 10)
	f.Close()
}
```
Which resulted in the modification time being `time.Now()` instead of `time.Unix(824522342, 0)` as I suspected it would.

This led me to believe this is probably not an issue in my own code as I'm just executing in the order of the commands I am getting. To verify this I started looking at the behaviour of `drakkan/sftpgo` by firing up the latest docker container:
`docker run --name some-sftpgo -p 8080:8080 -p 2022:2022 -d "drakkan/sftpgo:latest"`

I created a basic `test` user on local storage and uploaded a file (note the `-p` flags which are supposed to preserve timestamps):

```
pvansanten@host  sftp -p -P 2022 test@127.0.0.1:
test@127.0.0.1's password: 
Connected to 127.0.0.1.
Changing to: /.
sftp> PUT -p /home/pvansanten/Pictures/dockergopher.png
Uploading /home/pvansanten/Pictures/dockergopher.png to /dockergopher.png
/home/pvansanten/Pictures/dockergopher.png                                                                                                               100%  396KB  41.2MB/s   00:00    
sftp> ls -al
-rw-r--r--    1 1000     1000       405692 Nov 24 09:41 dockergopher.png
```
and again, the timestamps are not being preserved, which led me to here, making this issue. 

[...]

PR #493, opened 2022-02-02
Add example request server that uses the local filesystem

vansante:
I made this to test issue #481, it might be more useful in the future to test more stuff :)

### DIFF
--- a/request-filesystem.go
+++ b/request-filesystem.go
@@ -1,0 +1,160 @@
+package sftp
+
+import (
+	"errors"
+	"io"
+	"os"
+	"path/filepath"
+	"time"
+)
+
+// FilesystemHandler returns a Handlers object with the test handlers.
+func FilesystemHandler(root string) Handlers {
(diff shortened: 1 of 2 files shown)


drakkan:
to test #481 it is good, if you plan to get this merged you should move the implementation should be moved away from the root path. Please see the first comment in #487

request-filesystem.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+	return Handlers{r, r, r, r}
+}
+
+type fsRoot struct {
+	root string
+}
+
+func (fs *fsRoot) path(path string) string {
+	// Make sure we do not escape the root:
```
drakkan, line 24: are you sure that a symlink outside the root path will not be followed?

vansante:
It looks like that other implementation in #487 is already way better than mine, so I'll close this MR. If I had seen that MR I probably wouldnt have bothered with this :sweat_smile:.

PR #498, opened 2022-03-01
request server: add WithStartDirectory option

drakkan:
Alternative approach to support a start directory.

I tested #497 and it works fine but I don't like that pkg/sftp clean the path and I have to clean the raw path again in my code.

This approach seems cleaner and the `RealPathFileLister` is useless if we go this way, we can remove this interface.

The only drawback is that [NewRequest](https://github.com/pkg/sftp/blob/aa9a37d6394f61dbcfadf3478acf028a4f39b866/request.go#L147) now requires an already cleaned path. Prior to this patch the path passed to NewRequest was converted to an absolute UNIX path using `/` as base. I'll try to fix this compatibility issue later

### DIFF
--- a/request-server.go
+++ b/request-server.go
@@ -27,6 +27,8 @@ type RequestServer struct {
 	*serverConn
 	pktMgr *packetManager
 
+	startDirectory string
+
 	mu           sync.RWMutex
 	handleCount  int
 	openRequests map[string]*Request
@@ -47,6 +49,14 @@ func WithRSAllocator() RequestServerOption {
 	}
 }
 
+// WithStartDirectory sets a start directory to use as base for relative paths.
+// If unset the default is "/"
+func WithStartDirectory(startDirectory string) RequestServerOption {
+	return func(rs *RequestServer) {
+		rs.startDirectory = cleanPath(startDirectory)
+	}
+}
+
 // NewRequestServer creates/allocates/returns new RequestServer.
 // Normally there will be one server per user-session.
 func NewRequestServer(rwc io.ReadWriteCloser, h Handlers, options ...RequestServerOption) *RequestServer {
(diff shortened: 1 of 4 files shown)


drakkan:
Hi @puellanivis 

thanks for your review, I'll fix the PR based on your comments.

I know that removing the RealPathLister is a breaking change, but it does not work for real use cases. It only works if you have a client that initially requests the start directory and then sends absolute paths based on the response. I think it is very unlikely that anyone actually uses this interface. If someone is really using it, they should fix their code by migrating to the new server option.

puellanivis:
We can just label the interface as deprecated, and say that implementing it won’t actually do anything. The same as we have `ErrSshFxOk`. Saying “they should fix their code” is all fine and good, but semver is pretty strict on this, and we need to bump the major version in order to remove it.

request-server.go
```diff
@@ -47,6 +49,14 @@ func WithRSAllocator() RequestServerOption {
 	}
 }
 
+// WithStartDirectory sets a start directory to use as base for relative paths.
+// If unset the default is "/"
+func WithStartDirectory(startDirectory string) RequestServerOption {
+	return func(rs *RequestServer) {
+		rs.startDirectory = cleanPath(startDirectory)
```
puellanivis, line 56: 🤔 What happens if someone uses `WithStartDirectory("")`?

drakkan: `/` will be used as base

puellanivis (approved):

drakkan:
Thank you!

shanehooker:
I am not finding an example of how to use this with a real file system. I am a bit confused by the InMemHandler(). I would have prefered the WithStartDirectory() to have been a method applied to NewServer() which would have done the same thing as starting 'sftp-server -d [startDirectory]'.

Is there an example of how this can be called and work with a real file system?

puellanivis:
The reason for `NewRequestServer` over `NewServer` is that the latter can only operate upon the local OS filesystem. Meaning, you cannot establish some other backing store, like s3, or even a caching temporary filesystem held in memory, lost during system restart (that is `InMemoryHandler`). So, we have elected to essentially “deprecate” `NewServer` as an insuffienciently extendable interface.

`sftp-server -d [startDirectory]` works by first doing an `os.Chdir()` before starting up the service, and letting the inherent OS filesystem management take care of the rest. You should be able to get the `NewServer` to work the same way.

PR #499, opened 2022-03-03
writeToSequential: improve tests for write errors

drakkan:
I tryed to improve test cases for write errors as requested in #494

Of course I kept the author for the original commits

### DIFF
--- a/client_integration_test.go
+++ b/client_integration_test.go
@@ -1249,6 +1249,44 @@ func TestClientReadSequential(t *testing.T) {
 	}
 }
 
+type writerFunc func(b []byte) (int, error)
+
+func (f writerFunc) Write(b []byte) (int, error) {
+	return f(b)
+}
+
+func TestClientWriteSequential_WriterErr(t *testing.T) {
+	sftp, cmd := testClient(t, READONLY, NODELAY)
+	defer cmd.Wait()
+	defer sftp.Close()
+
+	sftp.disableConcurrentReads = true
+	d, err := ioutil.TempDir("", "sftptest-writesequential")
+	require.NoError(t, err)
+
+	defer os.RemoveAll(d)
+
+	f, err := ioutil.TempFile(d, "write-sequential-test")
+	require.NoError(t, err)
+	fname := f.Name()
+	content := []byte("hello world")
+	f.Write(content)
+	f.Close()
+
+	sftpFile, err := sftp.Open(fname)
+	require.NoError(t, err)
+	defer sftpFile.Close()
+
+	want := errors.New("error writing")
+	n, got := io.Copy(writerFunc(func(b []byte) (int, error) {
+		return 10, want
+	}), sftpFile)
+
+	require.Error(t, got)
+	assert.ErrorIs(t, want, got)
+	assert.Equal(t, int64(10), n)
+}
+
 func TestClientReadDir(t *testing.T) {
 	sftp1, cmd1 := testClient(t, READONLY, NODELAY)
 	sftp2, cmd2 := testClientGoSvr(t, READONLY, NODELAY)
--- a/client_integration_test.go
+++ b/client_integration_test.go
@@ -1249,10 +1249,19 @@ func TestClientReadSequential(t *testing.T) {
 	}
 }
 
-type writerFunc func(b []byte) (int, error)
+type lastChunkErrSequentialWriter struct {
+	expected      int
+	written       int
+	writtenReturn int
+}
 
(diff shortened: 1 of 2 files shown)


puellanivis (commented):
If we rebreak the code, will the test actually fail? I cloned the code locally, and checked, and it seems like it does not.

Monkeying around with things, I think we need to use `MaxPacketChecked(3)`, seed text of `"12345"`, change the check for `len(b) == 2` to `len(b) > 3`, and the test is `written != 4`.

Reasons:
* With seed text `"hello world"` we are not reaching `EOF` of the input before the read error
* With seed text `"1234"` and `MaxPacketChecked(2)` the original code does not fail
* With seed text `"12345"` and `MaxPacketChecked(3)` the second call is made with `len(b) == 2` not `len(b) == 3`; so we need `len(b) > 3`.
* With the new `MaxPacketChecked(3)` we now do one 3-byte write, and one 1-byte write. This also allows us to distinguish from bad-behavior full-write 3+2, from good-behavior 3+1.

🙃 This corner case is actually pretty hard to reproduce intentionally with minimal code.

drakkan:
If now you set the seed text to `1234` the test will not fail since the second write will always return an error

client_integration_test.go
```diff
@@ -1249,55 +1249,50 @@ func TestClientReadSequential(t *testing.T) {
...
+	writeCounter int
 }
 
 func (w *lastChunkErrSequentialWriter) Write(b []byte) (int, error) {
-	chunkSize := len(b)
-	w.written += chunkSize
-	if w.written == w.expected {
-		return w.writtenReturn, errors.New("test error")
+	if len(b) > 3 {
+		return 0, errors.New("this writer require maxPacket <= 3, please set MaxPacketChecked(2)")
```
puellanivis, line 1259: The error message here is unclear. The first time this runs, if `maxPacket > 3` then we haven’t set `MaxPacketChecked(3)`, and we _need_ this to be `MaxPacketChecked(3)` not just any other value. Maybe, we add the check in `w.writeCounter == 1` block to ensure that `len(b) == 3` for the first call? That should be a strong enough assertion to avoid all other values than 3.

client_integration_test.go
```diff
@@ -1249,55 +1249,50 @@ func TestClientReadSequential(t *testing.T) {
 	}
 }
 
+// this writer requires maxPacket = 3 and always returns an error for the second write call
 type lastChunkErrSequentialWriter struct {
-	expected      int
-	written       int
-	writtenReturn int
+	writeCounter int
```
puellanivis, line 1254: [nit] we don’t really need the `write` of the `writeCounter`. This is a writer after all.

client_integration_test.go
```diff
@@ -1249,55 +1249,50 @@ func TestClientReadSequential(t *testing.T) {
...
-	gotWritten, gotErr := sftpFile.writeToSequential(&w)
-	require.NotErrorIs(t, io.EOF, gotErr)
-	require.Equal(t, int64(shortWrite), gotWritten)
+	w := &lastChunkErrSequentialWriter{}
+	written, err := sftpFile.writeToSequential(w)
+	assert.Error(t, err)
+	if written != 4 {
+		t.Errorf("sftpFile.Write() = %d, but expected 4", written)
+	}
+	assert.Equal(t, 2, w.writeCounter)
```
puellanivis, line 1295: [nit] This test is relying on the default typing of untyped integers assigned to `interface{}` to be `int`, and thus not as stable as it could be.

puellanivis (commented):
these are just nitpicks.

puellanivis (approved):
Go ahead and merge as you will with or without addressing the nits.

drakkan:
> Go ahead and merge as you will with or without addressing the nits.

If you have other comments I will be glad to fix the code, as always thanks!

client_integration_test.go
```diff
@@ -1249,6 +1249,53 @@ func TestClientReadSequential(t *testing.T) {
...
+// this writer requires maxPacket = 3 and always returns an error for the second write call
+type lastChunkErrSequentialWriter struct {
+	counter int
+}
+
+func (w *lastChunkErrSequentialWriter) Write(b []byte) (int, error) {
+	w.counter++
+	if w.counter == 1 {
+		if len(b) != 3 {
+			return 0, errors.New("this writer require maxPacket = 3, please set MaxPacketChecked(3)")
```
puellanivis, line 1261: [typo] “requires”

puellanivis (approved):

Issue #506, opened 2022-05-16
sftp.NewClient returns "EOF" erro


mraxrorxon:
I create sshClient like this:
if sshClient, err = ssh.Dial("tcp", addr, clientConfig); err != nil {
		return nil, err
}; and sshClient is created successfully, after that create sftp client like this:

if sftpClient, err = sftp.NewClient(sshClient); err != nil {
		return nil, err
} here I get an error "EOF"
Please help this, where is a problam?

puellanivis:
Could you maybe share the whole of your function(s)? And encase it in a triple-backtick code block? There is not really enough information here to diagnose what might be going on.

PR #513, opened 2022-06-29
More context for EOF during client setup

puellanivis:
Addresses concerns raised from #506 where a bare EOF was being returned, and thus being quite unclear about where the EOF happened.

This adds more context to the returned errors, and replaces `EOF` with `UnexpectedEOF` since we definitely aren’t expecting one.

However, I’m thinking a better message should be returned than even unexpected EOF. I’ll be digging through documentation to see what the best option here is.

### DIFF
--- a/client.go
+++ b/client.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/binary"
 	"errors"
+	"fmt"
 	"io"
 	"math"
 	"os"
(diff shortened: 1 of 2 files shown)


puellanivis:
The secsh-filexfer-02 document is silent about what to do in this case. I think however the error `"server unexpectedly closed connection"` might be even more clear than an `unexpected EOF`.

drakkan:
> The secsh-filexfer-02 document is silent about what to do in this case. I think however the error `"server unexpectedly closed connection"` might be even more clear than an `unexpected EOF`.

I agree

client.go
```diff
@@ -267,8 +275,13 @@ func (c *Client) nextID() uint32 {
 func (c *Client) recvVersion() error {
 	typ, data, err := c.recvPacket(0)
 	if err != nil {
+		if err == io.EOF {
+			return io.ErrUnexpectedEOF
```
drakkan, line 279: Maybe we could do something like this here

```
return fmt.Errorf("server unexpectedly closed connection: %w", io.ErrUnexpectedEOF)
```

so `errors.Is(err, io.ErrUnexpectedEOF)` will work

puellanivis:
Alright, I think we’re good now. 👍

drakkan:
Thanks!

PR #522, opened 2022-09-09
Add ablility for server to pass client request info to callers

capnspacehook:
Allows users to build an audit log of what client requests were sent, or take other actions depending on details in client requests.

### DIFF
--- a/server.go
+++ b/server.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"io/ioutil"
 	"os"
 	"path/filepath"
@@ -28,6 +29,7 @@ const (
 type Server struct {
 	*serverConn
 	debugStream   io.Writer
+	reqCallback   RequestCallback
 	readOnly      bool
 	pktMgr        *packetManager
 	openFiles     map[string]*os.File
@@ -83,6 +85,10 @@ func NewServer(rwc io.ReadWriteCloser, options ...ServerOption) (*Server, error)
 	s := &Server{
 		serverConn:  svrConn,
 		debugStream: ioutil.Discard,
+		// default to setting reqCallback to a function that does nothing
+		// so we don't have to check if reqCallback is nil every time we
+		// want to call it
+		reqCallback: func(_ RequestPacket) {},
 		pktMgr:      newPktMgr(svrConn),
 		openFiles:   make(map[string]*os.File),
 	}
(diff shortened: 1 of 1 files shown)


server.go
```diff
@@ -517,8 +696,23 @@ func (p *sshFxpSetstatPacket) respond(svr *Server) responsePacket {
...
+		}
+	}
+	svr.reqCallback(RequestPacket{
+		Type: Setstat,
+		Path: p.Path,
+		Attributes: &Attributes{
+			Size:             fileSize,
+			UID:              fileUID,
+			GID:              fileGID,
+			Permissions:      (*fs.FileMode)(permissions),
```
puellanivis, line 710: It is well known that `fs.FileMode` is not binary compatible with [`filexfer.FileMode`](https://pkg.go.dev/github.com/pkg/sftp@v1.13.5/internal/encoding/ssh/filexfer#FileMode)

server.go
```diff
@@ -115,6 +121,71 @@ func ReadOnly() ServerOption {
...
+	GID              *uint32
+	Permissions      *fs.FileMode
+	AccessTime       *time.Time
+	ModificationTime *time.Time
+}
+
+// RequestCallback is the type of function called by a Server when
+// WithRequestCallback is set. reqPacket will contain details of
+// a client request the server has recieved.
+type RequestCallback func(reqPacket RequestPacket)
```
puellanivis, line 178: It makes no sense to pass this struct by value. In fact, since it contains pointers, this mixed-use case of value and reference can easily lead to misunderstanding and false assumptions about the safety of modifying the argument.

If a struct contains a pointer (or a map, or a slice, or an interface, because these also contain pointers) then it should always be passed also by pointer.

server.go
```diff
@@ -115,6 +121,71 @@ func ReadOnly() ServerOption {
...
+	Path string
+	// TargetPath is the new path in a rename request, or the new path in
+	// a symlink request.
+	TargetPath string
+	// Flags is any flags the request passed.
+	Flags      uint32
+	Attributes *Attributes
+	// Err is the error that occured from handling the request.
+	Err error
+}
```
puellanivis, line 162: This appears to be recreating the already existing [sftp.Request](https://pkg.go.dev/github.com/pkg/sftp#Request) struct.

What does this purposed API provide that cannot already be served via the `RequestServer` abstraction? I’m quite loathe to build in and export Yet Another Abstraction that fits no better the SSH_FILEXFER standards it is supposed to be representing.

puellanivis (changes requested):
¿This seems to provide far less utility than the existing `RequestServer` already provides?

capnspacehook:
My mistake, this is indeed inferior to the current `RequestServer`.

Issue #407, opened 2021-02-22
Proposal: io/fs presents us an opportunity to rework the whole api into a v2


puellanivis:
Most of the proposal is right there in the title. What with `io/fs` being rolled out in go1.16, we have a big opportunity to remodel, simplify, and rework our API. Current issues, as I see them:
* The current `sftp.Client` implements `github.com/kr/fs`, but we could switch to model more compatible with `io/fs.FS`
* The in-memory test filesystem cannot be removed from the main package without a breaking change, ideally if kept, this should be a subpackage.
* The `ErrSshFxXyz` errors are still around, because they cannot be removed without a breaking change. 
* The `RequestServer` and `Server` implementations are significantly different and almost parallel implementations.
* The `RequestServer` could be better implemented via a base interface with additional extension interfaces. (à la `io/fs`)
* From an earlier discussion, many of the raw implementation concepts of the protocol itself could be isolated into its own subpackage, providing a better separation of responsibilities.
* Some features can be reassessed for utility, and either properly folded into functionality without needing feature flags (like the `allocator`?) or dropped entirely (we could switch `WriteTo` so that it does not depend upon filesizes, thus saving us the need of an `UseFstat()` option).

drakkan:
@puellanivis I'm interested to improve the `RequestServer`. I don't like the `Server` implementation: even if you need to work with the local filesystem I think it is better to handle the logic within your own application. For example using the request server you can easily chroot your users, using the Server implementation we need to add this feature to the library and in a way that is good for all use cases. 

This is my plan for the `RequestServer`:

1) we currently use multiple workers for write packets. For appends we should process the received packets sequentially, we should not rely on packet offset in this case according to the specs
2) I'll try to add V6 support and some filexfer extensions like file hashing and server side copying

I'm not sure how we can improve the request interfaces, they already seem well designed. If we want to change them we should do it in a way that does not require to checking if an interface is implemented for each packet (for example for writes). Do you already have some sort of draft?

The V6 support is a long term plan, I have other priorities for now, I think I can work on this after releasing SFTPGo 2.1

puellanivis:
Well, as mentioned above, the current Client targets a different FS object interface than does the new standard library FS defines. Many of the distinctions are probably small, like `ReadDir` instead of `Readdir`, but these are the sort of thing that we would need to make a breaking change in order to implement.

[...]

PR #528, opened 2022-10-14
Add support for working directory in Server

mafredri:
This commit allows the working directory for the (old) Server
implementation to be changed without doing a `os.Chdir` first.

The feature can be enabled with `sftp.WithServerWorkingDirectory(dir)`
passed as an option to `sftp.NewServer`.

It is useful when the `sftp` is used as part of a larger service that
does more than just serve `sftp` and using `os.Chdir` is not an option.

The fallback behavior (when the option is not specified) is that the
path remains unmodified (as before).

**Motivation:** I know the new `sftp.RequestServer` already supports this, however, since it currently only has an in-memory filesystem and is potentially going to see a larger refactor in the future, I think it's a good idea to include it for this older implementation as well.


drakkan:
Hi,

thanks for your PR. At first glance it looks ok, but I have never user the server implementation myself. If no other contributors have time to review this PR you will have to wait a while for a proper review. Thanks for your patience.

Please note that no larger refactor is planned for the request server, we just don't provide a sample implementation that uses the local filesystem but you can easily write your own implementation and it is much more customizable than the old server implementation. I'm not sure we'll continue to maintain the old server implementation in the future, maybe we'll rewrite it using the request server interfaces (and probably outside the root package)

request-plan9.go
```diff
@@ -16,7 +16,15 @@ func testOsSys(sys interface{}) error {
 	return nil
 }
 
-func toLocalPath(p string) string {
+func toLocalPath(workDir, p string) string {
+	if workDir != "" {
+		if !filepath.IsAbs(p) && !path.IsAbs(p) {
```
puellanivis, line 21: I’m not sure why you’re using `filepath.IsAbs(p)` here. Paths coming in over the wire should be conformant with `path` only.

That’s why we use `filepath.FromSlash(p)` below.

request-unix.go
```diff
@@ -22,6 +24,14 @@ func testOsSys(sys interface{}) error {
 	return nil
 }
 
-func toLocalPath(p string) string {
+func toLocalPath(workDir, p string) string {
+	if workDir != "" {
+		if !filepath.IsAbs(p) && !path.IsAbs(p) {
```
puellanivis, line 29: `filepath.Abs` is equivalent to `path.IsAbs` under build tag `unix`.

request-unix.go
```diff
@@ -22,6 +24,14 @@ func testOsSys(sys interface{}) error {
...
+			// Ensure input is always in the same format.
+			p = filepath.ToSlash(p)
```
puellanivis, line 31: `filepath.ToSlash(p)` is a no-op under build tag `unix`.

request_test.go
```diff
@@ -247,3 +249,108 @@ func TestOpendirHandleReuse(t *testing.T) {
...
+			goos: "windows",
+			args: args{workDir: "C:\\Users\\User", p: "/C:/file"},
+			want: "C:\\file",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(fmt.Sprintf("%s %s %s", tt.goos, tt.args.workDir, tt.name), func(t *testing.T) {
+			if runtime.GOOS != tt.goos {
+				t.Skipf("Skipping test for %s on %s", tt.goos, runtime.GOOS)
+			}
```
puellanivis, line 347: Prefer to not skipping tests like this. There are already build tags available, which will exclude the tests or not depending on the filename of the test.

request_test.go
```diff
@@ -247,3 +249,108 @@ func TestOpendirHandleReuse(t *testing.T) {
...
+		},
+		{
+			name: "absolute unix-like path",
+			goos: "windows",
+			args: args{workDir: "C:\\Users\\User", p: "/C:/file"},
+			want: "C:\\file",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(fmt.Sprintf("%s %s %s", tt.goos, tt.args.workDir, tt.name), func(t *testing.T) {
```
puellanivis, line 344: I don’t know why we wouldn’t just pre-compose all information necessary for the test into `tt.name` rather than sprinting it every time.

request_test.go
```diff
@@ -247,3 +249,108 @@ func TestOpendirHandleReuse(t *testing.T) {
...
+			if runtime.GOOS != tt.goos {
+				t.Skipf("Skipping test for %s on %s", tt.goos, runtime.GOOS)
+			}
+
+			var s Server
+			if tt.args.workDir != "" {
+				_ = WithServerWorkingDirectory(tt.args.workDir)(&s)
+			}
+			assert.Equal(t, tt.want, toLocalPath(s.workDir, tt.args.p), "wrong local path")
```
puellanivis, line 353: I’m not sure what utility there is in instantiating a whole `Server` when we could just pass `tt.args.workDir` directly into `toLocalPath`.

request_windows.go
```diff
@@ -14,7 +14,15 @@ func testOsSys(sys interface{}) error {
 	return nil
 }
 
-func toLocalPath(p string) string {
+func toLocalPath(workDir, p string) string {
+	if workDir != "" {
+		if !filepath.IsAbs(p) && !path.IsAbs(p) {
```
puellanivis, line 19: The code `lp := filepath.FromSlash` below demonstrates that the `p` argument should always be compliant with `path.Join` here.

request_test.go
```diff
@@ -247,3 +249,108 @@ func TestOpendirHandleReuse(t *testing.T) {
...
+		{
+			name: "relative path with .",
+			goos: "windows",
+			args: args{workDir: "C:\\Users\\User", p: "."},
+			want: "C:\\Users\\User",
+		},
+		{
+			name: "absolute path",
+			goos: "windows",
+			args: args{workDir: "C:\\Users\\User", p: "C:\\file"},
```
puellanivis, line 327: By the standards, such a `p` value should not be transmitted over the wire.

puellanivis (commented):
As drakkan says, you should be able to just implement a raw OS `RequestServer` and I just don’t see the value in adding features to a deprecated implementation.

There are a large number of caveats in path handling, and making sure they’re all correct, compatible, and won’t break others is way harder than one might think.

request-plan9.go
mafredri: I was working around an issue where I was getting the work dir prefixed multiple times or things like `C:\C:\`. Ultimately some of that was probably a bug I had in the original implementation, but reading the code made me unsure if I could trust the input.

For instance, correct me if I'm wrong but this behavior doesn't conform to that guarantee:

Path is converted here:

https://github.com/pkg/sftp/blob/88c72168ab7facc5aca8de2be92d458c9153a2e0/server.go#L259

Set here:

https://github.com/pkg/sftp/blob/88c72168ab7facc5aca8de2be92d458c9153a2e0/server.go#L269

And a local path passed to `toLocalPath` here:

https://github.com/pkg/sftp/blob/88c72168ab7facc5aca8de2be92d458c9153a2e0/server.go#L449

request-unix.go
mafredri: I wanted to keep the code consistent across platforms, but I'll simplify 👍🏻

request_test.go
mafredri: I can move this to separate files, but then the test body would have to be duplicated across those files. Skipping to match the platform seems like a lesser bad to me?

request_test.go
mafredri: It would duplicate information and introduce potential for mismatch between test name and what is actually being tested.

request_test.go
mafredri: The idea was to test the newly introduced API, if the implementation of `WithServerWorkingDirectory` ever changes, the test will catch the change in behavior. I can simplify.

mafredri:
Did I misinterpret what was being discussed in https://github.com/pkg/sftp/issues/407? To me it sounds like a refactor/rework of the API, and I think it sounds wonderful TBH. But it's also a reason why I wasn't looking to work with the `RequestServer` at this time.

The fact that `RequestServer` is missing a FS implementation is also a reason I opted out, it means `Server` is what people reach for when they need SFTP, so it'd be more battle tested. (My reasoning may of course be wrong.)

> As drakkan says, you should be able to just implement a raw OS `RequestServer` and I just don’t see the value in adding features to a deprecated implementation.

I appreciate the review, but I'm left wondering if I should clean up this PR or discard it? It sounds like you're not interested in merging it but I could be wrong.

I tried to make sure that no code paths are altered unless this new feature is enabled so that it'd be easier to merge in, but I can understand if you don't want it at all.

request_windows.go
mafredri: I understand that's the intent, but it doesn't give any guarantees for the input. `FromSlash` is pretty much just no-op on a backslash-based path. And if we try to `path.Join` with a unix-y workDir, then we're in for trouble.

request-plan9.go
puellanivis:
```
tempFile := filepath.Join(t.TempDir(), "sftp") // C:\Path\To\Tmp\sftp
file, err := client.Create(tempFile)
```
I cannot find this code anywhere in the existing source code.

Such usage is fundamentally erroneous. However, the behavior of this being turned into the local path of `C:\Path\To\Workdir\C:\Path\To\Tmp\sftp` is however the correct behavior, which should lead to an error on Windows because paths cannot contain `:` and a non error in Linux because `/path/to/workdir/C:/Path/To/Temp/sftp` is valid.

Building a temp directory appropriate for the local client simply does not make any sense in the aspect of making a file on the remote server.

request-unix.go
puellanivis: Platform-specific code should always be code specific to the platform.

If the code could be written generically, then it wouldn’t need to be put in platform-specific code.

request_test.go
puellanivis: The name of a test should concisely define the entire contract of the test. This is the programmer expressing the _intent_ of the test. If the test fails to meet this intent, then there is a bug in the tests.

Is there a bug in these tests?
```
tests := []test{
  {
    name: "test if",
    input: 2,
    expect: 2,
  }
}

for _, tt := range tests {
  tt.Run(fmt.Sprintf("%s %d = %d", tt.name, tt.input, tt.expect),  func(t *testing.T) {
    require.Equal(t, tt.input, tt.expect)
  })
}
```

What about here:
```
tests := []test{
  {
    name: "test if 3 = 2",
    input: 2,
    expect: 2,
  }
}

for _, tt := range tests {
  tt.Run(fmt.Sprintf("%s %d = %d", tt.name, tt.input, tt.expect),  func(t *testing.T) {
    require.Equal(t, tt.input, tt.expect)
  })
}
```

Better yet, what about here:
```
tests := []test{
  {
    name: "test if 3 = 2",
    input: 3,
    expect: 2,
  }
  {
    name: "test if 2 = 2",
    input: 2,
    expect: 2,
  }
}

for _, tt := range tests {
  tt.Run(fmt.Sprintf("%s %d = %d", tt.name, tt.input, tt.expect),  func(t *testing.T) {
    t.Parallel()
    require.Equal(t, tt.input, tt.expect)
  })
}
```

request_test.go
puellanivis: If you want to test the newly introduced API, then you need to actually test to make sure that a request made to that server is interpreted into the correct path.

All you’re testing here is that `s.workDir` is set by `WithServerWorkingDirectory`.

request_windows.go
puellanivis: `p` should never be non-unixy. This is a requirement of the standard.

puellanivis:
😐 It seems like github had been spamming out all of my draft responses. Learn from me, don’t do code review responses after a long hard day of toiling in the salt mines of legacy code…

request_test.go
mafredri: I think I understand the point you you're making, but I don't see how this example is applicable. There'd be no difference in making a static name like `test if 2 = 3` vs using `fmt.Sprintf`. The main bug is that `tt` wasn't shadowed in the loop when using parallel tests and thus `require` getting called with the wrong arguments. (I mean the require arguments are also flipped, but I don't think that's the point you were making since all the information is already there for fixing it.)

In any case, I'll update the names :+1:.

request_test.go
mafredri: I think the nuance of what I was going for is different, I wasn't so much testing that `server.workDir` gets set correctly, I was testing that the input a user would use for `WithServerWorkingDirectory` produces the right output from `toLocalPath`. So in that sense it tested exactly what I wanted, but I understand you feel it's not the best approach. Also, my naming was lazy, I shouldn't have put `workDir` under `tt.args` to better reflect the intent.

I'll fix up my tests according to your input :+1:.

request_test.go
puellanivis: Yeah, I was trying to think of a proper example, thus the extensive edits of the comment. Brain wasn’t really working the best at the time after 10 hours of heavy debugging.

The require arguments being flipped is a bonus example of why I recommend against using the assert/require packages in general. It’s all too easy to flip arguments. But if I’m writing out the `if got != want { t.Errorf("got %d but wanted %d", want, got) }` then it’s way easier to see that the arguments are reversed.

request-plan9.go
puellanivis: Ah yes. This was a similar thing for the posix versions as well, where a test can assume that `/tmp/asdf` is the same remote as it is local, and then it breaks when using the in memory handler. 🙃

request_test.go
puellanivis: > New files called request_windows_test.go and request_unix_test.go?

Yes please.

> I looked at the codebase and there's already a lot of t.Skip present, and barely any test files utilizing build flags.

Sadly, there’s a lot of inherited code that I would write differently now.

request_test.go
puellanivis: I understood what you were trying for, but there was no guarantee that `toLocalPath` is getting called with `server.workDir`. That part of the loop has to be closed manually in the function calls.

Actually… thinking about it, if we made `toLocalPath` a receiver method off of `Server` then this test would be the correct way to test it. It also closes the loop mentioned above by making the `WithServerWorkingDirectory` argument inherent to `toLocalPath` rather than explicit.

server.go
```diff
@@ -256,13 +267,14 @@ func handlePacket(s *Server, p orderedRequest) error {
 			rpkt = statusFromError(p.ID, err)
 		}
 	case *sshFxpOpendirPacket:
-		p.Path = toLocalPath(p.Path)
+		lp := toLocalPath(s.workDir, p.Path)
```
puellanivis, line 270: There’s another `p.Path` that wasn’t replaced by `lp` in the `&sshFxpOpenPacket` packet. I’d probably recommend backing out the `s/lp/p.Path/` change instead.

Less change, less chance of new bug.

request_test.go
```diff
@@ -247,3 +248,113 @@ func TestOpendirHandleReuse(t *testing.T) {
...
+			name: "workdir and absolute path on Windows",
+			goos: "windows",
+			args: args{workDir: cleanPath("C:\\Users\\User"), p: "/C:/file"},
+			want: "C:\\file",
+		},
+		{
+			name: "workdir and non-unixy path on Windows prefixes workdir",
+			goos: "windows",
+			args: args{workDir: cleanPath("C:\\Users\\User"), p: "C:\\file"},
+			want: "C:\\Users\\User\\C:\\file",
```
puellanivis, line 348: I’d probably recommend adding a comment that while this looks like it’s a mistake, it is in fact correct behavior.

Since this is one of those fun things were the correct behavior looks like an obvious bug.

server.go
mafredri: I'm not sure I follow. The only other assignment to `p.Path` that I can find is in `(*sshFxpSetstatPacket).respond`.

This change was to prevent the following scenario: `toLocalPath(s.workDir, toLocalPath(s.workDir, p.Path))` which happens in `(*sshFxpOpenPacket).respond` because of this assignment to `p.Path`. This becomes a problem when workDir is being used and we're **not** checking `filepath.IsAbs`, resulting in double workDir prefix on Windows and Plan 9.

mafredri:
@puellanivis thanks for the thorough PR review! I've amended most of the feedback now. The only thing I have yet to revert is https://github.com/pkg/sftp/pull/528#discussion_r997951182 because doing so may introduce a bug when used with a `workDir`.

I went ahead with refactoring `toLocalPath` to be a method on `Server`. This also moved the code out of `request_*` files which seems appropriate, considering it's not used by the `RequestServer`.

server.go
puellanivis: Ach so! You’re absolutely right. The `toLocalPath` really should be deferred in that case until the `.respond` happens. Nice bug catch!

server_plan9.go
```diff
@@ -0,0 +1,30 @@
+//go:build plan9
+// +build plan9
```
puellanivis, line 2: These build tags should be unnecessary since the filename already ends with `_plan9.go`.

server_unix_test.go
```diff
@@ -0,0 +1,87 @@
...
+			// toLocalPath behavior.
+			s := &Server{}
+			if tt.withWorkDir != "" {
+				if err := WithServerWorkingDirectory(tt.withWorkDir)(s); err != nil {
+					t.Fatal(err)
+				}
+			}
+
+			if got := s.toLocalPath(tt.p); got != tt.want {
+				t.Errorf("Server.toLocalPath() = %v, want %v", got, tt.want)
```
puellanivis, line 83: I’m going to recommend the `%q` verb here instead of `%v` since we’re dealing with strings that could potentially be confusing when integrated directly.

server_unix_test.go
```diff
@@ -0,0 +1,87 @@
+//go:build !windows && !plan9
+// +build !windows,!plan9
```
puellanivis, line 2: 🤔 These tests should be able to run under plan9 as well, right? Might be better than trying to add a-whole-nother `server_plan9_test.go`. Maybe this file should be `server_nowindows_test.go`?

server_windows_test.go
```diff
@@ -0,0 +1,87 @@
+//go:build windows
+// +build windows
```
puellanivis, line 2: > If a file's name, after stripping the extension and a possible _test suffix, matches any of the following patterns:
```
*_GOOS
*_GOARCH
*_GOOS_GOARCH
```
> where GOOS and GOARCH represent any known operating system and architecture values respectively, then the file is considered to have an implicit build constraint requiring those terms (in addition to any explicit constraints in the file).

So, these build tags should also be redundant.

puellanivis (commented):
Looking nice. 👍

server_windows_test.go
mafredri: Ah, thanks for that. Wasn't aware it strips `_test`. 👍🏻

prochac:
Cheers guys, It's being awesome for me to come just a week after this feature has been landed. I was seeking for it, and it's not in release yet, and found this. Many thanks.

PR #537, opened 2023-02-13
Stop ReadFromWithConcurrency sending more data than it needs to

ncw:
It was discovered that the ReadFrom method for uploading files in
pkg/sftp was sending more data than it needed to.

This was tracked down to the ReadFromWithConcurrency method forgetting
to truncate the packets it was sending to the size Read.

This was giving the remote server more work to do as it was writing
and re-writing parts of a file.

See: https://github.com/rclone/rclone/issues/6763

### DIFF
--- a/client.go
+++ b/client.go
@@ -1674,7 +1674,7 @@ func (f *File) ReadFromWithConcurrency(r io.Reader, concurrency int) (read int64
 					Handle: f.handle,
 					Offset: uint64(off),
 					Length: uint32(n),
-					Data:   b,
+					Data:   b[:n],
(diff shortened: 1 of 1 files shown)


drakkan:
Thanks!

puellanivis:
Huh… good catch!

puellanivis:
Yeah, I was thinking of this as an alternative to the code here, basically, use `io.ReadFull` instead. I figure the change wouldn’t be so big at all, so, if you want to spin up a PR, and at least get some benchmark information about if it’s better or not…

Issue #544, opened 2023-03-27
IETF Link has rotted


puellanivis:
The IETF link https://datatracker.ietf.org/doc/html/draft-ietf-secsh-filexfer-02 has rotted, and no longer points to `-02`, but instead  silently and automagically forwards one directly to `-13`, and will not allow `-02` to be selected directly, but instead, appears to only be accessible through a side-by-side diff.

😭

drakkan:
This link https://filezilla-project.org/specs/draft-ietf-secsh-filexfer-02.txt is still valid

puellanivis:
A good work around, but there are also no section anchors. :( So, some links will require users to manually locate to the sections.

PR #545, opened 2023-03-27
refactor sshfx encoding, fix link rot, go fmt

puellanivis:
I’ve poked around with the wire formatting code, and switched from a “check error every time” to “accumulate an error” which allows for more direct and concise unmarshals in general.

Also, I found the ietf draft link rot #544 and updated each of the ones that points to a version earlier than `-13`. (The `-13` links are still valid.)

Also, we’ve accumulated some `go fmt` changes that show up in go 1.20, so I threw those in as well.

### DIFF
--- a/internal/encoding/ssh/filexfer/openssh/posix-rename.go
+++ b/internal/encoding/ssh/filexfer/openssh/posix-rename.go
@@ -4,38 +4,38 @@ import (
 	sshfx "github.com/pkg/sftp/internal/encoding/ssh/filexfer"
 )
 
-const extensionPosixRename = "posix-rename@openssh.com"
+const extensionPOSIXRename = "posix-rename@openssh.com"
 
-// RegisterExtensionPosixRename registers the "posix-rename@openssh.com" extended packet with the encoding/ssh/filexfer package.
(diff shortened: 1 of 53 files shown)


internal/encoding/ssh/filexfer/openssh/posix-rename.go
```diff
@@ -4,54 +4,54 @@ import (
 	sshfx "github.com/pkg/sftp/internal/encoding/ssh/filexfer"
 )
 
-const extensionPosixRename = "posix-rename@openssh.com"
+const extensionPOSIXRename = "posix-rename@openssh.com"
```
puellanivis, line 7: `POSIX` is an all-caps abbreviation that should be all-caps in variable names. (Are you glad we had this all in `internal` so we could change this bad early draft choice? :trollface: )

drakkan (approved):
I took a quick look and it looks ok. I don't have time for a full review, sorry. In general I may reduce my involvement in open source in the coming months.

puellanivis:
It’s ok. I’ve had to scale back from time to time as well. Especially, when a job comes in that doesn’t have a synergy with working on the project.

PR #546, opened 2023-04-12
Add ability for server to receive file permissions for Open and Mkdir requests

capnspacehook:
With this change, SFTP clients that support sending file attributes in SFTP `open` and `mkdir` requests can now specify the permissions of files and directories when they are created. To avoid breaking backwards compatibility and to keep this PR as small as possible, the SFTP client is not changed so support this.


server.go
```diff
@@ -19,6 +19,9 @@ import (
 const (
 	// SftpServerWorkerCount defines the number of workers for the SFTP server
 	SftpServerWorkerCount = 8
+
+	defaultFileMode = 0o644
+	defaultDirMode  = 0o755
```
puellanivis, line 24: These constants should be grouped individually in their own `const` so that a block-level documentation can cover them both.

Prefer to also have them be typed constants, so that we can simply use `mode := defaultDirMode` .

server.go
```diff
@@ -218,8 +221,15 @@ func handlePacket(s *Server, p orderedRequest) error {
 			rpkt = statusFromError(p.ID, err)
 		}
 	case *sshFxpMkdirPacket:
-		// TODO FIXME: ignore flags field
-		err := os.Mkdir(s.toLocalPath(p.Path), 0o755)
+		var mode os.FileMode = defaultDirMode
+		if p.Attrs != nil {
+			attrs, _ := unmarshalFileStat(p.Flags, p.Attrs.([]byte))
```
puellanivis, line 226: Either `p.Attrs` should be typed `[]byte` or we should properly handle the case where it is not `[]byte`.

It should be sufficient to use:
```
switch pattrs := p.Attrs.(type) {
case []byte:
  …
case *FileStat:
  … // since this would be a fast toFileMode(pattrs.Mode) might as well support it.
}
```

(I can’t speak to the legacy code doing differently. If I were redoing the FXP_SET_STAT packets, I’d do it quite a bit differently… I mean, I already did it’s in `internal/encoding/ssh/filexfer`…)

puellanivis (commented):
I’m a bit warry at providing asymmetric capabilities. This code might allow the server to support this, but the client still wouldn’t be able to support this. As well, I’m unsure how this would impact the request-server-side of the implementation.

But it could still have some utility, so I’m not hostile to the idea of patching this in.

puellanivis:
This was superseded by #574

Issue #468, opened 2021-09-07
Bug: wrong error message returned by Client.Remove


awaken:
Hi,
Great job! I really appreciate your effort for maintaining this library!
I mainly use your library inside a proprietary Big Data project, that is running in production since 2 years ago, using Linux VMs in the cloud, and moves many TBs of data per day.

`func (c *Client) Remove(path string) error` can return an error for many reasons, of course.
In most cases, the library just works fine.

If `Remove` is called for removing just a file (and not a directory), in some cases I experienced that `Remove` can return an error containing this message: "Not a directory".
In my opinion, it is a bug.

Looking at the `Remove` implementation, I see that, in case `err.Code` is equal to `sshFxFailure` or if there is a permission error, the returned error is the one got from `RemoveDirectory`.
I think if `Remove` is called for removing a file, it should never return an error with a message like "Not a directory".

I would suggest this fixed implementation:

```
func (c *Client) Remove(path string) error {
	err := c.removeFile(path)
	// some servers, *cough* osx *cough*, return EPERM, not ENODIR.
	// serv-u returns ssh_FX_FILE_IS_A_DIRECTORY
	// EPERM is converted to os.ErrPermission so it is not a StatusError
	if err2, ok := err.(*StatusError); ok {
		switch err2.Code {
		case sshFxFailure:
			if c.RemoveDirectory(path) == nil { return nil }
			return err
		case sshFxFileIsADirectory:
			return c.RemoveDirectory(path)
		}
	}
	if os.IsPermission(err) {
		if c.RemoveDirectory(path) == nil { return nil }
	}
	return err
}
```

Thanks for your attention.

puellanivis:
Hm… yeah, that does seem like a confusing error to be returning. Looking into the Go source code for `os.Remove` https://cs.opensource.google/go/go/+/refs/tags/go1.17:src/os/file_unix.go;l=314-322

🤔 Hm… curious. Not sure if your proposed change would actually fix things for all cases, or just for this one case. I’ll be on a plane flying back to Germany in 24-hours, so I’ll ponder on things and try and get something going when I get back.

Alternatively, I’m supposed to be working on reimplementing https://github.com/pkg/sftp/pull/432 (there were so many merge conflicts rebasing the PR, I’ve moved on to reimplementing) and this seems like something that would be good to address in that larger refactor, rather than giving too much priority into a confusing error message. 🤔

awaken:
Hi,
that's a server from one of our clients.
[...]

PR #548, opened 2023-04-26
Export RemoveFile method

narcisobenigno:
This tries to address partially similar to issue #468 in our case.

In the same way we have the `RemoveDirectory` method as an exported method, we put the decision in the hands of the developer of which one to choose.

What do you think?


puellanivis:
I’m kind of averse to this. We probably shouldn’t have exposed `RemoveDirectory` in the first place. As it seems it was exported as a least-change work-around for an issue people were seeing, but a more robust solution would have been to properly address the out-of-spec situation.

In the linked issue, I covered what the correct solution should be, I just haven’t had time to implement it.

puellanivis:
Essentially obsoleted by #618

PR #550, opened 2023-05-13
Implemented Delete All Resources Function

tanishqsinghal7:
DeleteAllResources delete files recursively in the directory and Recursively delete subdirectories.
An error will be returned if no file or directory with the specified path exists

### DIFF
--- a/client.go
+++ b/client.go
@@ -924,19 +924,19 @@ func (c *Client) MkdirAll(path string) error {
 	return nil
 }
 
-// DeleteAllResources delete files recursively in the directory and Recursively delete subdirectories.
+// RemoveAll delete files recursively in the directory and Recursively delete subdirectories.
 // An error will be returned if no file or directory with the specified path exists
-func (c *Client) DeleteAllResources(filePath string) error {
+func (c *Client) RemoveAll(path string) error {
 
 	// Get the file/directory information
-	fileInfo, err := c.Stat(filePath)
+	fi, err := c.Stat(path)
 	if err != nil {
 		return err
 	}
 
-	if fileInfo.IsDir() {
+	if fi.IsDir() {
 		// Delete files recursively in the directory
-		files, err := c.ReadDir(filePath)
+		files, err := c.ReadDir(path)
 		if err != nil {
 			return err
 		}
@@ -944,13 +944,13 @@ func (c *Client) DeleteAllResources(filePath string) error {
 		for _, file := range files {
 			if file.IsDir() {
 				// Recursively delete subdirectories
-				err = c.DeleteAllResources(filePath + "/" + file.Name())
+				err = c.RemoveAll(path + "/" + file.Name())
 				if err != nil {
 					return err
 				}
 			} else {
 				// Delete individual files
-				err = c.Remove(filePath + "/" + file.Name())
+				err = c.Remove(path + "/" + file.Name())
 				if err != nil {
 					return err
 				}
@@ -958,15 +958,15 @@ func (c *Client) DeleteAllResources(filePath string) error {
 		}
 
 		// Delete the empty directory
-		err = c.RemoveDirectory(filePath)
+		err = c.RemoveDirectory(path)
 		if err != nil {
-			return err
+			return c.RemoveDirectory(path)
 		}
 	} else {
 		// Delete individual files
-		err = c.Remove(filePath)
+		err = c.Remove(path)
 		if err != nil {
-			return err
+			return c.Remove(path)
 		}
 	}
 
--- a/client.go
+++ b/client.go
@@ -924,6 +924,55 @@ func (c *Client) MkdirAll(path string) error {
 	return nil
 }
 
+// DeleteAllResources delete files recursively in the directory and Recursively delete subdirectories.
+// An error will be returned if no file or directory with the specified path exists
+func (c *Client) DeleteAllResources(filePath string) error {
+
+	// Get the file/directory information
+	fileInfo, err := c.Stat(filePath)
+	if err != nil {
+		return err
+	}
+
+	if fileInfo.IsDir() {
+		// Delete files recursively in the directory
+		files, err := c.ReadDir(filePath)
+		if err != nil {
+			return err
+		}
+
+		for _, file := range files {
+			if file.IsDir() {
+				// Recursively delete subdirectories
+				err = c.DeleteAllResources(filePath + "/" + file.Name())
+				if err != nil {
+					return err
+				}
+			} else {
+				// Delete individual files
+				err = c.Remove(filePath + "/" + file.Name())
+				if err != nil {
+					return err
+				}
+			}
+		}
+
+		// Delete the empty directory
+		err = c.RemoveDirectory(filePath)
+		if err != nil {
+			return err
+		}
+	} else {
+		// Delete individual files
+		err = c.Remove(filePath)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
 // File represents a remote file.
 type File struct {
 	c      *Client
--- a/client.go
+++ b/client.go
@@ -957,12 +957,9 @@ func (c *Client) RemoveAll(path string) error {
 			}
 		}
 
-		// Delete the empty directory
-		return c.Remove(path)
-
 	}
 
-	return nil
+	return c.Remove(path)
 
 }
 
--- a/client.go
+++ b/client.go
@@ -958,12 +958,11 @@ func (c *Client) RemoveAll(path string) error {
 		}
 
 		// Delete the empty directory
-		return c.RemoveDirectory(path)
-
-	} else {
-		// Delete individual files
 		return c.Remove(path)
+
 	}
+
+	return nil
 
 }
 
--- a/client.go
+++ b/client.go
@@ -958,19 +958,13 @@ func (c *Client) RemoveAll(path string) error {
 		}
 
 		// Delete the empty directory
-		err = c.RemoveDirectory(path)
-		if err != nil {
-			return c.RemoveDirectory(path)
-		}
+		return c.RemoveDirectory(path)
+
 	} else {
 		// Delete individual files
-		err = c.Remove(path)
-		if err != nil {
-			return c.Remove(path)
-		}
+		return c.Remove(path)
 	}
 
-	return nil
 }
 
 // File represents a remote file.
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -924,6 +924,55 @@ func (c *Client) MkdirAll(path string) error {
 	return nil
 }
 
+// DeleteAllResources delete files recursively in the directory and Recursively delete subdirectories.
+// An error will be returned if no file or directory with the specified path exists
+func (c *Client) DeleteAllResources(filePath string) error {
```
puellanivis, line 929: Name of function and godoc should be modeled on https://pkg.go.dev/os@go1.20.4#RemoveAll to remain as consistent as possible with the `os` package.

The argument should also parallel the use in the other `Remove` functions. This means the signature should be:
```
func (c *Client) RemoveAll(path string) error
```

client.go
```diff
@@ -924,6 +924,55 @@ func (c *Client) MkdirAll(path string) error {
...
+					return err
+				}
+			}
+		}
+
+		// Delete the empty directory
+		err = c.RemoveDirectory(filePath)
+		if err != nil {
+			return err
+		}
```
puellanivis, line 964: This code would be more clear as simply `return c.RemoveDirectory(path)`

client.go
```diff
@@ -924,6 +924,55 @@ func (c *Client) MkdirAll(path string) error {
...
+	} else {
+		// Delete individual files
+		err = c.Remove(filePath)
+		if err != nil {
+			return err
+		}
```
puellanivis, line 970: Same as above, this would be more clear as simply `return c.Remove(path)`.

Since the `if` block will end in a `return`, this block does not need to be wrapped in an `else` block.

Once this isn’t wrapped in an `else` block, the `return c.RemoveDirectory(path)` is behaviorally the same as this new ending: `return c.Remove(path)`, that is, we could just fall out of the `if` block and let the `Remove` take care of removing both the file and the directory.

So, the code should look more like:
```
if fi.IsFile() { // `fi` is a well-known abbreviation for a `FileInfo` variable.
	files, err := c.ReadDir(path)
	…
	for _, file := range files {
		if file.IsDir() {
			…
		} else {
			…
		}
	}
}
return c.Remove(path)
```

puellanivis:
Oh, additionally, new tests should be added to test this function.

client_integration_test.go
```diff
@@ -651,6 +651,50 @@ func TestClientRemove(t *testing.T) {
...
+	f, err := ioutil.TempFile(tempDir, "sftptest-removeAll*.txt")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer f.Close()
+
+	d, err := ioutil.TempDir(tempDir, "sftptest-removeAll1")
+	if err != nil {
+		t.Fatal(err)
+	}
```
puellanivis, line 676: This test would accidentally pass even if it did not have recursion, since as I noted in the code review comments, `Client.Remove()` would remove an empty directory.

Please build a directory tree that has at least a depth of 3. So `/tmp/sftptest-removeAll/foo/bar/baz`

client_integration_test.go
```diff
@@ -651,6 +651,50 @@ func TestClientRemove(t *testing.T) {
...
+	defer os.RemoveAll(d)
```
puellanivis, line 677: Is this necessary, when it is a child of `tempDir` which means it should be deleted in line 664?

client_integration_test.go
```diff
@@ -651,6 +651,50 @@ func TestClientRemove(t *testing.T) {
...
+		t.Fatal(err)
+	}
+	defer os.RemoveAll(tempDir)
+
+	// Create file and directory within the temporary directory
+	f, err := ioutil.TempFile(tempDir, "sftptest-removeAll*.txt")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer f.Close()
```
puellanivis, line 671: If you’re just creating the file, and not writing anything to it, then you do not need to `defer` this. I think we also have helper functions available to write out temporary files?

client.go
```diff
@@ -924,6 +924,55 @@ func (c *Client) MkdirAll(path string) error {
...
+					return err
+				}
+			}
+		}
+
+		// Delete the empty directory
+		err = c.RemoveDirectory(path)
+		if err != nil {
+			return c.RemoveDirectory(path)
+		}
```
puellanivis, line 964: You have misunderstood my previous comments. I was not saying to replace only the `return err` with `return c.RemoveDirectory(path)`. I was saying you can replace _the whole block_.

Any code of the pattern:
```
err := doThing(thing)
if err != nil {
	return err
}
return nil
```
can be replaced by the much simpler `return doThing(thing)`.

If `err != nil` it returns `err`, and if `err == nil` it returns `nil`, which is equivalent to returning `err`. So, both code paths are equivalent to `return err`, and so there is no need to split the code paths in the first place.

Then, once the code is `err := doThing(thing); return err` there’s no longer any need to store the return in a temporary variable.

client.go
```diff
@@ -924,6 +924,55 @@ func (c *Client) MkdirAll(path string) error {
...
+	} else {
+		// Delete individual files
+		err = c.Remove(path)
+		if err != nil {
+			return c.Remove(path)
+		}
+	}
+
+	return nil
```
puellanivis, line 973: Previous comment has not been addressed. This whole block should be replaced by `return c.Remove(path)`

client.go
puellanivis: We are returning the `error` being returned by `RemoveDirectory`, there’s just no point in assigning it into a temporary variable, since both code paths are returning that error with no further logic performed on it, we can just return it directly.

client_integration_test.go
puellanivis: > Remove removes the specified file _or directory_. An error will be returned if no file or directory with the specified path exists, or if the specified directory is not empty. [emphasis added]

https://pkg.go.dev/github.com/pkg/sftp#Client.Remove

tanishqsinghal7: Yes, I have red this documentation but it is not working as expected atleast in my case

client_integration_test.go
```diff
@@ -651,6 +651,69 @@ func TestClientRemove(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("Failed to create file: %v", err)
+	}
+	err = ioutil.WriteFile(file2, []byte("File 2"), 0644)
+	if err != nil {
+		t.Fatalf("Failed to create file: %v", err)
+	}
+	err = ioutil.WriteFile(file3, []byte("File 3"), 0644)
+	if err != nil {
+		t.Fatalf("Failed to create file: %v", err)
```
puellanivis, line 690: [fyi] `t.Fatalf("…: %v", err)` is equivalent to `t.Fatal("…:", err)`

client_integration_test.go
puellanivis: That’s really weird. Let’s file an `Issue` and investigate that. That should be the proper functioning.

client_integration_test.go
tanishqsinghal7: Okay!

puellanivis: No, this PR still has value. It does exist in `os.RemoveAll`.

But `os.Remove´ and `Client.Remove` only work on _empty_ directories. Unlike `os.RemoveAll`. If `Client.Remove` was not working on empty directories, then that’s a problem.

tanishqsinghal7: Okay! I just rechecked again and client.Remove() is deleting empty directory. Its weird earlier I was getting error but when I tested it today it deleted empty directory. (I cleared cookie and then retry and it worked now)

puellanivis: It might have had something in it when you tried before? Especially like a hidden file or something? 🤷‍♀️

client.go
```diff
@@ -924,6 +924,49 @@ func (c *Client) MkdirAll(path string) error {
...
+			}
+		}
+
+		// Delete the empty directory
+		return c.RemoveDirectory(path)
+
+	} else {
+		// Delete individual files
+		return c.Remove(path)
+	}
```
puellanivis, line 966: As a matter of style in Go, if an `if` block ends in a return, then you shouldn’t use an `else` block, since the only way execution could possibly reach here is if the `if` block were never executed.

client_integration_test.go
tanishqsinghal7: No I just had multiple files for testing, all the files deleted but empty directory was not deleted earlier. But today when I test it again and debug it, first all files got deleted and then empty directory.

client.go
tanishqsinghal7: Right, Got it!

tanishqsinghal7: Hey @puellanivis, any other thoughts, I am modifying this. So that we can close on this?

puellanivis: Well, once you move the `return c.Remove(path)` to outside of the `else`, the `return c.RemoveDirectory(path)` will be equivalent to the `return c.Remove(path)`. So, it could just be removed.

Or, in the alternative, since we’re here in the same package, you could switch to the unexported `return c.removeFile(path)`.

tanishqsinghal7: Please check

client.go
```diff
@@ -924,6 +924,48 @@ func (c *Client) MkdirAll(path string) error {
...
+				// Delete individual files
+				err = c.Remove(path + "/" + file.Name())
+				if err != nil {
+					return err
+				}
+			}
+		}
+
+		// Delete the empty directory
+		return c.Remove(path)
```
puellanivis, line 961: You’ve accidentally introduced a bug, where this code will no longer delete a file if pointed at a file, rather than a directory.

Let’s add a Test case for this.

The `return c.Remove(path)` should be where you have the `return nil` below.

tanishqsinghal7: I am deleting file above using for condition, this will only be called to delete empty directory.
I also tested it by creating multiple files and directories and I am not getting any error. Also it is passing current Test case.
So I am bit confused

tanishqsinghal7: I am bit confused here, file will be deleted by checking in for condition.

Also I have tested with several files and directory and it working fine and also passing through current test case.

puellanivis: `os.RemoveAll` _can_ be called on a file, and so can this. It should remove the file.

tanishqsinghal7:
Hey! Thanks for your guidance throughout this process. Really appreciate it
and got to learn many things from you.
Please let know if we can collaborate on some other things and would really
like to contribute under your guidance.

Thanks
Tanishq Singhal

PR #553, opened 2023-07-13
Marshal extended data if file info supports it

peterverraedt:
If the file information returned by ListAt supports the FileInfoExtendedData interface, use it to retrieve extended data and pass it back to the client. In similar fashion, we add the FileInfoUidGid interface to allow for implementations to return uid and gid data, even if the os lacks support in syscall.Stat_t.

### DIFF
--- a/attrs.go
+++ b/attrs.go
@@ -69,6 +69,20 @@ func fileInfoFromStat(stat *FileStat, name string) os.FileInfo {
 	}
 }
 
+// FileInfoUidGid extends os.FileInfo and adds callbacks for Uid and Gid retrieval,
+// as an alternative to *syscall.Stat_t objects on unix systems.
+type FileInfoUidGid interface {
+	os.FileInfo
+	Uid() uint32
+	Gid() uint32
+}
+
+// FileInfoUidGid extends os.FileInfo and adds a callbacks for extended data retrieval.
+type FileInfoExtendedData interface {
+	os.FileInfo
+	Extended() []StatExtended
+}
+
 func fileStatFromInfo(fi os.FileInfo) (uint32, *FileStat) {
 	mtime := fi.ModTime().Unix()
 	atime := mtime
@@ -85,6 +99,23 @@ func fileStatFromInfo(fi os.FileInfo) (uint32, *FileStat) {
 
 	// os specific file stat decoding
 	fileStatFromInfoOs(fi, &flags, fileStat)
+
+	// The call above will include the sshFileXferAttrUIDGID in case
+	// the os.FileInfo can be casted to *syscall.Stat_t on unix.
+	// If fi implements FileInfoUidGid, retrieve Uid, Gid from it instead.
+	if fiExt, ok := fi.(FileInfoUidGid); ok {
+		flags |= sshFileXferAttrUIDGID
+		fileStat.UID = fiExt.Uid()
+		fileStat.GID = fiExt.Gid()
+	}
+
+	// if fi implements FileInfoExtendedData, retrieve extended data from it
+	if fiExt, ok := fi.(FileInfoExtendedData); ok {
+		fileStat.Extended = fiExt.Extended()
+		if len(fileStat.Extended) > 0 {
+			flags |= sshFileXferAttrExtended
+		}
+	}
 
 	return flags, fileStat
 }
--- a/ls_formatting.go
+++ b/ls_formatting.go
@@ -60,6 +60,13 @@ func runLs(idLookup NameLookupFileLister, dirent os.FileInfo) string {
 		uid = lsFormatID(sys.UID)
 		gid = lsFormatID(sys.GID)
 	default:
+		if fiExt, ok := dirent.(FileInfoUidGid); ok {
+			uid = lsFormatID(fiExt.Uid())
+			gid = lsFormatID(fiExt.Gid())
+
+			break
+		}
+
 		numLinks, uid, gid = lsLinksUIDGID(dirent)
 	}
 
--- a/request-interfaces.go
+++ b/request-interfaces.go
@@ -75,7 +75,7 @@ type StatVFSFileCmder interface {
 // Note in cases of an error, the error text will be sent to the client.
 // Called for Methods: List, Stat, Readlink
 //
-// Since Filelist returns an os.FileInfo, this can make it non-ideal for impelmenting Readlink.
+// Since Filelist returns an os.FileInfo, this can make it non-ideal for implementing Readlink.
 // This is because the Name receiver method defined by that interface defines that it should only return the base name.
 // However, Readlink is required to be capable of returning essentially any arbitrary valid path relative or absolute.
 // In order to implement this more expressive requirement, implement [ReadlinkFileLister] which will then be used instead.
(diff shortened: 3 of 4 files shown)


attrs.go
```diff
@@ -86,5 +99,18 @@ func fileStatFromInfo(fi os.FileInfo) (uint32, *FileStat) {
...
+	if fiExt, ok := fi.(FileInfoUidGid); ok {
+		flags |= sshFileXferAttrUIDGID
+		fileStat.UID = fiExt.Uid()
+		fileStat.GID = fiExt.Gid()
+	}
+
+	// if fi implements FileInfoExtendedData, retrieve extended data from it
+	if fiExt, ok := fi.(FileInfoExtendedData); ok {
+		flags |= sshFileXferAttrExtended
+		fileStat.Extended = fiExt.Extended()
```
puellanivis, line 112: Consider that if `len(fileStat.Extended) == 0` we have a better encoding by not asserting the `sshFileXferAttrExtended` flag in the first place?

attrs.go
```diff
@@ -69,6 +69,19 @@ func fileInfoFromStat(stat *FileStat, name string) os.FileInfo {
 	}
 }
 
+// FileInfoUidGid extends os.FileInfo and adds callbacks for Uid and Gid retrieval.
```
puellanivis, line 72: Hm, I’m not sure what the value is of exporting this interface here. Yes, it’s used immediately below, but it’s not anywhere where a caller would make use of it to understand where it is relevant?

puellanivis (commented):
I’m certainly not against this. I think you basically have the right interface going, just trying to think about the best way to fit it in.

attrs.go
peterverraedt: True, this should obviously be checked.

attrs.go
peterverraedt: I've added a comment to the code to specify where it can be required. I'm using the RequestServer code to implement an sftp server that looks like a unix filesystem but is actually stored in a kind of object store. There is no reason to have *syscall.Stat_t around for just including uid/gid support.

I guess it could be an unexported interface too, of course.

puellanivis: Yeah, I think our long-term plan should be that users can just return a `sshfx.Attributes` or `sshfx.NameEntry` more or less directly, meaning we can avoid needing to do too many work arounds to allow access to the full feature set.

But this is still a bit of a ways off, waiting for v2.

attrs.go
```diff
@@ -86,5 +100,27 @@ func fileStatFromInfo(fi os.FileInfo) (uint32, *FileStat) {
...
+	// setting uids and gids while running on non-unix systems or
+	// without using *syscall.Stat_t underneath on unix, we offer
+	// an alternative way to support this through the FileInfoUidGid
+	// interface.
+	// If fi implements FileInfoUidGid, retrieve Uid, Gid from it.
+	if fiExt, ok := fi.(FileInfoUidGid); ok {
+		flags |= sshFileXferAttrUIDGID
+		fileStat.UID = fiExt.Uid()
+		fileStat.GID = fiExt.Gid()
+	}
```
puellanivis, line 115: Consider, you might want to look at https://github.com/pkg/sftp/blob/master/ls_formatting.go#L55-L64 and how the `runLs` is going to work with this as well.

It is a known edge-case that some clients _are_ parsing that field for data, even though they were explicitly told not to.

peterverraedt: I didn't came across this `runLs` formatting yet, but of course it should be adjusted as well.

ls_formatting.go
```diff
@@ -60,6 +60,13 @@ func runLs(idLookup NameLookupFileLister, dirent os.FileInfo) string {
 		uid = lsFormatID(sys.UID)
 		gid = lsFormatID(sys.GID)
 	default:
+		if fiExt, ok := dirent.(FileInfoUidGid); ok {
```
puellanivis, line 63: This is already in a type switch. 😉  You can just use:
```
case FileInfoUidGid:
```

peterverraedt: The type switch is on `dirent.Sys()` and not on `dirent`. That is usually the same of course.

The `.Sys()` function on a `fs.FileInfo` is something that does not really mix well with with interface assertions, I prefer to only use it if the fs.FileInfo originates from a "real filesystem operation".

puellanivis: 🤦‍♀️ But of course!

attrs.go
```diff
@@ -86,5 +100,27 @@ func fileStatFromInfo(fi os.FileInfo) (uint32, *FileStat) {
 	// os specific file stat decoding
 	fileStatFromInfoOs(fi, &flags, fileStat)
 
+	// The call above will include the sshFileXferAttrUIDGID in case
+	// the os.FileInfo can be casted to *syscall.Stat_t on unix.
+	// If implementations of RequestServers Handlers want to implement
+	// setting uids and gids while running on non-unix systems or
+	// without using *syscall.Stat_t underneath on unix, we offer
+	// an alternative way to support this through the FileInfoUidGid
+	// interface.
```
puellanivis, line 109: So, I think my only remaining criticism, as small as it is, is that I was less worried about us internally tracking the proper usage of these extension `FileInfo` interfaces. That’s no problem, we maintain the code, and can “just get” how this works.

My concern was for importing code, and users of our package to know how and when the FileInfo extension interfaces should be used and what they will do. This means the documentation probably has to go into that for [`ListerAt`](https://pkg.go.dev/github.com/pkg/sftp#ListerAt) or [`FileLister`](https://pkg.go.dev/github.com/pkg/sftp#FileLister) right?

peterverraedt: Right. I'll try to move the documentation of this.

request-interfaces.go
```diff
@@ -136,6 +136,13 @@ type NameLookupFileLister interface {
 // error if at end of list. This is testable by comparing how many you
 // copied to how many could be copied (eg. n < len(ls) below).
 // The copy() builtin is best for the copying.
+// ListerAt provides entries that satisfy the os.FileInfo interface.
+// Uid and gid information will on unix systems be retrieved from
+// [os.FileInfo.Sys] if this function returns a [syscall.Stat_t].
+// Alternatively, if the entry implements [FileInfoUidGid], it will be
+// used for uid and gid information.
+// If the entry implements [FileInfoExtendedData], extended attributes will
+// also be returned to the client.
```
puellanivis, line 145: One last nitpick: paragraphs. It seems it was not well paragraphed already, but we can take this opportunity to clean up the text.

Also, if we could use semantic line breaks: https://sembr.org/ over arbitrarily wrapping at around 80 cols.

Broadly:
```
// InterfaceName does an important thing presented "as a headline".
//
// Important details about the InterfaceName that are tricky or notable in the implementation,
// but not so broadly applicable that they need to be in the headline.
//
// If a Special Case applies,
// then we note the details about that special case.
// It should start with the context of when the condition applies,
// so readers can skip the whole paragraph if the condition is irrelevant to them.
//
// If more Special Cases apply,
// then they should keep following on new paragraphs.
// Putting each Special Case in its own paragraph means readers can quickly skip to the next paragraph.
//
// If a set of special conditions exist,
// and each is fairly straight-forward enough to just list them, go ahead and do that directly:
//
//	A function domain = a special cases
//	A type name: and it’s short simple situation
//
// Note: Notes should show up at the bottom,
// so that readers can skim up from the end for quick access to gotchas, and other minutiae.
```

puellanivis (approved):
Perfection. :) Thanks for the help, and your patience.

PR #559, opened 2023-10-03
Add missing errors to request-errors.go

martin-viggiano:
Added errors according to [section 9.1 of the SSH File Transfer Protocol RFC](https://datatracker.ietf.org/doc/html/draft-ietf-secsh-filexfer-13#section-9.1).

### DIFF
--- a/request-errors.go
+++ b/request-errors.go
@@ -6,15 +6,38 @@ type fxerr uint32
 // direct control of the errors being sent vs. letting the library work them
 // out from the standard os/io errors.
 const (
-	ErrSSHFxOk               = fxerr(sshFxOk)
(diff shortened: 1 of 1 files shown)


drakkan:
Thank you, but we implement [protocol version 3](https://datatracker.ietf.org/doc/html/draft-ietf-secsh-filexfer-02#section-7), these errors are from protocol version 6

puellanivis:
In v2, we’ll have these errors covered in the `encoding/ssh/filexfer` package https://github.com/pkg/sftp/blob/dev-v2/encoding/ssh/filexfer/fx.go

But as mentioned, we target protocol version 3, so we don’t want to merge these generally.

PR #565, opened 2023-11-10
Cancelable ReadDirContext

ungerik:
Added method `Client.ReadDirContext` to be able to cancel long-running dir listings.

The passed context can cancel the operation returning all entries listed up to the cancellation.

### DIFF
--- a/client.go
+++ b/client.go
@@ -333,7 +333,7 @@ func (c *Client) ReadDir(p string) ([]os.FileInfo, error) {
 // The passed context can be used to cancel the operation
 // returning all entries listed up to the cancellation.
 func (c *Client) ReadDirContext(ctx context.Context, p string) ([]os.FileInfo, error) {
-	handle, err := c.opendir(p)
+	handle, err := c.opendir(ctx, p)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


client.go
```diff
@@ -321,17 +322,28 @@ func (c *Client) Walk(root string) *fs.Walker {
...
 	defer c.close(handle) // this has to defer earlier than the lock below
-	var attrs []os.FileInfo
+	var entries []os.FileInfo
 	var done = false
 	for !done {
+		if err = ctx.Err(); err != nil {
+			return entries, err
+		}
 		id := c.nextID()
 		typ, data, err1 := c.sendPacket(nil, &sshFxpReaddirPacket{
```
puellanivis, line 348: This creates internally a channel, that is used to waiting on for a response from the server. https://github.com/pkg/sftp/blob/master/conn.go#L131

Instead we would want to use `dispatchRequest` and then we can:
```
select {
case <-ctx.Done():
  return entries, err
case result := <-ch:
  typ, data, err1 = s.typ, s.data, s.err
}
```

Otherwise, we’re going to be stuck until the server responds with at least one `sshFxpName` packet.

ungerik: Right, I have added the context to `opendir` and `sendPacket`

puellanivis:
Hm… probably not the change I would have made… but probably a change for the better… 🤔

ungerik:
It has become a reflex to add ctx everywhere. So far it never turned out that there was too much ctx in any code base, if there were problems then because of not enough ctx ;-)

ungerik:
But I can rewrite the PR if you want...

puellanivis:
No, there is no need to rewrite, I don’t think the plumbed context will cause any performance issues.

PR #567, opened 2023-12-12
fix: add support for attrs in `sshFxpOpenPacket` for `Server`

mafredri:
Sending permissions (chmod) and atime/mtime is not uncommon for sftp clients via the `sshFxpOpenPacket`, however, flags/attrs seems to have been explicitly ignored even though they're part of the RFC (https://filezilla-project.org/specs/draft-ietf-secsh-filexfer-02.txt).

This leads to issues like https://github.com/mutagen-io/mutagen/issues/459, where the client is relying on the attrs being applied.

This only adds support for the feature in the old `Server` implementation, which is what I used. Further work is most likely required to support it in the `RequestServer` as well (however, I have no experience with that part of the code base). I believe no backwards-incompatible changes were introduced since all we're doing is parsing more of the packet.

A bug was also fixed in `sshFxpSetstatPacket` and `sshFxpFsetstatPacket` where attrs were parsed in the wrong order.

### DIFF
--- a/packet.go
+++ b/packet.go
@@ -681,12 +681,13 @@ type sshFxpOpenPacket struct {
 	ID     uint32
 	Path   string
 	Pflags uint32
-	Flags  uint32 // ignored
+	Flags  uint32
+	Attrs  interface{}
 }
 
 func (p *sshFxpOpenPacket) id() uint32 { return p.ID }
 
-func (p *sshFxpOpenPacket) MarshalBinary() ([]byte, error) {
+func (p *sshFxpOpenPacket) marshalPacket() ([]byte, []byte, error) {
 	l := 4 + 1 + 4 + // uint32(length) + byte(type) + uint32(id)
 		4 + len(p.Path) +
 		4 + 4
@@ -698,7 +699,14 @@ func (p *sshFxpOpenPacket) MarshalBinary() ([]byte, error) {
 	b = marshalUint32(b, p.Pflags)
 	b = marshalUint32(b, p.Flags)
 
-	return b, nil
+	payload := marshal(nil, p.Attrs)
+
+	return b, payload, nil
+}
+
+func (p *sshFxpOpenPacket) MarshalBinary() ([]byte, error) {
+	header, payload, err := p.marshalPacket()
+	return append(header, payload...), err
 }
 
 func (p *sshFxpOpenPacket) UnmarshalBinary(b []byte) error {
@@ -709,9 +717,10 @@ func (p *sshFxpOpenPacket) UnmarshalBinary(b []byte) error {
 		return err
 	} else if p.Pflags, b, err = unmarshalUint32Safe(b); err != nil {
 		return err
-	} else if p.Flags, _, err = unmarshalUint32Safe(b); err != nil {
+	} else if p.Flags, b, err = unmarshalUint32Safe(b); err != nil {
 		return err
 	}
+	p.Attrs = b
 	return nil
 }
 
--- a/server.go
+++ b/server.go
@@ -467,7 +467,7 @@ func (p *sshFxpOpenPacket) respond(svr *Server) responsePacket {
 	// being created. Otherwise, the permissions are ignored.
 	if b, ok := p.Attrs.([]byte); ok && (p.Flags&sshFileXferAttrPermissions) != 0 {
 		fs, _ := unmarshalFileStat(p.Flags, b)
-		mode = os.FileMode(fs.Mode)
+		mode = toFileMode(fs.Mode)
 	}
 
 	name := svr.toLocalPath(p.Path)
@@ -536,7 +536,7 @@ func (p *sshFxpSetstatPacket) respond(svr *Server) responsePacket {
 	if (p.Flags & sshFileXferAttrPermissions) != 0 {
 		var mode uint32
 		if mode, b, err = unmarshalUint32Safe(b); err == nil {
-			err = os.Chmod(p.Path, os.FileMode(mode))
+			err = os.Chmod(p.Path, toFileMode(mode))
 		}
 	}
 	if err != nil {
@@ -591,7 +591,7 @@ func (p *sshFxpFsetstatPacket) respond(svr *Server) responsePacket {
 	if (p.Flags & sshFileXferAttrPermissions) != 0 {
 		var mode uint32
 		if mode, b, err = unmarshalUint32Safe(b); err == nil {
-			err = f.Chmod(os.FileMode(mode))
+			err = f.Chmod(toFileMode(mode))
 		}
 	}
 	if err != nil {
--- a/server.go
+++ b/server.go
@@ -462,22 +462,29 @@ func (p *sshFxpOpenPacket) respond(svr *Server) responsePacket {
 		osFlags |= os.O_EXCL
 	}
 
-	f, err := os.OpenFile(svr.toLocalPath(p.Path), osFlags, 0o644)
-	if err != nil {
-		return statusFromError(p.ID, err)
-	}
+	name := svr.toLocalPath(p.Path)
+	mode := os.FileMode(0o644)
+
+	var applyAttrs []func(*os.File) error
 
 	// Both `sshFileXferAttrPermissions` and `sshFileXferAttrACmodTime` are set
 	// by e.g. `sftp`. Just in case, we handle all other cases as well.
-	if b, ok := p.Attrs.([]byte); ok {
+	// Only apply for newly created files.
+	var err error
+	useAttrs := true
+	if _, err := os.Stat(name); err == nil {
+		useAttrs = false
+	}
+	if b, ok := p.Attrs.([]byte); useAttrs && ok {
 		if (p.Flags & sshFileXferAttrSize) != 0 {
 			var size uint64
 			if size, b, err = unmarshalUint64Safe(b); err == nil {
-				err = f.Truncate(int64(size))
+				applyAttrs = append(applyAttrs, func(f *os.File) error {
+					return f.Truncate(int64(size))
+				})
 			}
 		}
 		if err != nil {
-			_ = f.Close()
 			return statusFromError(p.ID, err)
 		}
 		if (p.Flags & sshFileXferAttrUIDGID) != 0 {
@@ -486,21 +493,22 @@ func (p *sshFxpOpenPacket) respond(svr *Server) responsePacket {
 			if uid, b, err = unmarshalUint32Safe(b); err != nil {
 			} else if gid, b, err = unmarshalUint32Safe(b); err != nil {
 			} else {
-				err = f.Chown(int(uid), int(gid))
+				applyAttrs = append(applyAttrs, func(f *os.File) error {
+					return f.Chown(int(uid), int(gid))
+				})
 			}
 		}
 		if err != nil {
-			_ = f.Close()
 			return statusFromError(p.ID, err)
 		}
 		if (p.Flags & sshFileXferAttrPermissions) != 0 {
-			var mode uint32
-			if mode, b, err = unmarshalUint32Safe(b); err == nil {
-				err = f.Chmod(os.FileMode(mode))
+			var attrMode uint32
+			if attrMode, b, err = unmarshalUint32Safe(b); err == nil {
+				// Optionally, we could apply umask here.
+				mode = os.FileMode(attrMode)
 			}
 		}
 		if err != nil {
-			_ = f.Close()
 			return statusFromError(p.ID, err)
 		}
 		if (p.Flags & sshFileXferAttrACmodTime) != 0 {
@@ -511,10 +519,23 @@ func (p *sshFxpOpenPacket) respond(svr *Server) responsePacket {
 			} else {
 				atimeT := time.Unix(int64(atime), 0)
 				mtimeT := time.Unix(int64(mtime), 0)
-				err = os.Chtimes(f.Name(), atimeT, mtimeT)
+				applyAttrs = append(applyAttrs, func(f *os.File) error {
+					return os.Chtimes(f.Name(), atimeT, mtimeT)
+				})
 			}
 		}
 		if err != nil {
+			return statusFromError(p.ID, err)
+		}
+	}
+
+	f, err := os.OpenFile(name, osFlags, mode)
+	if err != nil {
+		return statusFromError(p.ID, err)
+	}
+
+	for _, applyAttr := range applyAttrs {
+		if err := applyAttr(f); err != nil {
 			_ = f.Close()
 			return statusFromError(p.ID, err)
 		}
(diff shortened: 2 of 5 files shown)


server.go
```diff
@@ -462,11 +462,85 @@ func (p *sshFxpOpenPacket) respond(svr *Server) responsePacket {
...
+	// Both `sshFileXferAttrPermissions` and `sshFileXferAttrACmodTime` are set
+	// by e.g. `sftp`. Just in case, we handle all other cases as well.
+	// Only apply for newly created files.
+	var err error
+	useAttrs := true
+	if _, err := os.Stat(name); err == nil {
+		useAttrs = false
+	}
+	if b, ok := p.Attrs.([]byte); useAttrs && ok {
+		if (p.Flags & sshFileXferAttrSize) != 0 {
```
puellanivis, line 479: We’re just going to entirely repeat `unmarshalFileStat`? This probably isn’t a good idea.

server.go
```diff
@@ -462,11 +462,85 @@ func (p *sshFxpOpenPacket) respond(svr *Server) responsePacket {
...
+	mode := os.FileMode(0o644)
+
+	var applyAttrs []func(*os.File) error
+
+	// Both `sshFileXferAttrPermissions` and `sshFileXferAttrACmodTime` are set
+	// by e.g. `sftp`. Just in case, we handle all other cases as well.
+	// Only apply for newly created files.
+	var err error
+	useAttrs := true
+	if _, err := os.Stat(name); err == nil {
```
puellanivis, line 475: Race-condition. The file could come to exist in the time between this test, and the application of these attributes below.

server.go
```diff
@@ -462,11 +462,85 @@ func (p *sshFxpOpenPacket) respond(svr *Server) responsePacket {
...
+		useAttrs = false
+	}
+	if b, ok := p.Attrs.([]byte); useAttrs && ok {
+		if (p.Flags & sshFileXferAttrSize) != 0 {
+			var size uint64
+			if size, b, err = unmarshalUint64Safe(b); err == nil {
+				applyAttrs = append(applyAttrs, func(f *os.File) error {
+					return f.Truncate(int64(size))
```
puellanivis, line 483: I guess there is technically a possibility of overflow here. Converting from a `uint64` to an `int64`.

server.go
```diff
@@ -462,11 +462,85 @@ func (p *sshFxpOpenPacket) respond(svr *Server) responsePacket {
...
+			}
+		}
+		if err != nil {
+			return statusFromError(p.ID, err)
+		}
+		if (p.Flags & sshFileXferAttrPermissions) != 0 {
+			var attrMode uint32
+			if attrMode, b, err = unmarshalUint32Safe(b); err == nil {
+				// Optionally, we could apply umask here.
+				mode = os.FileMode(attrMode)
```
puellanivis, line 508: This conversion is not actually valid. These two types disagree on quite a number of bits.

server.go
```diff
@@ -509,33 +583,41 @@ func (p *sshFxpSetstatPacket) respond(svr *Server) responsePacket {
...
+			err = os.Chown(p.Path, int(uid), int(gid))
+		}
+	}
+	if err != nil {
+		return statusFromError(p.ID, err)
+	}
 	if (p.Flags & sshFileXferAttrPermissions) != 0 {
 		var mode uint32
 		if mode, b, err = unmarshalUint32Safe(b); err == nil {
 			err = os.Chmod(p.Path, os.FileMode(mode))
```
puellanivis, line 604: Oh, lovely. We’re conflating these two incompatible types in existing code…

server.go
```diff
@@ -556,33 +638,41 @@ func (p *sshFxpFsetstatPacket) respond(svr *Server) responsePacket {
...
+			err = f.Chown(int(uid), int(gid))
+		}
+	}
+	if err != nil {
+		return statusFromError(p.ID, err)
+	}
 	if (p.Flags & sshFileXferAttrPermissions) != 0 {
 		var mode uint32
 		if mode, b, err = unmarshalUint32Safe(b); err == nil {
 			err = f.Chmod(os.FileMode(mode))
```
puellanivis, line 659: And again, we conflate it again… 🤦‍♀️

puellanivis (commented):
I don’t think this has been done yet, just because it’s so particularly clunky.

And we kind of set out a plan for a new v2 in the `internal/encoding/ssh/filexfer` package that was supposed to clean up so much of this.

packet.go
```diff
@@ -698,7 +699,14 @@ func (p *sshFxpOpenPacket) MarshalBinary() ([]byte, error) {
 	b = marshalUint32(b, p.Pflags)
 	b = marshalUint32(b, p.Flags)
 
-	return b, nil
+	payload := marshal(nil, p.Attrs)
```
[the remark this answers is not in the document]
mafredri: I kinda made the assumption that the old code is OK and didn't review it at all. This is the same thing that's being done for the stat packets as well. Is there a need to handle open differently?

server.go
mafredri: I saw this method, but it wasn't being used by any of the `Server` implementations, so I omitted it. I can switch to using that method if you prefer.

server.go
mafredri: Good catch. However, is there any way to solve it? I don't think `OpenFile` gives us this information either. However, `mode` is no-op in `OpenFile` if the file exists. So the only thing I can come up with is that we don't support anything else than mode change, which is a bit disappointing. I've seen time attributes being sent as well.

server.go
mafredri: This follows the existing implementation elsewhere. I think this is mostly theoretical though, either buggy clients or files exceeding 9000 PB?

server.go
mafredri: Isn't the same conversion being done in the newer implementations as well? https://github.com/pkg/sftp/blob/5bdc2b0e679d0d8336dc63f8b790108667ea4280/internal/encoding/ssh/filexfer/attrs.go#L186

server.go
puellanivis: I can only really think of opening with the equivalent of `SSH_FXF_EXCL` , which will fail if the file already exists, which would then take us to a short-circuit path that simply opens without `SSH_FXF_CREAT` and doesn’t apply the settings.

Alternatively, check exactly what what openssh does, and copy that.

packet.go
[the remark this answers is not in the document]
puellanivis: I would probably use:
```
switch attr := p.Attr.(type) {
case []byte:
  return b, payload, nil
default:
  b = marshal(b, p.Attrs) // in this case, we may as well use the append feature of the marshaling.
  return b
}
```

server.go
puellanivis: Yeah, probably safe, just noticed it. (inb4 new 8 EiB of filespace ought to be enough for anybody.)

server.go
puellanivis: We’ll want to wrap all of these through either [toFileMode](https://github.com/pkg/sftp/blob/master/stat_posix.go#L53) or [fromFileMode](https://github.com/pkg/sftp/blob/master/stat_posix.go#L87)

server.go
mafredri: It seems OpenSSH only uses `mode` in open, so I'll take that route.

server.go
mafredri: While looking at the OpenSSH implementation, I saw that OpenSSH also has no handling and uses the mode as-is.

packet.go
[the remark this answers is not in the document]
mafredri: Changing this seems like a can of worms I would like to avoid opening. 😅 There seem to exist code-paths for when an unexported `marshalPacket` exists (compared to `MarshalBinary` only, as your suggestion would require), and changing this would mean we diverge from the impl. in `sshFxpSetstatPacket` and `sshFxpFsetstatPacket`.

server.go
mafredri: This was removed anyway (only the old existing code for stat packets is left).

mafredri:
@puellanivis thanks for the thorough and timely reviews/feedback, as always. It's great to come back to this project every once in a while and see that it's still kicking. ☺️

I believe I've amended all of the feedback (at least what I felt confident/comfortable changing), including a little bit of cleanup of the old code.

server.go
puellanivis: OK, so, this is kind of weird to explain. Yes, OpenSSH uses the mode directly, because the SFTP standard is that the field conforms to the “POSIX filemode bit assignment”.

However, _Go_ doesn’t use the POSIX filemode bit assignment. It uses a subtly different one, which is why we have the `toFileMode` and `fromFileMode` functions in the first place.

packet.go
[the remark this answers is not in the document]
puellanivis: 🤔 OK, sounds good. We can likely push it off. When (if? 😭) we get the V2 API working, the `encoding/ssh/filexfer` covers this without relying on `any` types.

server.go
mafredri: Oh, TIL. That's good context, thanks for sharing!

mafredri:
@puellanivis hey, just wanted to check if there are any blockers/actions left for this PR? Cheers, and happy new year!

server.go
```diff
@@ -545,45 +518,36 @@ func (p *sshFxpFsetstatPacket) respond(svr *Server) responsePacket {
 		return statusFromError(p.ID, EBADF)
 	}
 
-	// additional unmarshalling is required for each possibility here
+	debug("fsetstat name \"%s\"", f.Name())
 	b := p.Attrs.([]byte)
-	var err error
+	return statusFromError(p.ID, applyAttrsToFile(f.Name(), p.Flags, b))
```
puellanivis, line 523: Oh dear. :( I’m unsure of how this subtle change from operating on a handle to operating on the filename would work. We’ve run into issues before where performing operations on what should be an open handle but done to the filename, or on a filename but through an open/close block.

I know `Chtimes` only exists for the filename, and not on a handle… but we could at least attempt to isolate this operation.

Also know that various systems will open a file, then delete it. This is a really weird corner case, but it seems like everyone runs into this weird pattern eventually and get tripped up assuming that a file will always be accessible from its filename while it is currently open.

server_test.go
```diff
@@ -208,6 +208,51 @@ func TestOpenStatRace(t *testing.T) {
...
 }
 
+func TestOpenWithPermissions(t *testing.T) {
+	client, server := clientServerPair(t)
+	defer client.Close()
+	defer server.Close()
+
+	tmppath := path.Join(os.TempDir(), "open_permissions")
+	pflags := flags(os.O_RDWR | os.O_CREATE | os.O_TRUNC)
+	ch := make(chan result, 2)
```
puellanivis, line 218: This channel should never have more than one in flight. (The design of `clientConn` guarantees an “at most once” delivery on a channel given to `dispatchRequest`, and after the first `dispatchRequest` this goroutine blocks waiting for that response.)

server_test.go
```diff
@@ -208,6 +208,51 @@ func TestOpenStatRace(t *testing.T) {
...
+
+	// New files should have their permissions set.
+	client.dispatchRequest(ch, &sshFxpOpenPacket{
+		ID:     id1,
+		Path:   tmppath,
+		Pflags: pflags,
+		Flags:  sshFileXferAttrPermissions,
+		Attrs:  []byte{0x0, 0x0, 0x1, 0xe5}, // 0o745 -- a slightly strange permission to test.
+	})
+	<-ch
```
puellanivis, line 230: Should probably test the error returned from the `dispatchRequest` channel return?

Why not actually just avoid channel tedium here entirely and use the slightly higher level `sendPacket` which will do the receive and extract the `error` from the `result` into a return value for you?

server_test.go
```diff
@@ -208,6 +208,51 @@ func TestOpenStatRace(t *testing.T) {
...
+	stat, err := os.Stat(tmppath)
+	assert.NoError(t, err)
```
puellanivis, line 232: This needs to be a `require.NoError` as if `err != nil` then `stat == nil` and the following line will panic.

server_test.go
```diff
@@ -208,6 +208,51 @@ func TestOpenStatRace(t *testing.T) {
 	checkServerAllocator(t, server)
 }
 
+func TestOpenWithPermissions(t *testing.T) {
+	client, server := clientServerPair(t)
+	defer client.Close()
+	defer server.Close()
+
+	tmppath := path.Join(os.TempDir(), "open_permissions")
+	pflags := flags(os.O_RDWR | os.O_CREATE | os.O_TRUNC)
```
puellanivis, line 217: You need to not use the `os.O_*` flags. SFTP does not use the same flags.

For SFTP:
```
        #define SSH_FXF_READ            0x00000001
        #define SSH_FXF_WRITE           0x00000002
        #define SSH_FXF_CREAT           0x00000008
        #define SSH_FXF_TRUNC           0x00000010
```

for `os` imported from `syscall`:
```
	O_RDWR                           = 0x2
	O_CREAT                          = 0x40
	O_TRUNC                          = 0x200
```

If the test was working with these flags, then 😰 uh-oh.

server_test.go
```diff
@@ -208,6 +208,51 @@ func TestOpenStatRace(t *testing.T) {
...
+		Attrs:  []byte{0x0, 0x0, 0x1, 0xed}, // 0o755
+	})
+	<-ch
+	stat, err = os.Stat(tmppath)
+	assert.NoError(t, err, "mode should not have changed")
+	if !assert.Equal(t, os.FileMode(0o745), stat.Mode()&os.ModePerm) {
+		t.Logf("stat.Mode() = %v", stat.Mode())
+	}
+
+	os.Remove(tmppath)
```
puellanivis, line 252: Please put this in a `defer` so that it gets performed even if a `t.Fail` is called.

server_test.go
```diff
@@ -208,6 +208,51 @@ func TestOpenStatRace(t *testing.T) {
...
+	client.dispatchRequest(ch, &sshFxpOpenPacket{
+		ID:     id2,
+		Path:   tmppath,
+		Pflags: pflags,
+		Flags:  sshFileXferAttrPermissions,
+		Attrs:  []byte{0x0, 0x0, 0x1, 0xed}, // 0o755
+	})
+	<-ch
+	stat, err = os.Stat(tmppath)
+	assert.NoError(t, err, "mode should not have changed")
```
puellanivis, line 247: This test failure message doesn’t make sense. We don’t know if the mode has changed or not yet. We just know that the `Stat` failed, and this could point to a deeper issue than just “mode should not have changed“.

Fix: just don’t put a log message here. Simply there being an error message popping up here is clue enough that an unexpected error occurred, and deeper review is necessary of what is happening.

puellanivis (commented):
Sorry for the delay in re-reviewing. Since the code itself is pretty much fine with only a simple concern, this time I’ve taken a closer look at the test.

server_test.go
puellanivis: I am wrong here. `flags` does the conversion. (Excellent name there.)

puellanivis:
Closing in deference to #574 which is pulling in the changes.

Thanks for your contribution though!

PR #571, opened 2024-01-06
server.go: "/" for windows

### LINKED ISSUE #569: In Windows, `Server` should allow a pseudo-directory for `/` which enumerates available drives
When I use `sshfs host:/ mnt`, the current working directory gets mounted to `mnt`.
Is it possible to let the sshfs server behave the way a cygwin compiled openssh server does, e.g. that it lists `/c:`, `/d:`, ... in the root directory?

powellnorma:
fix #569


server_windows.go
```diff
@@ -37,3 +44,76 @@ func (s *Server) toLocalPath(p string) string {
...
+}
+func (s *dummyDriveStat) Sys() any {
+	return nil
+}
+
+type WinRoot struct {
+	dummyFile
+}
+
+func (f *WinRoot) Readdir(int) ([]os.FileInfo, error) {
```
puellanivis, line 102: Be aware of this implementation detail from `os.File.Readdir`:
> If n > 0, Readdir returns at most n FileInfo structures. In this case, if Readdir returns an empty slice, it will return a non-nil error explaining why. At the end of a directory, the error is io.EOF.

This behavior has to be respected in our implementation.

As is, we would end up in an infinite loop of `Readdir` because we will never return the eventually expected `io.EOF`.

server_windows.go
```diff
@@ -37,3 +44,76 @@ func (s *Server) toLocalPath(p string) string {
...
+	}
+	infos := []os.FileInfo{}
+	for _, drive := range drives {
+		infos = append(infos, &dummyDriveStat{drive})
+	}
+	return infos, nil
+}
+
+func openFileLike(path string, flag int, mode fs.FileMode) (FileLike, error) {
+	if path == "/" {
```
puellanivis, line 115: This is a bug, which means we would never actually be returning `&WinRoot{}`.

`Server.toLocalPath` has already mangled `path` with at minimum `filepath.FromSlash` meaning this path should be now in local Windows format.

I think I would also like some sort of consideration that `\` is a valid path in Windows meaning “top-level of current drive” vs what we’re trying to do here, which is list the logical drives available.

We could maybe internally use here the filepath `\\.\` which is the start of the Win32 Device Namespace, and otherwise invalid.

Also, nothing here so far is going to force `/` to be followed by a volume indicator. So `/c:/windows` and `/windows` is likely going to point to the same place.

If we use the `\\.\` device namespace prefix, then at least we could distinguish `/c:/windows` as `\\.\c:\windows` from `/windows` as `\windows`. Otherwise, we’re going to need some hefty work here making sure we don’t just make the same weird situation where some paths are valid and can be opened despite being unlistable.

puellanivis:
I suppose I want to add that changing from this `/` is the root of the current drive, to `/` is the drive-space root is a breaking change of behavior, and should properly be gated behind a https://pkg.go.dev/github.com/pkg/sftp#ServerOption

server_windows.go
```diff
@@ -37,3 +44,76 @@ func (s *Server) toLocalPath(p string) string {
...
+var kernel32, _ = syscall.LoadLibrary("kernel32.dll")
+var getLogicalDrivesHandle, _ = syscall.GetProcAddress(kernel32, "GetLogicalDrives")
+
+func bitsToDrives(bitmap uint32) []string {
+	var drive rune = 'A'
+	var drives []string
+
+	for bitmap != 0 {
+		if bitmap&1 == 1 {
+			drives = append(drives, string(drive))
```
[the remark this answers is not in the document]
powellnorma: I guess it also should be `var drive rune = 'a'` instead of `var drive rune = 'A'` then

server.go
```diff
@@ -21,6 +22,51 @@ const (
...
+	return os.ErrPermission
+}
+func (f *dummyFile) Chown(uid, gid int) error {
+	return os.ErrPermission
+}
+func (f *dummyFile) Close() error {
+	return os.ErrPermission
+}
+
+var _ = dummyFile{} // ignore unused
```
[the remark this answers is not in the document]
powellnorma: Hm, I think this however makes it super easy to extend the current codebase for more "virtual paths", for which only "a small subset of file operations" is implemented. One can now just "inherit" `dummyFile` (via promoted fields) and "overwrite" the methods one needs to implement.

> This also solves the problem you’ve created here where dummyFile.Name() always returns "dummyFile" which is a valid filename

Or one just removes "Name()" from the list of functions a default implementation is provided? What filename do you suggest?

server_windows.go
```diff
@@ -37,3 +44,76 @@ func (s *Server) toLocalPath(p string) string {
...
+		return nil, fmt.Errorf("GetLogicalDrives: %w", callErr)
+	} else {
+		drives := bitsToDrives(uint32(ret))
+		return drives, nil
+	}
+}
+
+type dummyDriveStat struct {
+	name string
+}
```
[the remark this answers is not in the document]
powellnorma: > Not sure that as currently designed, this needs to be a struct. One can make a receiver on a shape of string just as well.

What do you mean with "a receiver on a shape of string"?

Good idea to use the underlying os.Stat of the driver-path!

server_windows.go
powellnorma: Good point!

Hmm.. I see that currently, in server.go:
```
func (p *sshFxpReaddirPacket) respond(svr *Server) responsePacket {
	[..]
	dirents, err := f.Readdir(128)
	if err != nil {
		return statusFromError(p.ID, err)
	}
```

Does that mean the server is currently unable to list directories with > 128 entries?

powellnorma: I think we are supposed to only return `EOF` once we returned an empty slice - So we'd have to implement keeping track of subsequent calls in some way.

server_windows.go
powellnorma: So we'd also have to adjust `toLocalPath` to always use `\\.\`? And then automatically paths like `/windows` would resolve to `\\.\windows\` -> not valid anymore (which is wanted)?

powellnorma:
> I suppose I want to add that changing from this / is the root of the current drive, to / is the drive-space root is a breaking change of behavior, and should properly be gated behind a https://pkg.go.dev/github.com/pkg/sftp#ServerOption

Ok, but I'd think it would be good to make it opt-out by default - Because I'd guess most people have not yet thought about this edge case, and would not expect the old behavior. What do you think?

server_windows.go
powellnorma: > So we'd have to implement keeping track of subsequent calls in some way.

Ok have done that - What about the 128 limit in `server.go`, though?

server_windows.go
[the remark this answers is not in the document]
puellanivis: Yes

server.go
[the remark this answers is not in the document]
puellanivis: [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)

server_windows.go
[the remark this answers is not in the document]
puellanivis:
```
type driveInfo string

func (d driveInfo) Name() string {
  return string(d)
}
```

server_windows.go
puellanivis: > Does that mean the server is currently unable to list directories with > 128 entries?

No, it only returns 128 entries at a time. The client is to request a `SSH_FXP_READDIR` repeatedly until it receives an `EOF`: https://github.com/pkg/sftp/blob/v1.13.6/client.go#L363-L366

puellanivis:
The correct default state of the option should be the old behavior. Because it’s the current behavior, and [we don’t know if people are depending on this current behavior.](https://www.hyrumslaw.com/)

server_windows.go
```diff
@@ -1,8 +1,16 @@
+//go:build go1.18
```
puellanivis, line 1: I am unsure that this go build tag is necessary. `x/sys/windows` should be compatible with at least some older versions?

server_windows.go
```diff
@@ -37,3 +45,80 @@ func (s *Server) toLocalPath(p string) string {
...
+	name string
+}
+
+func (i *driveInfo) Name() string {
+	return i.name // since the Name() returned from a os.Stat("C:\\") is "\\"
+}
+
+type winRoot struct {
+	dummyFile
+	doneDirs int
```
puellanivis, line 83: We could also just store a `drives []string` that is filled in at `openFile` and then pop off the return values as they are read:

```
drives = f.drives
if len(drives) > n {
  drives = drives[:n]
}
f.drives = f.drives[:len(drives)]
```

Then we don’t have to make a syscall every time, and potentially receiving different results because say, a new drive attached or not.

server_windows.go
```diff
@@ -37,3 +45,80 @@ func (s *Server) toLocalPath(p string) string {
...
+	}
+
+	if f.doneDirs >= len(drives) {
+		return nil, io.EOF
+	}
+	drives = drives[f.doneDirs:]
+
+	var infos []os.FileInfo
+	for i, drive := range drives {
+		if i >= n {
```
puellanivis, line 99: Consider now this detail from `os.File.Readdir`:
> If n <= 0, Readdir returns all the FileInfo from the directory in a single slice. In this case, if Readdir succeeds (reads all the way to the end of the directory), it returns the slice and a nil error. If it encounters an error before the end of the directory, Readdir returns the FileInfo read until that point and a non-nil error.

server.go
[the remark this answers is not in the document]
powellnorma: How do you know? It doesn't add much complexity IMO. It allows one to do interesting stuff, for example one could create a /proc like "filesystem" that is completely implemented in the sftp server! I surely could see some interesting use cases for that, which I might even add later (in a private fork). I think one should consider the pros and cons, I don't see many cons in this case? But maybe I am missing something?

server_windows.go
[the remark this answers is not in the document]
powellnorma: Promoted `fs.FileInfo` is used now in `driveInfo`

server_windows.go
powellnorma: It was needed for using `any` - Now we should be ok without it

server_windows.go
powellnorma: I think it should be `f.drives = f.drives[:len(drives)]`?

powellnorma: fixed!

powellnorma:
How to add the an additional server option? Can you please help out?

server.go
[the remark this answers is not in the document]
puellanivis: If you want to add it to your private fork, I cannot stop you. But we don’t need this code in this package.

server_windows.go
```diff
@@ -37,3 +43,86 @@ func (s *Server) toLocalPath(p string) string {
...
+func (f *winRoot) Readdir(n int) ([]os.FileInfo, error) {
+	drives := f.drives
+	if n > 0 {
+		if len(drives) > n {
+			drives = drives[:n]
+		}
+		f.drives = f.drives[len(drives):]
+		if len(drives) == 0 {
+			return nil, io.EOF
+		}
```
puellanivis, line 103: This code needs to be outside of the `if n > 0` block.

Otherwise, a `Readdir(-1)` would just continuously return all the drives forever.

server_windows.go
```diff
@@ -37,3 +43,86 @@ func (s *Server) toLocalPath(p string) string {
...
+	}
+
+	var infos []os.FileInfo
+	for _, drive := range drives {
+		fi, err := os.Stat(drive)
+		if err != nil {
+			return nil, err
+		}
```
puellanivis, line 111: 🤔 actually `os.Stat("C:")` does not reference the same directory as `os.Stat("C:\")`. If you’re on the C: drive, then the first is equivalent to `os.Stat(".")`, while the later is equivalent to `os.Stat("\"`)`.

powellnorma: good catch!

server.go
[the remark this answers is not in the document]
powellnorma: Mixing `dummyFile` with `winRoot` IMO mixes two separate issues.. The first is to implement a "dummy" that realizes the `file` interface, the second is to implement the `winRoot` functionality.

When one looks at `winRoot`, one imediatly sees the gist of what it is doing. It is not obscurred by all the "dummy implementations", because those are contained in `dummyFile`. 

> But we don’t need this code in this package.

Who is "we"? Do you know who is going to fork it, and potentially want to add functionality? I don't think so.

> YAGNI

I think this principle is useful for deciding if one wants to invest the time into implementing something.. But in this case it is already implemented, and IMO is cleaner (for the reason described above).

If this gets into "No you have to change X and Y and Z because I like it more" without mentioning real reasons behind that, then I must tell you that I don't have time for that. It is similar to the "No you need to use `permission denied`, because I say so, and disregarding the point I gave on why I think `Operations cannot be performed on this handle` is more matching for a situation where not the privilege level of the user is the problem, but that the operation makes no sense for the referenced resource.

This is your repository so you are free to do what you want. But don't expect me to change this while not giving any clear reasons, thanks 🙏

drakkan:
@powellnorma  I'm not familiar with the old Server implementation and so I can't help you much here and I don't even have time to do so. 

The feature you propose makes sense and would be useful for Windows users (although you can use RequestServer as an alternative) but please:

1) remember we are volunteers, no one pays us for this work or to help you so don't request for step by step help and carefully test your code before sending a PR
2) respect the guideline of the project owners, @puellanivis is a really experienced developer, she is very thorough in reviewing code and test cases and I really appreciated when she helped me improve my PRs in the past

Thanks for understanding and for trying to contribute to this package

puellanivis (changes requested):

server.go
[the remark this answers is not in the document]
powellnorma: Thank you for the thorough explanation.

> Clients are far more likely to be operating under the assumption that os.ErrPermission is a “do not retry this operation, but handle is still open for further use”, while SSH_FX_FAILURE is so generic that a client cannot really recover from the failure programmatically at all, as it is completely undefined and unknown whether any further operations are possible at that point.

Okay, if that is the case I guess `SSH_FX_PERMISSION_DENIED` is practically the more appropriate choice.
However, I want to note, that it feels like `os.ErrPermission` then takes the role of "generic error" that SSH_FX_FAILURE perhaps should have.
And I still think that `os.ErrPermission` is / can be misleading to the end-user who reads this error message.

An error like `ENOSYS` would be ideal I guess. Hmm, `SSH_FX_OP_UNSUPPORTED` comes to mind.

> > indicates that an attempt was made to perform an operation which is not supported for the server
> [..]
> The server supports this operation. It just cannot perform this operation on that specific handle.

Well, one could say that the server does not implement e.g. copying a directory to `/` (the virtual root). But it _could_ be implemented to e.g. mean "create a new drive and copy the folder contents onto there".

So I think it would be more similar to when e.g. Linux gives `ENOSYS` (or `EOPNOTSUPP`) for deleting a file for a specific filesystem. It's not that linux is incapable of deleting files, but just the current implementation of the filesystem is (thus rendering linux unable to do that for that filesystem).

> The original form of Server has used a map[string]*os.File since it was first introduced in a merged in September 2015. We have only now 9 years later, seen good cause to convert this into an interface

If you think about it, this argument goes both ways. You originally said that implementing this would be a lot of work, and you don't see anyone having time for that. In effect, a useful feature thus might have never been implemented, because the code looks too "tedious" to be refactored/extended in such a way.

I see your point of keeping implementations "clean and lean". However, I'd say if it is not "a lot of complexity" that gets added, the advantages (easier to be extended in the future) outweighs the disadvantage.
And in this case, I personally even feel like separating `dummyFile` and `winRoot` is cleaner, even if one only uses `dummyFile` once.

But I guess it's a matter of taste, so I have changed the code according to your wishes.

> So, I must be patient, but insistent. If you interpret this professional courtesy as “passive aggressive”, then I’m sorry.

No, what I perceived as "passive aggressive" was when you just repeated things you already said, which I had given an answer to, but you (thereby) ignored what I said.
It felt as if you were not willing to listen to the points I made, and instead just were half-annoyed that I dared to question some of your statements.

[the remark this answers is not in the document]
puellanivis: > Hmm, SSH_FX_OP_UNSUPPORTED comes to mind.

I had already addressed the appropriateness of this status value 5 days ago: https://github.com/pkg/sftp/issues/569#issuecomment-1879738231

> So I think it would be more similar to when e.g. Linux gives ENOSYS (or EOPNOTSUPP) for deleting a file for a specific filesystem.

Except that `ENOSYS` is not listed at all for `unlink(2)` https://man7.org/linux/man-pages/man2/unlink.2.html

In fact, instead, it is noted specifically that for linux:
```
       EPERM (Linux only)
              The filesystem does not allow unlinking of files.
```

`EOPNOTSUPP` is defined for _sockets only_, (https://pubs.opengroup.org/onlinepubs/9699919799/functions/V2_chap02.html#tag_15_03)
```
[EOPNOTSUPP]
Operation not supported on socket. The type of socket (address family or protocol) does not support the requested operation. A conforming implementation may assign the same values for [EOPNOTSUPP] and [ENOTSUP].
```

You might note that it allows for it to be conflated with `ENOTSUP`. However, even when conflated, only socket operations should be returning the `EOPNOTSUPP` synonym, while other operations should still be returning `ENOTSUP`.

As for the appropriateness of `ENOTSUP` not only does my argument against `SSH_FX_OP_UNSUPPORTED` still apply, so does my argument against `ENOSYS`.

> because the code looks too "tedious" to be refactored/extended in such a way.

You’ve mistaken entirely what my anticipation of the tedium would be.

Your originally proposed method of implementing this was to modify every single point in the code where we called `getHandle` and treat a magic handle token specially. I warned you against that course of action, because this would have indeed been horrendously tedious. I directed you instead, to the least tedious implementation.

Instead, the tedium, which I anticipated, was never in constructing a pseudo-directory implementation. It was anticipating the process of ensuring everything worked properly, and that we covered all the “gotchas” that invariably result from implementing a parallel implementation—in this case to `*os.File`. Namely:
* if you call `Readdir(n)` with an `n > 0`, you have to ensure to return a most that number of entries.
* if you call `Readdir(n)` with an `n <= 0`, you have to ensure that it returns all entries.
* when you call `Readdir` a second time, you should not return any of the entries already returned.
* once `Readdir` is called and there are no remaining entries, it must return `io.EOF`, but not ever before
* the `Readdir` has to return filenames that actually reflect the usage that clients should be using them as (i.e. `c:` not just `c`)
* Checking for what `os.Stat("c:\")` actually returns, and how we need to change it in order to fit our use-case, so that we can be sure we’re returning the proper and appropriate values from all of the `fs.FileInfo` values for the individual drives
* Properly encoding the `toLocalPath` operations to ensure that we don’t allow accesses to `\Windows` but instead require a drive name

> No, what I perceived as "passive aggressive" was when you just repeated things you already said, which I had given an answer to, but you (thereby) ignored what I said.
> It felt as if you were not willing to listen to the points I made, and instead just were half-annoyed that I dared to question some of your statements.

If I have repeated myself, it is because you refused to acknowledge or listen to the point I had already made, or brought up a point we had already moved from, while trying to get me to relitigate it. (see, above, where you again drag out `SSH_FX_OP_UNSUPPORTED` after I had already written a detailed essay about why it would be inappropriate.)

Your arguments have repeatedly refused to actually listen to anything for which I was taking hours to research and compose, in order to patiently explain all this to you.

Even now, your response is not, “I understand. Thank you for taking so much time explaining these fine, minute, and subtle details, and helping to ensure that my code covers all of the gotcha corner cases involved in the standards. Let me know if this code addresses your concerns,” but instead you have answered, “I still refuse to acknowledge your arguments. I think I’m still right, but I have constructed a face-saving reason for me to comply. So, I will do it your way.”

You have mistaken “no matter how many times you repeat this argument, the standards and references continue to refute it,” for “I am ignoring your arguments.”

If I were ignoring your arguments, I would not be pulling up the POSIX documentation in order to evaluate the validity and soundness of your arguments. Meanwhile, in this single message alone, you’ve:
* repeated an already refuted `SSH_FX_OP_UNSUPPORTED`
* thrown out ENOSYS in a situation where under your very own scenario, linux is returning EPERM
* and tossed in `EOPNOTSUPP` without actually reading the POSIX standard to see that it is entirely constrained to sockets.

To me, it really does seems like you are not only not reading my explanations; you’re not even reading your own references.

----

Under advisement from @drakkan , this matter has gotten quite a bit out of hand, so we’re parking everything involved for 2 weeks, so that all parties (including me) can cool off, and we can come back to this later, when emotions are not being run raw.

Since I am both: the person with an imbalance of control in this situation, and also the one closing down the discussion; I want to be clear that I’m not doing this to silence you. You may post a reasonable response, but no further actions will be taken from my side for two weeks.†

†: With the caveat that there is one linting comment to address a typo, which will follow this.

server.go
```diff
@@ -116,6 +130,14 @@ func ReadOnly() ServerOption {
 	}
 }
 
+// configures a Server to serve a virtual '/' for windows that lists all drives
```
puellanivis, line 133: [golint]
```
server.go:113:1: comment on exported function WindowsRootEnumeratesDrives should be of the form "WindowsRootEnumeratesDrives ..."
```

server.go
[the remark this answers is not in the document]
powellnorma: I feel like if we look at the bigger picture, we both mostly have the same goal of making this software more useful. It's not useful for either of us to spend much time discussing minor details, like whether to implement something in one way or a slightly different way.

> EOPNOTSUPP is defined for sockets only

Two things:

sshfs itself uses this error code when it gets an 'SSH_FX_OP_UNSUPPORTED':
https://github.com/libfuse/sshfs/blob/sshfs-3.7.3/sshfs.c#L1704

If we look at the linux source code, we see that it is actually used a lot in filesystem (fs/*) code:
https://elixir.bootlin.com/linux/v4.0/ident/EOPNOTSUPP

> As for the appropriateness of ENOTSUP not only does my argument against SSH_FX_OP_UNSUPPORTED still apply

Your argument was that the server does generally support the operation (e.g. unlink), so it should not return this error.
However, my counter-argument was that "operation" might entail the subject (e.g. the path) that is operated on.

If we look at the docs for `SSH_FX_PERMISSION_DENIED`, they say: `The user does not have sufficient permissions to perform the operation`.
So it seems that here the term "operation" clearly entails the subject (path) that is operated on, because the user probably can e.g. unlink his own files.

I don't think you have addressed that "counter-argument" yet, but instead you say things like `repeated an already refuted SSH_FX_OP_UNSUPPORTED`.

Maybe you have just missed/overlooked that, it's not (overall) a big deal. I think it will help us to assume that the other person had good intentions. Then probably it will be easier for us to see the point the other person is making, instead of going into some kind of "defensive mode".

> Instead, the tedium, which I anticipated, was never in constructing a pseudo-directory implementation

You said: `just even adding support to use a pseudo-directory in Server would be a pain, because right now our handles are all mapping directly to an actual *os.File which—as noted above—cannot exist for this pseudo-directory.`

This clearly references the restrictiveness of the old implementation, and that (alone) would be "a pain".

And true, making sure that everything is implemented correctly is quite some work too! And I am grateful that you took the time to review it that carefully.

> but instead you have answered “I still refuse to acknowledge your arguments. I think I’m still right, but I have constructed a face-saving reason for me to comply. So, I will do it your way.”

I said:
* Thank you for the thorough explanation.
* Okay, if that is the case I guess SSH_FX_PERMISSION_DENIED is practically the more appropriate choice
* I see your point of keeping implementations "clean and lean"

Still, I have my own opinions. And while I might do X as requested because I agree with you on Y, it doesn't mean that this has to "invalidate" my previous opinion Z which may also be about X.

> To me, it really does seems like you are not only not reading my explanations; you’re not even reading your own references.

I don't think such accusations are helpful in any way.

Besides, I am not even sure what you mean by "own references". I mostly just said that I think for the enduser a specific error code could be confusing. And that another one seems more fitting IMO.

I don't think that there is a clear "right" or "wrong", which is probably why one could argue about this endlessly. But I don't think there is much value in that discussion, as both "perspectives" are somewhat valid, and in the end its IMO a minor detail of this feature.

puellanivis (approved):

drakkan:
Not sure if I'm testing this PR in the correct way but I get this

```
sftp> ls /
Can't ls: "/" not found
```

I'm testing using `examples/go-sftp-server/main.go` with this change (it would be useful to add an option for win root to the example)

```
git diff examples/go-sftp-server/main.go
diff --git a/examples/go-sftp-server/main.go b/examples/go-sftp-server/main.go
index ba902b6..30d1071 100644
--- a/examples/go-sftp-server/main.go
+++ b/examples/go-sftp-server/main.go
@@ -127,6 +127,7 @@ func main() {
                } else {
                        fmt.Fprintf(debugStream, "Read write server\n")
                }
+               serverOptions = append(serverOptions, sftp.WindowsRootEnumeratesDrives())
 
                server, err := sftp.NewServer(
                        channel,
```

The sftp cli sends a lstat packet and the server errors out [here](https://github.com/powellnorma/sftp/blob/win-root/server.go#L220)

puellanivis:
Oh wow, yeah, that’s not translating the remote path into local path. 😮

puellanivis:
OK, I think we don’t want to block the existing changes on implementing `Lstat` on the `\\?\` path.

But right now, until we have full compatibility with the openssh CLI client, we just cannot merge this. Or, we merge this, and then fix the issues ourselves… there are a few nitpicks I’d like to address as well anyways. Either way, I’d like to release a patch version covering the zos changes first.

powellnorma:
Happy new year! :)

It should work fine now

server_windows.go
```diff
@@ -33,7 +44,148 @@ func (s *Server) toLocalPath(p string) string {
...
+
+func (s *Server) openfile(path string, flag int, mode fs.FileMode) (file, error) {
+	if path == `\\.\` && s.winRoot {
+		return newWinRoot()
+	}
+	return os.OpenFile(path, flag, mode)
+}
+
+type winRootFileInfo struct {
+	fs.FileInfo
```
puellanivis, line 163: [bug] This does not seem to be either set or used? So, it should probably be removed.

server_windows.go
```diff
@@ -33,7 +44,148 @@ func (s *Server) toLocalPath(p string) string {
...
+	name string
+}
+
+func (w winRootFileInfo) Name() string       { return w.name }
+func (w winRootFileInfo) Size() int64        { return 0 }
+func (w winRootFileInfo) Mode() fs.FileMode  { return fs.ModeDir | 0555 } // read+execute for all
+func (w winRootFileInfo) ModTime() time.Time { return time.Time{} }
```
puellanivis, line 170: [nit] We probably want to return a time that has a chance of fitting into a `uint32`? (That is something around `time.Now` or the `time.Now` when we instantiate the `winRootFileInfo`?

server_windows.go
```diff
@@ -33,7 +44,148 @@ func (s *Server) toLocalPath(p string) string {
...
+			name:     drive,
+		}
+		infos = append(infos, di)
+	}
+
+	return infos, nil
+}
+
+func (f *winRoot) Stat() (os.FileInfo, error) {
+	return nil, os.ErrPermission
```
puellanivis, line 131: [nit] I think we implement `Stat` now that we have `winRootFileInfo`?

server_windows.go
```diff
@@ -33,7 +44,148 @@ func (s *Server) toLocalPath(p string) string {
...
+	return os.ErrPermission
+}
+func (f *winRoot) Chmod(mode fs.FileMode) error {
+	return os.ErrPermission
+}
+func (f *winRoot) Chown(uid, gid int) error {
+	return os.ErrPermission
+}
+func (f *winRoot) Close() error {
+	return nil
```
puellanivis, line 152: [improvement] Maybe we want to clear the internal `drives` field?

server_windows.go
```diff
@@ -33,7 +44,148 @@ func (s *Server) toLocalPath(p string) string {
...
 	}
 
 	return lp
 }
+
+func bitsToDrives(bitmap uint32) []string {
+	var drive rune = 'a'
+	var drives []string
+
+	for bitmap != 0 {
```
puellanivis, line 61: [invalid value guarding] Something I thought of, we might want also a `drive <= 'z'` … I’m pretty sure windows should never return a set bit higher than the 26th bit, but the guard probably isn’t a bad idea still.

server_windows.go
```diff
@@ -33,7 +44,148 @@ func (s *Server) toLocalPath(p string) string {
...
+
+func (w winRootFileInfo) Name() string       { return w.name }
+func (w winRootFileInfo) Size() int64        { return 0 }
+func (w winRootFileInfo) Mode() fs.FileMode  { return fs.ModeDir | 0555 } // read+execute for all
+func (w winRootFileInfo) ModTime() time.Time { return time.Time{} }
+func (w winRootFileInfo) IsDir() bool        { return true }
+func (w winRootFileInfo) Sys() interface{}   { return nil }
+
+// Create a new root FileInfo
+var rootFileInfo = winRootFileInfo{
```
puellanivis, line 175: [nit] I would suggest we implement this through pointer receivers, and use a pointer-to-struct.

There have been at times kind of wonky things when dealing with non-pointers in an interface.

puellanivis (commented):
Looking nice, just a few improvements, I think. Thanks for getting back to this.

powellnorma:
Good points, updated as per your suggestion.

puellanivis (approved):

Issue #572, opened 2024-01-11
close file double time sometimes cause copy problem wh


lindmin:
Hello,

version of go : `1.20.5`
version of lib sftp : `v1.13.6`
version of sftp server: `OpenSSH_8.9p1 Ubuntu-3ubuntu0.6`

I sometime experiment a bug when my code close 2 times the same sftp.File ref in concurrent situation

here the code to reproduce :

main.go
```
package main

import (
	"bytes"
	"fmt"
	"net"
	"os"
	"sync"

	"github.com/pkg/sftp"
	"golang.org/x/crypto/ssh"
)

func main() {
	connSSH, errSSH := ssh.Dial("tcp", net.JoinHostPort("localhost", "22"), &ssh.ClientConfig{
		HostKeyCallback: ssh.InsecureIgnoreHostKey(),
		User:            "USER",
		Auth: []ssh.AuthMethod{
			ssh.Password("PASSWORD"),
		},
	})
	if errSSH != nil {
		panic(errSSH)
	}
	client, errSftp := sftp.NewClient(connSSH, sftp.UseConcurrentReads(true), sftp.UseConcurrentWrites(true), sftp.UseFstat(true))
	if errSftp != nil {
		panic((errSftp))
	}
	if errSftp != nil {
		panic(errSftp.Error())
	}
	input := make(chan int, 1000)

	for i := 1; i < 1000; i++ {
		input <- i
	}

	wait := &sync.WaitGroup{}
	for i := 1; i <= 1000; i++ {
		wait.Add(1)
		go func(wait *sync.WaitGroup) {
			defer wait.Done()
			for {
				idx, ok := <-input
				if !ok {
					return
				}
				filename := fmt.Sprintf("/tmp/file-%d.dat", idx)
				err := copy(client, filename, "small content")
				if err != nil {
					fmt.Println(err.Error())
				}
			}
		}(wait)
	}
	close(input)
	wait.Wait()

}

func copy(client *sftp.Client, filename string, content string) error {
	file, errCreate := client.OpenFile(filename, os.O_WRONLY|os.O_CREATE|os.O_TRUNC)
	if errCreate != nil {
		return errCreate
	}
	_, errCopy := file.ReadFrom(bytes.NewBufferString(content))
	if errCopy != nil {					
               fmt.Printf("💣💣💣💣💣 COPY: %s", err.Error())
		//clean it
		client.Remove(filename)
		return errCopy
	}
	defer file.Close()
	//force double close
	file.Close()
[...]

PR #574, opened 2024-01-19
Rework client to prevent after-Close usage, and support perm at Open

puellanivis:
Two things in this:
* First, clear out the `File.handle` after a `Close` which can then be used to ensure that there is no use-after-close issue: https://github.com/pkg/sftp/issues/572
* Second, after cleaning up other `Client` code, we can now easily implement the PR changes in: https://github.com/pkg/sftp/pull/567

### DIFF
--- a/attrs.go
+++ b/attrs.go
@@ -32,10 +32,10 @@ func (fi *fileInfo) Name() string { return fi.name }
 func (fi *fileInfo) Size() int64 { return int64(fi.stat.Size) }
 
 // Mode returns file mode bits.
-func (fi *fileInfo) Mode() os.FileMode { return toFileMode(fi.stat.Mode) }
+func (fi *fileInfo) Mode() os.FileMode { return fi.stat.FileMode() }
 
 // ModTime returns the last modification time of the file.
-func (fi *fileInfo) ModTime() time.Time { return time.Unix(int64(fi.stat.Mtime), 0) }
+func (fi *fileInfo) ModTime() time.Time { return fi.stat.ModTime() }
 
 // IsDir returns true if the file is a directory.
 func (fi *fileInfo) IsDir() bool { return fi.Mode().IsDir() }
@@ -54,6 +54,21 @@ type FileStat struct {
 	UID      uint32
 	GID      uint32
 	Extended []StatExtended
+}
+
+// ModTime returns the Mtime SFTP file attribute converted to a time.Time
+func (fs *FileStat) ModTime() time.Time {
+	return time.Unix(int64(fs.Mtime), 0)
+}
+
+// AccessTime returns the Atime SFTP file attribute converted to a time.Time
+func (fs *FileStat) AccessTime() time.Time {
+	return time.Unix(int64(fs.Atime), 0)
+}
+
+// FileMode returns the Mode SFTP file attribute converted to an os.FileMode
+func (fs *FileStat) FileMode() os.FileMode {
+	return toFileMode(fs.Mode)
 }
 
 // StatExtended contains additional, extended information for a FileStat.
--- a/client.go
+++ b/client.go
@@ -1023,10 +1023,6 @@ func (f *File) Read(b []byte) (int, error) {
 	f.mu.Lock()
 	defer f.mu.Unlock()
 
-	if f.handle == "" {
-		return 0, os.ErrClosed
-	}
-
 	n, err := f.readAt(b, f.offset)
 	f.offset += int64(n)
 	return n, err
@@ -1095,14 +1091,15 @@ func (f *File) ReadAt(b []byte, off int64) (int, error) {
 	f.mu.RLock()
 	defer f.mu.RUnlock()
 
-	if f.handle == "" {
-		return 0, os.ErrClosed
-	}
-
 	return f.readAt(b, off)
 }
 
 func (f *File) readAt(b []byte, off int64) (int, error) {
+	if f.handle == "" {
+		return 0, os.ErrClosed
+	}
+	handle := f.handle // need a local copy to prevent aberrent race detection
+
 	if len(b) <= f.c.maxPacket {
 		// This should be able to be serviced with 1/2 requests.
 		// So, just do it directly.
@@ -1154,7 +1151,7 @@ func (f *File) readAt(b []byte, off int64) (int, error) {
 
 			f.c.dispatchRequest(res, &sshFxpReadPacket{
 				ID:     id,
-				Handle: f.handle,
+				Handle: handle,
 				Offset: uint64(offset),
 				Len:    uint32(chunkSize),
 			})
@@ -1302,6 +1299,7 @@ func (f *File) WriteTo(w io.Writer) (written int64, err error) {
 	if f.handle == "" {
 		return 0, os.ErrClosed
 	}
+	handle := f.handle // need a local copy to prevent aberrent race detection
 
 	if f.c.disableConcurrentReads {
 		return f.writeToSequential(w)
@@ -1387,7 +1385,7 @@ func (f *File) WriteTo(w io.Writer) (written int64, err error) {
 
 			f.c.dispatchRequest(res, &sshFxpReadPacket{
 				ID:     id,
-				Handle: f.handle,
+				Handle: handle,
 				Offset: uint64(off),
 				Len:    uint32(chunkSize),
 			})
@@ -1740,14 +1738,15 @@ func (f *File) ReadFromWithConcurrency(r io.Reader, concurrency int) (read int64
 	f.mu.Lock()
 	defer f.mu.Unlock()
 
-	if f.handle == "" {
-		return 0, os.ErrClosed
-	}
-
 	return f.readFromWithConcurrency(r, concurrency)
 }
 
 func (f *File) readFromWithConcurrency(r io.Reader, concurrency int) (read int64, err error) {
+	if f.handle == "" {
+		return 0, os.ErrClosed
+	}
+	handle := f.handle // need a local copy to prevent aberrent race detection
+
 	// Split the write into multiple maxPacket sized concurrent writes.
 	// This allows writes with a suitably large reader
 	// to transfer data at a much faster rate due to overlapping round trip times.
@@ -1792,7 +1791,7 @@ func (f *File) readFromWithConcurrency(r io.Reader, concurrency int) (read int64
 
 				f.c.dispatchRequest(res, &sshFxpWritePacket{
 					ID:     id,
-					Handle: f.handle,
+					Handle: handle,
 					Offset: uint64(off),
 					Length: uint32(n),
 					Data:   b[:n],
--- a/client.go
+++ b/client.go
@@ -438,7 +438,10 @@ func (c *Client) Lstat(p string) (os.FileInfo, error) {
 			return nil, &unexpectedIDErr{id, sid}
 		}
 		attr, _, err := unmarshalAttrs(data)
-		return fileInfoFromStat(attr, path.Base(p)), err
+		if err != nil {
+			return nil, err
+		}
+		return fileInfoFromStat(attr, path.Base(p)), nil
 	case sshFxpStatus:
 		return nil, normaliseError(unmarshalStatus(id, data))
 	default:
@@ -664,6 +667,9 @@ func (c *Client) stat(path string) (*FileStat, error) {
 			return nil, &unexpectedIDErr{id, sid}
 		}
 		attr, _, err := unmarshalAttrs(data)
+		if err != nil {
+			return nil, err
+		}
 		return attr, err
 	case sshFxpStatus:
 		return nil, normaliseError(unmarshalStatus(id, data))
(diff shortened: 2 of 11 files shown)


attrs.go
```diff
@@ -56,6 +56,21 @@ type FileStat struct {
...
+
+// AccessTime returns the Atime SFTP file attribute converted to a time.Time
+func (fs *FileStat) AccessTime() time.Time {
+	return time.Unix(int64(fs.Atime), 0)
+}
+
+// FileMode returns the Mode SFTP file attribute converted to an os.FileMode
+func (fs *FileStat) FileMode() os.FileMode {
+	return toFileMode(fs.Mode)
+}
```
puellanivis, line 72: This was located in a different file, which is a little weird, since it operates on `FileStat`.

client.go
```diff
@@ -257,7 +257,7 @@ func NewClientPipe(rd io.Reader, wr io.WriteCloser, opts ...ClientOption) (*Clie
 // read/write at the same time. For those services you will need to use
 // `client.OpenFile(os.O_WRONLY|os.O_CREATE|os.O_TRUNC)`.
 func (c *Client) Create(path string) (*File, error) {
-	return c.open(path, flags(os.O_RDWR|os.O_CREATE|os.O_TRUNC))
+	return c.open(path, toPflags(os.O_RDWR|os.O_CREATE|os.O_TRUNC))
```
puellanivis, line 260: This function is changing names, so that it’s clear that it is converting `os` flags into `pflags` flags.

client.go
```diff
@@ -510,7 +510,7 @@ func (c *Client) Symlink(oldname, newname string) error {
 	}
 }
 
-func (c *Client) setfstat(handle string, flags uint32, attrs interface{}) error {
+func (c *Client) fsetstat(handle string, flags uint32, attrs interface{}) error {
```
puellanivis, line 513: This name changes to reflect the actual packet it’s sending.

client.go
```diff
@@ -976,16 +976,26 @@ func (c *Client) RemoveAll(path string) error {
 type File struct {
 	c      *Client
 	path   string
-	handle string
 
-	mu     sync.Mutex
+	mu     sync.RWMutex
+	handle string
```
puellanivis, line 981: `handle` is now protected by mutex.

client.go
```diff
@@ -976,16 +976,26 @@ func (c *Client) RemoveAll(path string) error {
...
-	return f.c.close(f.handle)
+	f.mu.Lock()
+	defer f.mu.Unlock()
+
+	if f.handle == "" {
+		return os.ErrClosed
+	}
+
+	handle := f.handle
+	f.handle = ""
```
puellanivis, line 996: This is the meat of the fix for double closes.

Though I think we might want to change this to:
```
err := f.c.close(f.handle)
if err == nil {
	f.handle = ""
}
```

Actually… no. Looking up the code, `openssh/sftp-server.c` will release the handle no matter what the result of the `close` operation is.

So this is correct. (TODO: add a comment explaining this.)

client.go
```diff
@@ -1006,7 +1016,11 @@ func (f *File) Read(b []byte) (int, error) {
 	f.mu.Lock()
 	defer f.mu.Unlock()
 
-	n, err := f.ReadAt(b, f.offset)
+	if f.handle == "" {
+		return 0, os.ErrClosed
+	}
+
+	n, err := f.readAt(b, f.offset)
```
puellanivis, line 1023: This function has to change names, so that `ReadAt` can read-lock, while this function needs to hold the write-lock.

So, `readat` needs to be called with a lock, but it doesn’t matter which lock.

client.go
```diff
@@ -1071,6 +1085,17 @@ func (f *File) readAtSequential(b []byte, off int64) (read int, err error) {
 // the number of bytes read and an error, if any. ReadAt follows io.ReaderAt semantics,
 // so the file offset is not altered during the read.
 func (f *File) ReadAt(b []byte, off int64) (int, error) {
+	f.mu.RLock()
+	defer f.mu.RUnlock()
```
puellanivis, line 1089: Mutex is converted to a `RWMutex` so that we can distinguish between “critical only one request is ever pending at a time” from “must be allowed to run with essentially arbitrary parallelism”

client.go
```diff
@@ -1456,9 +1485,20 @@ func (f *File) WriteTo(w io.Writer) (written int64, err error) {
...
+		return nil, os.ErrClosed
+	}
+
+	return f.stat()
+}
+
 // Stat returns the FileInfo structure describing file. If there is an
 // error.
-func (f *File) Stat() (os.FileInfo, error) {
+func (f *File) stat() (os.FileInfo, error) {
```
puellanivis, line 1501: This function has to be renamed because `File.Seek` can need to call into this, while `File.Stat` can run in parallel.

client.go
```diff
@@ -1675,6 +1730,17 @@ func (f *File) WriteAt(b []byte, off int64) (written int, err error) {
...
+	defer f.mu.Unlock()
+
+	if f.handle == "" {
+		return 0, os.ErrClosed
+	}
+
+	return f.readFromWithConcurrency(r, concurrency)
+}
+
+func (f *File) readFromWithConcurrency(r io.Reader, concurrency int) (read int64, err error) {
```
puellanivis, line 1743: The original of this never grabbed the mutex, even though it modifies `f.offset` 😱

client.go
```diff
@@ -1927,20 +2001,61 @@ func (f *File) Seek(offset int64, whence int) (int64, error) {
...
+	defer f.mu.RUnlock()
+
+	if f.handle == "" {
+		return os.ErrClosed
+	}
+
+	return f.c.fsetstat(f.handle, sshFileXferAttrUIDGID, &FileStat{
+		UID: uint32(uid),
+		GID: uint32(gid),
+	})
```
puellanivis, line 2014: This started the `attr` cleanup work. Wanted to be able to put this data into a proper struct.

client.go
```diff
@@ -1927,20 +2001,61 @@ func (f *File) Seek(offset int64, whence int) (int64, error) {
...
+
+	return f.c.fsetstat(f.handle, sshFileXferAttrPermissions, toChmodPerm(mode))
+}
+
+// Truncate sets the size of the current file. Although it may be safely assumed
+// that if the size is less than its current size it will be truncated to fit,
+// the SFTP protocol does not specify what behavior the server should do when setting
+// size greater than the current size.
+// We send a SSH_FXP_FSETSTAT here since we have a file handle
+func (f *File) Truncate(size int64) error {
```
puellanivis, line 2036: This function is moved to be next to the other `fsetstat` calls.

client.go
```diff
@@ -1990,7 +2096,7 @@ func normaliseError(err error) error {
 
 // flags converts the flags passed to OpenFile into ssh flags.
 // Unsupported flags are ignored.
-func flags(f int) uint32 {
+func toPflags(f int) uint32 {
```
puellanivis, line 2099: 🎉 longer more descriptive names for functions that aren’t on the same screen.

packet.go
```diff
@@ -56,6 +56,11 @@ func marshalFileInfo(b []byte, fi os.FileInfo) []byte {
 	flags, fileStat := fileStatFromInfo(fi)
 
 	b = marshalUint32(b, flags)
+
+	return marshalFileStat(b, flags, fileStat)
+}
+
+func marshalFileStat(b []byte, flags uint32, fileStat *FileStat) []byte {
```
puellanivis, line 63: Spin this out so that we can call it elsewhere, like if an open/(f)setstat packet has a `*FileStat` as it’s `p.Attrs`.

packet.go
```diff
@@ -103,6 +107,8 @@ func marshal(b []byte, v interface{}) []byte {
 		return marshalUint64(b, v)
 	case string:
 		return marshalString(b, v)
+	case []byte:
+		return append(b, v...)
```
puellanivis, line 111: May as well shortcut this, rather than going through the `reflect` library.

packet.go
```diff
@@ -180,8 +186,6 @@ func unmarshalFileStat(flags uint32, b []byte) (*FileStat, []byte) {
 	}
 	if flags&sshFileXferAttrUIDGID == sshFileXferAttrUIDGID {
 		fs.UID, b, _ = unmarshalUint32Safe(b)
-	}
-	if flags&sshFileXferAttrUIDGID == sshFileXferAttrUIDGID {
```
puellanivis, line 184: 🤷‍♀️ not sure why these two `if`s were separated.

packet.go
```diff
@@ -709,12 +727,25 @@ func (p *sshFxpOpenPacket) UnmarshalBinary(b []byte) error {
...
+	p.Attrs = b
 	return nil
 }
 
+func (p *sshFxpOpenPacket) unmarshalFileStat(flags uint32) *FileStat {
+	switch attrs := p.Attrs.(type) {
+	case *FileStat:
+		return attrs
+	case []byte:
+		fs, _ := unmarshalFileStat(flags, attrs)
```
puellanivis, line 742: This `unmarshalFileStat` should probably return an `error`

packet.go
```diff
@@ -709,12 +727,25 @@ func (p *sshFxpOpenPacket) UnmarshalBinary(b []byte) error {
...
+		return fs
+	default:
+		panic(fmt.Sprintf("invalid type in unmarshalFileStat: %T", attrs))
```
puellanivis, line 745: “don’t panic” but also… we don’t have an `error`?

All the more reason why `unmarshalFileStat` should return an `error`.

server.go
```diff
@@ -462,7 +461,15 @@ func (p *sshFxpOpenPacket) respond(svr *Server) responsePacket {
 		osFlags |= os.O_EXCL
 	}
 
-	f, err := os.OpenFile(svr.toLocalPath(p.Path), osFlags, 0o644)
+	mode := os.FileMode(0o644)
+	// Like OpenSSH, we only handle permissions here, if the file is being created.
+	// Otherwise, the permissions are ignored.
```
puellanivis, line 466: The semantic line break here is actually between the first line before the `if`. It should be a full stop not a comma.

server.go
```diff
@@ -545,41 +541,32 @@ func (p *sshFxpFsetstatPacket) respond(svr *Server) responsePacket {
...
-		var atime uint32
-		var mtime uint32
-		if atime, b, err = unmarshalUint32Safe(b); err != nil {
-		} else if mtime, b, err = unmarshalUint32Safe(b); err != nil {
-		} else {
-			atimeT := time.Unix(int64(atime), 0)
-			mtimeT := time.Unix(int64(mtime), 0)
-			err = os.Chtimes(f.Name(), atimeT, mtimeT)
+		if err == nil {
+			err = os.Chtimes(path, fs.AccessTime(), fs.ModTime())
```
puellanivis, line 564: The only operation here that doesn’t have an “on a handle` operation.

🤔 This should probably actually be last as well. I don’t know if changing the UID/GID below would reset the the `mtime` or the `atime`, right?

server_integration_test.go
```diff
@@ -591,18 +591,25 @@ ls -l /usr/bin/
...
-					bad = true
+				if goWord != opWord {
+					switch j {
+					case 1, 2, 3, 7:
+						// words[1] as the link count for directories like proc is unstable
+						// words[2] and [3] as these are users & groups
+						// words[7] as timestamps on dirs can vary for things like /tmp
+					case 8:
+						// words[8] can either have full path or just the filename
+						bad = !strings.HasSuffix(opWord, "/" + goWord)
```
puellanivis, line 609: 🤷‍♀️ I don’t know what is up about the macOS I’m doing dev on, but `openssh` is dumping full paths, while our go server is not.

But like, this test was passing for other machines, right? So, ¿ 😕 ?

server_test.go
```diff
@@ -178,21 +178,22 @@ func TestOpenStatRace(t *testing.T) {
...
+	pflags := toPflags(os.O_RDWR | os.O_CREATE | os.O_TRUNC)
 	ch := make(chan result, 3)
 	id1 := client.nextID()
+	id2 := client.nextID()
 	client.dispatchRequest(ch, &sshFxpOpenPacket{
 		ID:     id1,
 		Path:   tmppath,
 		Pflags: pflags,
 	})
-	id2 := client.nextID()
```
puellanivis, line 189: move this, so we have the potential for a tighter race possibility.

client.go
```diff
@@ -974,18 +977,32 @@ func (c *Client) RemoveAll(path string) error {
...
+	if f.handle == "" {
+		return os.ErrClosed
+	}
+
+	// When `openssh-portable/sftp-server.c` is doing `handle_close`,
+	// it will unconditionally mark the handle as unused,
+	// so we need to also unconditionally mark this handle as invalid.
+
+	handle := f.handle
+	f.handle = ""
```
puellanivis, line 1003: This is the meat of the fix for double closes.

server_integration_test.go
```diff
@@ -591,18 +591,25 @@ ls -l /usr/bin/
...
-					bad = true
+				if goWord != opWord {
+					switch j {
+					case 1, 2, 3, 7:
+						// words[1] as the link count for directories like proc is unstable
+						// words[2] and [3] as these are users & groups
+						// words[7] as timestamps on dirs can vary for things like /tmp
+					case 8:
+						// words[8] can either have full path or just the filename
+						bad = !strings.HasSuffix(opWord, "/"+goWord)
```
puellanivis, line 609: 🤷‍♀️ I don’t know what is up about the macOS I’m doing dev on, but `openssh` is dumping full paths, while our go server is not.

But like, this test was passing for other machines, right? So, ¿ 😕 ?

client.go
```diff
@@ -1071,6 +1088,18 @@ func (f *File) readAtSequential(b []byte, off int64) (read int, err error) {
...
+	defer f.mu.RUnlock()
+
+	return f.readAt(b, off)
+}
+
+func (f *File) readAt(b []byte, off int64) (int, error) {
+	if f.handle == "" {
+		return 0, os.ErrClosed
+	}
+	handle := f.handle // need a local copy to prevent aberrent race detection
```
puellanivis, line 1101: Since we use `f.handle` in a sub-goroutine, the access is caught by the race detector as being performed “outside of” the lock that we’re holding in this primary goroutine, which triggers the race condition detector.

So, we need to, in this primary goroutine, extract the `f.handle` value into a local variable, which we can then access through closure in the sub goroutine. We’re safe to use a copy, because it cannot be updated as long as we’re holding either the read or write lock… which is why it’s an aberrent race detection. It just cannot tell that this sub goroutine’s lifetime is guaranteed to be limited by the primary goroutine.

puellanivis:
Alright, I feel this is pretty solid now. I’ll be merging it shortly.

puellanivis:
> I plan to do some testing this weekend, but I'm sure I won't find anything wrong. Sounds like a great job, thanks!

I’ll hold off until Monday then. More broad testing is better than the very narrow testing that I’ve done so far.

client.go
drakkan: ok, so this is why we use `f.mu.Lock()` in `Read/Write` and `f.mu.RLock()` in `ReadAt/WriteAt` even though they both call `f.readAt/f.writeAt`

drakkan:
LGTM, since we are adding support for perm at Open for the server implementation maybe we can also add them in `requestFromPacket` for the request server, but we can also delay this to a future PR if required.

drakkan (approved):

client.go
puellanivis: Yes, this is why.

drakkan:
> Adding support in `requestFromPacket` was simple enough. Just a one-liner. Adding deeper support to the MemFile implementation to act as an example would take much more work, so I’ll leave that part for a later PR… or an existing PR. :trollface:

Thank you!

PR #577, opened 2024-02-09
add CloserListerAt

drakkan:
The `ListerAt` is stored in the `Request` state and reused across requests. Some implementations don't store the entire `[]os.FileInfo` buffer in the `ListerAt` implementation but instead return an open file and get/return `[]os.FileInfo` on request. For these implementation a `Close` method is required.

The use case is if you want to list a large directory on memory constrained systems.

### DIFF
--- a/request-interfaces.go
+++ b/request-interfaces.go
@@ -144,6 +144,8 @@ type NameLookupFileLister interface {
 //
 // If a populated entry implements [FileInfoExtendedData], extended attributes will also be returned to the client.
 //
+// The request server code will call Close() on ListerAt if an io.Closer type assertion succeeds.
+//
 // Note in cases of an error, the error text will be sent to the client.
 type ListerAt interface {
 	ListAt([]os.FileInfo, int64) (int, error)
--- a/request.go
+++ b/request.go
@@ -121,6 +121,22 @@ func (s *state) getListerAt() ListerAt {
 	return s.listerAt
 }
 
+func (s *state) closeListerAt() error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	var err error
+
+	if s.listerAt != nil {
+		if c, ok := s.listerAt.(io.Closer); ok {
+			err = c.Close()
+		}
+		s.listerAt = nil
+	}
+
+	return err
+}
+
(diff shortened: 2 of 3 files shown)


drakkan:
weird, `TestRequestStatVFS` seems to consistently fail in 32-bit tests

request-interfaces.go
```diff
@@ -149,6 +149,11 @@ type ListerAt interface {
 	ListAt([]os.FileInfo, int64) (int, error)
 }
 
+// CloserListerAt is a ListerAt that implements the Close method.
+type CloserListerAt interface {
```
puellanivis, line 153: Hm, convention seems to suggest this would be `ListAtCloser`.

Also, ¿typo? the interface doesn’t include `ListAt(…)` method?

request.go
```diff
@@ -111,6 +111,12 @@ func (s *state) setListerAt(la ListerAt) {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 
+	if s.listerAt != nil {
+		if closer, ok := s.listerAt.(CloserListerAt); ok {
+			closer.Close()
```
puellanivis, line 116: Do we really want to be closing this _now_ at the `setListerAt`?

request-interfaces.go
drakkan: ops, fixed

request.go
drakkan: just defensive programming. I think listerAt will be always nil here

request.go
```diff
@@ -229,6 +241,8 @@ func (r *Request) close() error {
 		}
 	}()
 
+	r.state.closeListerAt()
```
puellanivis, line 244:
```
err := r.state.closeListerAt()
```

Perhaps? This obviates the need for line 248 as well.

drakkan: maybe we can just check for io.Closer like for readers/writers instead of extending the interface

puellanivis: Yeah, I think that’s a simpler mor direct change.

drakkan: > Yeah, I think that’s a simpler mor direct change.

Done

puellanivis (approved):

Issue #579, opened 2024-03-15
Performance is much slower than a Java implementation when running in z/OS


drimmeer:
I tried to receive a file from Linux to USS of z/OS (running in a Linux emulator which is called ZD&T), and compared the performance between using GoLang and Java.

When running the built code written in GoLang and this sftp package,  it took 42 minutes and 30 seconds to receive a 940M file.
But when running a Java program (using Jsch to do sftp) using JVM, it took only 8 minutes 14 seconds to receive the same 940M file.

How come a compiled code is even slower than a interpreting code? And so slow? What could be the reason of the performance issue?

puellanivis:
Are you using [Use
ConcurrentReads](https://pkg.go.dev/github.com/pkg/sftp#UseConcurrentReads) and [Use
ConcurrentWrites](https://pkg.go.dev/github.com/pkg/sftp#UseConcurrentWrites) ?

> How come a compiled code is even slower than a interpreting code?

Look, I enjoy jokes at Java’s expense, but Java is compiled code, and not interpreted. The bytecode hasn’t been executed through an interpreter instead of a just-in-time compiler since the 90s.

But since the primary delays in transferring data across a network or the internet is usually the delay in transferring packets, I also wouldn’t be surprised if an actually interpreted language could still beat out our non-concurrent code, if it were issuing concurrent requests.

> And so slow?

An earlier design made concurrent reads and writes more dangerous, so we turned them off by default. I think we’ve worked out that issue now, but we haven’t yet switched it back to default on, because there are some really weird esoteric servers out there that have been known to unexpectedly delete people’s files if you look at the file wrong. So, out of an abundance of caution, and because it’s all documented right there in the documentation… 🤷‍♀️ 

Out-of-band, before firing off a similar hot take to any other project, I would strongly recommend checking everything a lot more thoroughly. You may be working from insufficient information, and/or invalid assumptions, and then your snark won’t come across as all too funny.

drimmeer:
Thank you Cassondra for the explanation and advice. 

First, I apologize for the harsh word I made. I didn't mean it. 
I was kind of surprised and disappointed when I saw that Go was so much slower than Java in my testing, cause the reason why I tried Go was to hope to replace a sftp tool written in Java and running in z/OS, which was too slow due to the performance of JVM, with a much faster solution.
Now your comment "But since the primary delays in transferring data across a network or the internet is usually the delay in transferring packets" brought me a second thought.

But anyway, I'd like to give you more info about what I've tried.
[...]

Issue #578, opened 2024-03-14
Does this package Support z/OS (s390) or not?


drimmeer:
Now when I tried to  run 'go build .' with my sftp project on z/OS USS, it failed with error
			../../go/pkg/mod/github.com/pkg/sftp@v1.13.6/attrs.go:101:2: undefined: fileStatFromInfoOs
			../../go/pkg/mod/github.com/pkg/sftp@v1.13.6/ls_formatting.go:70:24: undefined: lsLinksUIDGID

I am wondering if this sftp package supports IBM mainframe z system s390 or not? If yes, what could be the reason of the above problem?
If not, what needs to be done to add the support?

Thank you in advance.

puellanivis:
Looking around. I don’t know if we do. I get `go: unsupported GOOS/GOARCH pair zos/s390x` when I try. I’m also not finding support for `zos` in the builtin `syscall` library?

Can you try the `support-zos-s390x` branch and see if that works? If it doesn’t let me know the messages, and I’ll work on that.

drimmeer:
Thank you very much Cassondra, branch support-zos-s390x works! 
At least I could build my code and do a normal sftp with it!

PR #582, opened 2024-04-04
Add support for zos/s390x

### LINKED ISSUE #581: Proposal: Standardized FileInfo Structure. RE: z/OS performance investigation
Hello, Im with the Go on z/OS dev team at IBM. We got a request to look into some performance issues with this package and zos. (#578  
& #579)

When reading a file located on a linux machine to a zos machine, we see the poor performance reported by our users. The issue seems to be that despite enabling concurrent reads, [checking the filemode](https://github.com/pkg/sftp/blob/master/client.go#L1328) forces sequential reading. 

The fs.Mode returned from the remote machine is consistent with the linux format, but the zos `syscall` constants are used to mask and compare the value in `isRegular()`. These constants do not match the linux format and cause incorrect behaviour in `isRegular()` (And possibly any other functions that rely on the stat constants being consistent between machines)

Bypassing the `!isRegular()` check (forcing concurrent reads) shows performance consistent with the systems default sftp.

Maybe converting the stat packet response to the standardized [fs.FileInfo](https://pkg.go.dev/io/fs#FileInfo) struct could allow for the use of [FileMode.IsRegular()](https://pkg.go.dev/io/fs#FileMode.IsRegular)? I'd be happy to take a stab at making the change, but I'm just looking for some guidance or alternative suggestions first.

Thanks!

dustin-ward:
Closes #581 #579 #578

- Created `stat_zos.go` to alter the syscall constants that were not consistent with other platforms

- `toPflags` also needed a minor change to support platforms that have different values for `os.O_WRONLY`, `os.O_RDONLY` & `os.O_RDWR` (0x1, 0x2, 0x3 on zos respectively)

- Some other small modifications were made to tests so that they pass on zos/s390x


stat_zos.go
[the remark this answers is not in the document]
dustin-ward: Is there any reason we couldn't consolidate the filemode functions to be constant between all platforms, and keep separate implementations of the errno logic? i.e. a `stat.go` that uses the sshfx definitions for constants, and something like `errno_posix.go` & `errno_plan9.go`? Then removing both `syscall_fixed` and `syscall_good`?

dustin-ward:
Removed `stat_zos` and changed up `stat_posix`. Im not really happy with all of the type casting... But I still want to make the changes as small as possible. Any ideas?

stat_zos.go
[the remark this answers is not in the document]
puellanivis: Spinning off the `EBADF` and `syscall.Errno` handling into that one file. Then the `S_IFMT` can be put into the new common `stat.go`.

Hm… yeah. That might work. Earlier, I seem to recall that plan9 had some reason why we couldn’t use the S_ISUID, S_ISGID, or S_IVTX variables, but it looks like the `os.ModeSetuid` has been around since go1.10. I guess we were just stuck on “yeah, but we gotta use the `syscall` package if we’ve got it.”

But yeah, if we convert to using the `internal/encoding/ssh/filexfer` values, then there’s no reason why plan9 would need to be different in this matter.

👍 Good call.

client.go
```diff
@@ -2142,7 +2143,7 @@ func toPflags(f int) uint32 {
 // setuid, setgid and sticky in m, because we've historically supported those
 // bits, and we mask off any non-permission bits.
 func toChmodPerm(m os.FileMode) (perm uint32) {
-	const mask = os.ModePerm | s_ISUID | s_ISGID | s_ISVTX
+	const mask = os.ModePerm | iofs.FileMode(s_ISUID|s_ISGID|s_ISVTX)
```
puellanivis, line 2146: Consider aligning `os.ModePerm` and `io/fs.FileMode`.

client_integration_test.go
```diff
@@ -974,7 +975,7 @@ func TestClientSetuid(t *testing.T) {
 	f.Close()
 
 	const allPerm = os.ModePerm | os.ModeSetuid | os.ModeSetgid | os.ModeSticky |
-		s_ISUID | s_ISGID | s_ISVTX
+		iofs.FileMode(s_ISUID|s_ISGID|s_ISVTX)
```
puellanivis, line 978: Same here, let’s go ahead and align on `os` or `io/fs`.

puellanivis (commented):
On the lines where you’re adding `io/fs` for something, there are available `os` aliases. I’m not saying to use the `os` version, but it is a possible way to skirt the name clash between `io/fs` and `github.com/kr/fs`.

Alternatively, I would recommend renaming `kr/fs` rather than the standard library. (Even though this probably is a larger change surface.)

stat.go
```diff
@@ -0,0 +1,93 @@
...
+	return uint32(ret)
+}
+
+const (
+	s_ISUID = uint32(sshfx.ModeSetUID)
+	s_ISGID = uint32(sshfx.ModeSetGID)
+	s_ISVTX = uint32(sshfx.ModeSticky)
+)
+
+// Legacy export:
```
puellanivis, line 88: [golint] godoc should start with `S_IFMT`

(I know the comments on the pre-existing code wasn’t ideal, and you’re just moving it around.)

puellanivis (approved):
Looks great. So much thanks.

dustin-ward:
Thanks for your feedback! ✌️

stat.go
```diff
@@ -85,9 +85,5 @@ const (
 	s_ISVTX = uint32(sshfx.ModeSticky)
 )
 
-// Legacy export:
-//
-// Go defines S_IFMT on windows, plan9 and js/wasm as 0x1f000 instead of
-// 0xf000. None of the the other S_IFxyz values include the "1" (in 0x1f000)
-// which prevents them from matching the bitmask.
+// S_IFMT is a legacy export. `sshfx.ModeType` should be used instead
```
puellanivis, line 88: The problem is that the `sshfx` package is `internal` and so no one else can import it. So, the given alternative isn’t useful.

I would say that the local OS might use a different value from what we need to use internally, and there is no reason anyone should need this specific internal value.

We can also merge, and then I can adjust it after.

dustin-ward: Ah I see. That makes more sense. Better to have you leave the comment anyways. Thanks again

dustin-ward:
Would it also be possible to get a new tag that includes the recent changes?

puellanivis:
I want to get the `WithWindowsRootEnumeratesDrives` also checked in before I cut another release.

PR #583, opened 2024-04-25
Add SetExtendedAttrs to Client

peterverraedt:
Add function to set extended attributes in the sftp client.

### DIFF
--- a/client.go
+++ b/client.go
@@ -593,6 +593,20 @@ func (c *Client) Truncate(path string, size int64) error {
 	return c.setstat(path, sshFileXferAttrSize, uint64(size))
 }
 
+// SetExtendedData sets extended attributes of the named file. It uses the
+// SSH_FILEXFER_ATTR_EXTENDED flag in the setstat request.
+//
+// This flag provides a general extension mechanism for vendor-specific extensions.
+// Names of the attributes should be a string of the format "name@domain", where "domain"
+// is a valid, registered domain name and "name" identifies the method. Server
+// implementations SHOULD ignore extended data fields that they do not understand.
+func (c *Client) SetExtendedData(path string, extended []StatExtended) error {
+	attrs := &FileStat{
+		Extended: extended,
+	}
+	return c.setstat(path, sshFileXferAttrExtended, attrs)
+}
+
 // Open opens the named file for reading. If successful, methods on the
 // returned file can be used for reading; the associated file descriptor
 // has mode O_RDONLY.
--- a/client.go
+++ b/client.go
@@ -2058,6 +2058,28 @@ func (f *File) Chmod(mode os.FileMode) error {
 	return f.c.fsetstat(f.handle, sshFileXferAttrPermissions, toChmodPerm(mode))
 }
 
+// SetExtendedData sets extended attributes of the current file. It uses the
+// SSH_FILEXFER_ATTR_EXTENDED flag in the setstat request.
+//
+// This flag provides a general extension mechanism for vendor-specific extensions.
+// Names of the attributes should be a string of the format "name@domain", where "domain"
+// is a valid, registered domain name and "name" identifies the method. Server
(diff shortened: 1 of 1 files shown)


client.go
```diff
@@ -593,6 +593,16 @@ func (c *Client) Truncate(path string, size int64) error {
...
 }
 
+// SetExtendedAttrs sets the extended attributes of the named file.
+func (c *Client) SetExtendedAttrs(path string, extended []StatExtended) error {
+	type extendedAttrs struct {
+		Size     uint32
+		Extended []StatExtended
+	}
+	attrs := extendedAttrs{Size: uint32(len(extended)), Extended: extended}
+	return c.setstat(path, sshFileXferAttrExtended, attrs)
```
puellanivis, line 603: Please use `&FileStat{ Extended: extended }` instead of a custom `struct`.

client.go
```diff
@@ -593,6 +593,16 @@ func (c *Client) Truncate(path string, size int64) error {
 	return c.setstat(path, sshFileXferAttrSize, uint64(size))
 }
 
+// SetExtendedAttrs sets the extended attributes of the named file.
+func (c *Client) SetExtendedAttrs(path string, extended []StatExtended) error {
```
puellanivis, line 597: It seems like the common abbreviation for “extended attributes” is `xattr` So, that would suggest `SetXattr`, which would match the `setxattr` call in OpenSSH. But I don’t think we want our name that short. `SetExtAttrs()` might be a good inbetween? But I’m still fine with this longer name as well.

Maybe we could also implement this as `SetExtAttrs(path string, xattrs ...StatExtended) error` … though checking it on go playground, it seems this doesn’t allow for the type-elided shorthand `cl.SetExtAttr(path, { ExtType: "foo", ExtData: "bar" })` to be used. 🙁  Meanwhile, `cl.SetExtAttrs(path, []sftp.StatExtended{ { ExtType: "foo", ExtData: "bar" } })` only needs the one `sftp.StatExtended` type name to be used for any number of xattrs.

A common API design that some have picked is `SetExtAttrs(path string, xattrs ...string)` and then provide alternating `type` and `data` values that get turned into the `[]sftp.StatExtended` internally. But I’m not a big fan of this easy-to-misalign approach. But then again, it is integrated into the standard `slog` library, so… 🤷‍♀️ 

In the alternative, we could provide two functions: `SetExtAttrs(path string, xattrs []StatExtended)` and `SetExtAttr(path, typ, data string)` thus anyone wanting to set a single attribute can avoid the `StatExtended` altogether, and then for more than one, they can make use of the `[]StatExtended{ {}, {} }` type elision.

puellanivis (commented):
I suppose if we’re adding this to `*Client` we should could add a parallel `*File` implementation using the `fsetstat` as well.

client.go
peterverraedt: This was a very old commit locally, it seems that I followed the example of `Lchown` and others above.

client.go
peterverraedt: Following the [draft RFCs](https://datatracker.ietf.org/doc/html/draft-ietf-secsh-filexfer-13#section-7.14) the extended attributes in the SFTP protocol do not necessarily match to extended atributes on the system, so I would avoid the very short `xattr` to avoid confusion. Maybe `SetExtendedData` is a better name to correspond with [the FileInfoExtendedData interface](https://github.com/pkg/sftp/blob/637088883317ec6aced847f1f5d5b007354e0145/attrs.go#L96-L99) ?

Another catch is that it is very unspecified what `setstat` actually does with extended attributes. It might, in case of real extended attributes on files, just replace all current attributes.

puellanivis: Second checked the standard, and the field is called `extended_data`, so 👍

client.go
puellanivis: I think I changed the `Chown` function recently as well. 😂 It used to be just like `[]uint32{ uid, gid}` or something.

PR #584, opened 2024-04-25
Add WithMaxTxPacket server option

peterverraedt:
Add the WithMaxTxPacket and WithRSMaxTxPacket server options to increase the maximum tx packet size to a value above 32K. This allows to send bigger chunks of data to the client as response to a read request. As the client specifies the wanted length, it should be safe to increase the server maximum value.

This in particular allows the implemented Client with the MaxPacketUnchecked option to retrieve data in larger chunks.

### DIFF
--- a/server.go
+++ b/server.go
@@ -34,6 +34,7 @@ type Server struct {
 	openFilesLock sync.RWMutex
 	handleCount   int
 	workDir       string
+	maxTxPacket   uint32
 }
 
 func (svr *Server) nextHandle(f *os.File) string {
@@ -86,6 +87,7 @@ func NewServer(rwc io.ReadWriteCloser, options ...ServerOption) (*Server, error)
 		debugStream: ioutil.Discard,
 		pktMgr:      newPktMgr(svrConn),
 		openFiles:   make(map[string]*os.File),
+		maxTxPacket: defaultMaxTxPacket,
 	}
 
 	for _, o := range options {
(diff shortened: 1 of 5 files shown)


server.go
```diff
@@ -139,6 +141,24 @@ func WithServerWorkingDirectory(workDir string) ServerOption {
...
+// WithMaxTxPacket sets the maximum size of the payload returned to the client,
+// measured in bytes. The default value is 32768 bytes, and this option
+// can only be used to increase it. Setting this option to a larger value
+// should be safe as the client decides the size of the requested payload.
+//
+// The default maximum packet size is 32768 bytes.
+func WithMaxTxPacket(size uint32) ServerOption {
+	return func(s *Server) error {
+		if size < 32768 {
+			return errors.New("size must be greater or equal to 32768")
```
puellanivis, line 153: > “greater than or equal to”

server.go
```diff
@@ -513,16 +533,16 @@ func (p *sshFxpSetstatPacket) respond(svr *Server) responsePacket {
 
 	fs, err := p.unmarshalFileStat(p.Flags)
 
-	if err == nil && (p.Flags & sshFileXferAttrSize) != 0 {
+	if err == nil && (p.Flags&sshFileXferAttrSize) != 0 {
```
puellanivis, line 536: The changes here should be a result of a `go fmt` and ¿should have been `go fmt`ed before?

Just checking to make sure that this is consistent with `go fmt` of the most recent version.

peterverraedt: This is indeed the result of `go fmt`, my local version is `go1.21.1`.

puellanivis: I’ve checked with `go1.22.2` it does the same. 👍

Issue #587, opened 2024-06-03
custom interface fails to trigger concurrent reads in `ReadFrom`


emar-kar:
Hello there.

I have a project, where I need to upload different files via sftp and I use concurrent reads/writes with config:
```go
sftpCl, err := sftp.NewClient(
	sshConn,
	sftp.UseConcurrentReads(true),
	sftp.UseConcurrentWrites(true),
	sftp.MaxConcurrentRequestsPerFile(64),
	sftp.MaxPacket(32768), // Max possible number per packet.
)
```

 In addition, I need graceful shutdown thus, I wrapped my file reader with context:
```go
type ReaderWithContext struct {
	ctx context.Context
	r   io.Reader
}

func (r *ReaderWithContext) Read(p []byte) (int, error) {
	if err := r.ctx.Err(); err != nil {
		return 0, err
	}

	return r.r.Read(p)
}
```

But I noticed a huge regression in speed, where my test data ~15MB was uploading  >1 min compare to ~12 sec without custom interface. After debugging, I found out that the actual problem was inside the `ReadFrom` function, since it defines [remain](https://github.com/pkg/sftp/blob/54946566af05c94e0e823d42d2e31d04411ab249/client.go#L1928) but only if `io.Reader` implements Len(), Size(), Stat() or it is an `io.LimitedReader`. In other cases it starts read/write without concurrency. I assume it is the way to avoid situations when remain is 0 and signalise of `io.EOF` or reader errors. So, we perform single Read, to define this and not to start all the routines for concurrent read/writes.

I can solve my situation with explicit call to `ReadFromWithConcurrency` or wrap my context reader with `io.LimitedReader`.
But maybe there can be an addition to `ReadFrom` func docs, which explain/warn about this specific behaviour? Or cover situation when **remain** is 0 with an attempt to call `ReadFromWithConcurrency` anyway?

puellanivis:
We have had corner cases where we just want to be extra safe that we don’t use concurrency if we cannot know that the operations are safe.

However, I’m unsure if how much—if any—these concerns still exist since we reworked how concurrency works. Now, I think we’re just trying to attempt avoiding trying to spin off a ton of goroutines just for them to not pick up any work.

puellanivis:
I’m not opposed to either or both changes. If you have a good idea for the message you could open a PR with a suggestion. Otherwise, I’ll look into getting to it.

puellanivis:
Great job, thanks!

PR #588, opened 2024-06-04
update ReadFrom and ReadFromWithConcurrency docs

emar-kar:
Related #587

### DIFF
--- a/client.go
+++ b/client.go
@@ -1757,7 +1757,8 @@ func (f *File) writeAt(b []byte, off int64) (written int, err error) {
 //
 // Otherwise, the given concurrency will be capped by the Client's max concurrency.
 //
-// This method is preferred over calling ReadFrom to guarantee concurrent reads/writes.
+// When one needs to guarantee concurrent reads/writes, this method is preferred
+// over ReadFrom.
 func (f *File) ReadFromWithConcurrency(r io.Reader, concurrency int) (read int64, err error) {
 	f.mu.Lock()
 	defer f.mu.Unlock()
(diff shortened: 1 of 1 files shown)


client.go
```diff
@@ -1756,6 +1756,8 @@ func (f *File) writeAt(b []byte, off int64) (written int, err error) {
 // Giving a concurrency of less than one will default to the Client’s max concurrency.
 //
 // Otherwise, the given concurrency will be capped by the Client's max concurrency.
+//
+// This method is preferred over calling ReadFrom to guarantee concurrent reads/writes.
```
puellanivis, line 1760: I don’t know if we want to use the word “preferred”, or at least I would want to move the reason/motivation to the front.

That is, “When one needs to guarantee concurrent reads/writes, this method is preferred over ReadFrom.”

client.go
```diff
@@ -1916,6 +1918,17 @@ func (f *File) readFromWithConcurrency(r io.Reader, concurrency int) (read int64
...
+//
+// If client uses concurrent writes, given r needs to implement one of the interfaces:
+//
+//	Len()  int
+//	Size() int64
+//	Stat() (os.FileInfo, error)
+//
+// or be an instance of [io.LimitedReader] to determine the number of possible
+// concurrent requests. Otherwise, reads/writes are performed in nonparallel, sliced
+// in chunks with the max packet size. ReadFromWithConcurrency explicit call can
```
puellanivis, line 1930: [English usage] I think we need to say “are not performed in parallel”. Or more consistent with other usage in the package “are performed sequentially” or “are not performed concurrently”.

[over-specific] I’m also unsure we need to include the detail that it’s sliced up into chunks, because both implementations sequential and concurrent do this. By calling this out here, it implies that the concurrent implementation doesn’t.

puellanivis (commented):
Very good, just a few notes, and some small English grammar gotchas.

puellanivis (approved):

PR #611, opened 2024-12-30
Issue #609 - clarify what 'path contains a regular file' actually means

### LINKED ISSUE #609: client.MkdirAll not returning an error when docs state it should
Hi, I am just making a project learning about sftp and transferring data, anyways, I am making a directory, creating a file within it, and then extending that file directory by one. According to the docs of [MkdirAll](https://pkg.go.dev/github.com/pkg/sftp#Client.MkdirAll) it says:
> MkdirAll creates a directory named path, along with any necessary parents, and returns nil, or else returns an error. If path is already a directory, MkdirAll does nothing and returns nil. If path contains a regular file, an error is returned

however `nil` is still received back in the code below
```go
	// connect to ssh server
	conn, err := ssh.Dial("tcp", sshAddr, config)
	if err != nil {
		log.Fatal("Failed to dial: ", err)
	}

	defer conn.Close()

	// open an SFTP session over an existing ssh connection.
	client, err := sftp.NewClient(conn)
	if err != nil {
		log.Fatal(err)
	}
	defer client.Close()

	var user = "USER"
	var sftpBasePath = fmt.Sprintf("home/%s", user)
	err = client.MkdirAll(sftpBasePath)
	if err != nil {
		log.Fatal(err)
	}

	// leave your mark
	f, err := client.Create(fmt.Sprintf("%s/hello.txt", sftpBasePath))
	if err != nil {
		log.Fatal(err)
	}

	if _, err := f.Write([]byte("Hello world!")); err != nil {
		log.Fatal(err)
	}

	// error should be returned below as there is a ".txt"
        // file in the path before creating the `dir` directory
	err = client.MkdirAll(sftpBasePath + "/dir")
	if err != nil {
		log.Fatal(err)
	}

	f.Close()
```

I should also mention that the ssh server when I inspect the documents the file is created at that path and the `dir` directory

puellanivis:
The text of the `sftp.Client.MkdirAll` was found to be unclear as to what precisely “path contains a regular file” means. It means that if any path element in the chain of path elements is a regular file, that an error is returned. However, it could be mistakenly interpreted as “if any of the parent directories contains a regular file”. This new language is intended to exclude this possible reading.

Resolves #609 

The dev-v2 branch does not include such language, instead paralleling the `os.MkdirAll` language, which also does not explicitly call out this error condition.

### DIFF
--- a/client.go
+++ b/client.go
@@ -909,7 +909,7 @@ func (c *Client) Mkdir(path string) error {
 // MkdirAll creates a directory named path, along with any necessary parents,
 // and returns nil, or else returns an error.
 // If path is already a directory, MkdirAll does nothing and returns nil.
-// If path contains a regular file, an error is returned
+// If, while making any directory, that path is found to already be a regular file, an error is returned.
 func (c *Client) MkdirAll(path string) error {
 	// Most of this code mimics https://golang.org/src/os/path.go?s=514:561#L13
 	// Fast path: if we can tell whether path is a directory or file, stop with success or error.


ctfrancia (commented):
I find that this reads better

PR #616, opened 2025-02-21
add support wasm

labulakalia:


puellanivis (changes requested):
The `GOOS` / `GOARCH` is `js/wasm`, and we are only (generally) listing `GOOS` parameters. https://pkg.go.dev/internal/platform#pkg-variables

I think what we want to add is `wasip1` in this case, as that seems to be the missing `GOOS`. https://go.dev/blog/wasi

P.S. Oh, also, would you mind putting the new `wasip1` next to the `js`? That way they’re more grouped together. (Principle: spatial locality of related concepts.)

labulakalia:
ok,look it  good,I close this pr.

PR #619, opened 2025-03-04
A bit of a refactor, match a similar signature to protobuf opaque API

puellanivis:
There are a few code improvements, but mostly this is changing signatures on the `sshfx.Attributes` object.

### DIFF
--- a/localfs/server_plan9.go
+++ b/localfs/server_plan9.go
@@ -7,15 +7,15 @@ import (
 
 func toLocalPath(p string) (string, error) {
 	if path.IsAbs(p) {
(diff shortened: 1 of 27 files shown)


localfs/server_plan9.go
```diff
@@ -7,15 +7,15 @@ import (
 
 func toLocalPath(p string) (string, error) {
 	if path.IsAbs(p) {
-		tmp := lp[1:]
```
puellanivis, line 10: Yep… there was a compile bug in the `GOOS=plan9` build. 🤦‍♀️

PR #623, opened 2025-03-24
perf: ensure full write buffer for *File.ReadFrom

justfalter:
In `*File.ReadFrom`, ensure that the buffer is filled to capacity (`maxPacket` length) before performing the write.

Prior to this change, the amount of data read into the buffer was dictated by the `io.Reader`'s `Read` implementation, and write performance would suffer when the Read would return less than maxPacket bytes of data. An example source would be a `net/http` response `Body` from a TLS server, which seems to cap each read to 16384 bytes -- half the default max packet size of 32768 bytes.

### DIFF
--- a/client.go
+++ b/client.go
@@ -1842,7 +1842,8 @@ func (f *File) readFromWithConcurrency(r io.Reader, concurrency int) (read int64
 		off := f.offset
 
 		for {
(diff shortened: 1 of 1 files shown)


puellanivis (approved):
This does look like a solid improvement.

PR #624, opened 2025-03-25
fix: maybe hang for process *sshfx.ReadPacket for *Server.handle()"

cloudsbit:
In `*Server.handle`, if the *sshfx.ReadPacket's Length is less than maxDataLen, the client may get stuck during data reading.

### DIFF
--- a/server.go
+++ b/server.go
@@ -700,7 +700,7 @@ func (srv *Server) handle(req sshfx.Packet, hint []byte, maxDataLen uint32) (ssh
 				return nil, fmt.Errorf("read length request too large: %d", req.Length)
 			}
 
-			hint = slices.Grow(hint[:0], req.Length)[:req.Length]
+			hint = slices.Grow(hint[:0], int(req.Length))[:req.Length]
 			 
 			n, err := file.ReadAt(hint, int64(req.Offset))
 			if err != nil {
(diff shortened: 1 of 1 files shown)


server.go
```diff
@@ -699,7 +699,7 @@ func (srv *Server) handle(req sshfx.Packet, hint []byte, maxDataLen uint32) (ssh
 				return nil, fmt.Errorf("read length request too large: %d", req.Length)
 			}
 
-			n, err := file.ReadAt(hint, int64(req.Offset))
+			n, err := file.ReadAt(hint[:req.Length], int64(req.Offset))
```
puellanivis, line 702: 🤔 I think we need a check to ensure that `req.Length` isn’t greater than the length of `hint`. I mean, I think currently it’s a known logically true statement about this implementation because of line 698 and the allocation at 427 is for `maxDataLen`.

I would like to see something more robust, in case the hint changes. (Yes, I should have already done this. 🙈 )
```suggestion
			hint = slices.Grow(hint[:0], req.Length)[:req.Length]
			 
			n, err := file.ReadAt(hint, int64(req.Offset))
```

puellanivis (commented):
Good call, but we can also be a little more robust, in case `hint` isn’t populated.

puellanivis (approved):

Issue #625, opened 2025-03-26
CVE-2025-22869


kbutz:
Vulnerability info: https://pkg.go.dev/vuln/GO-2025-3487 and https://nvd.nist.gov/vuln/detail/CVE-2025-22869

impacting golang.org/x/crypto/ssh before v0.35.0

puellanivis:
Resolved through #626

PR #626, opened 2025-03-26
Update crypto version to v0.35.0 sftp#625

kbutz:
Related to issue https://github.com/pkg/sftp/issues/625


go.mod
```diff
@@ -1,10 +1,18 @@
 module github.com/pkg/sftp
 
-go 1.15
+go 1.23.0
+
+toolchain go1.24.0
```
puellanivis, line 5: I’m kind of the opinion that we do not want to have a `toolchain` directive for libraries. We’re running our tests against the specified versions in our CI to ensure that it is compatible with all recent toolchains. Having the toolchain directive basically requires the use of a specific version of the toolchain, but… we shouldn’t need or want that?

Like, we want our code to compile with whatever version of Golang they’re using. We’re targeting a sort of toolchain agnostic code compatibility.

kbutz: Thanks! I pushed a change to update to go 1.24.0 and remove the toolchain directive. The first push was the original output of go get & go mod tidy.

puellanivis (approved):

PR #630, opened 2025-03-29
Remove unreferenced error declarations

kbutz:


kbutz:
While I was in there, I spotted some err vars that are unused by the client and rest of the sftp repo. Feel free to close this PR is they should remain!

puellanivis (commented):
Hm… it seems there are codebases that are using this variable, even though it’s entirely unused otherwise.

This is basically why we have all the deprecated error variables that are now prefixed with `Err`.

While I think it’s generally a good idea to keep an API surface clean, we live with what we were given, and try not to break code that is using our code, and `!strings.Contains(err.Error(), sftp.ErrInternalInconsistency.Error())` isn’t broken right now, but would break with this change.

Usually, API cleanups are scheduled and delayed for a v2, release, which is already in the `dev-v2` branch.

kbutz:
That makes sense, I'll close this. Thanks!

Issue #627, opened 2025-03-28
dev-v2  check server.handle() return value


cloudsbit:
In server.go, When checking if the return value resp of server.handle() is nil, if the sshfx.Packet interface holds a pointer value, using resp == nil will not work.
```
		} else {
			resp, err = srv.handle(reqPkt.Request, dataHint, maxDataLen)
		}

		if resp == nil {
			resp = errorToStatus(err)
		}
```

puellanivis:
This was actually intentional. `Server.handle()` returns `(sshfx.Packet, error)` so, an `interface`, and an `error`. If we return anything other than an untyped nil as the `sshfx.Packet` I was thinking that’s what we want to use that as the response, regardless of any error being returned. So basically, if the `srv.handle()` returns _anything_ other than an untyped nil, then the intention that `srv.handle()` is trying to convey to the caller is that that is the response that should be used.

cloudsbit:
> This was actually intentional. `Server.handle()` returns `(sshfx.Packet, error)` so, an `interface`, and an `error`. If we return anything other than an untyped nil as the `sshfx.Packet` I was thinking that’s what we want to use that as the response, regardless of any error being returned. So basically, if the `srv.handle()` returns _anything_ other than an untyped nil, then the intention that `srv.handle()` is trying to convey to the caller is that that is the response that should be used.

A comprehensive explanation， but `Server.handle()` returns `(sshfx.Packet, error)` fails to capture the nil error,  when I call the StatVFS interface, `statvfser.StatVFS()` returns errors like `(nil，fs.ErrNotExist)`. Is this intended behavior?

puellanivis:
Ah… hm… I think I’m seeing the problem now. We’ll need to change the type returned by the `get()` generic, to return an `sshfx.Packet` and convert the typed nil to an untyped nil. 🤔

puellanivis:
Resolved

PR #631, opened 2025-03-31
ISSUE #627: convert typed-nils to untyped-nils on error

puellanivis:
Addresses #627 , we handle converting certain getters into an untyped-nil `sshfx.Packet` on non-nil error.


server.go
puellanivis, line 1: I’ve checked all the other cases of returns in `Server.handle()`, and they all test for `err != nil` immediately after.

cloudsbit: WOW, Solid change!

PR #633, opened 2025-06-01
Don't translate StatusPacket to os.Err(NotExist|Permission)

greatroar:
Doing this loses potentially useful information for the client. They should instead check with errors.Is(err, fs.ErrNotExist), etc.; the StatusPacket type has an Is method for this already.

### DIFF
--- a/client.go
+++ b/client.go
@@ -490,10 +490,6 @@ func statusToError(status *sshfx.StatusPacket, okExpected bool) error {
 
 	case sshfx.StatusEOF:
 		return io.EOF
-	case sshfx.StatusNoSuchFile:
-		return fs.ErrNotExist
-	case sshfx.StatusPermissionDenied:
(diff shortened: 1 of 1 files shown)


puellanivis:
Hm… 🤔 I can’t see any reason against this. It’s also for the `dev-v2` branch, and so the potential change of the behavior would be isolated to a future breaking change.

And as noted, unlike `io.EOF` that gets a lot of regular `== io.EOF` checks in the standard library and elsewhere, these errors really should be using `errors.Is(err, fs.ErrNotExist)` and `errors.Is(err, fs.ErrPermission)` as called out in the `os.IsNotExist` and `os.IsPermission`.

puellanivis:
It seems was introduced before this package even released as `v1.0.0`.

Regardless, we have the better option now, and v2 should support the new way rather than keep an old way.

PR #635, opened 2025-08-21
Updating go packages

Farokhcpu:


puellanivis:
Oh yes, I ran the checks to then merge. Remember that due to how Golang handles version satisfaction, the versions of libraries used by imported modules can be overridden by using a newer version explicitly in your own code.

PR #637, opened 2025-10-06
add security policy

drakkan:
This should enable reporting private security issues via GitHub

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,13 @@
+# Security Policy
+
+## Supported Versions
+
+Security updates are provided for the latest released version of this package.
+We also welcome vulnerability reports for the development version to help us ensure it is secure before the next release.
+
+## Reporting a Vulnerability
+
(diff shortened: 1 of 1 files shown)


puellanivis:
Great idea! While we offload our transport security to ssh, we of course, should still take security seriously within our own functionality.

drakkan:
For some reason the "Report a vulnerability" is still missing. See for example here: https://github.com/drakkan/sftpgo/security

puellanivis:
I do see it now, so 👍

drakkan:
> I do see it now, so 👍

Hmm, that’s odd, I’m still getting a 404 when I open [this link](https://github.com/pkg/sftp/security/advisories/new)

drakkan:
I don’t have the necessary permissions to change the settings myself

Screenshot From 2025-10-08 18-55-38

eikenb:
> I also do not have permissions.

You should have them now. I've added you as an Admin. Sorry about that, didn't realize I was the last admin standing.

PR #643, opened 2026-01-09
Add exported error constants for all error codes.

theraphim:
The error codes starting from 9 are defined internally and can be returned by the server, but without exported error constants it's impractical to match on and handle these errors.

This change adds Err constants and text error strings for these codes.

### DIFF
--- a/request-errors.go
+++ b/request-errors.go
@@ -6,15 +6,38 @@ type fxerr uint32
 // direct control of the errors being sent vs. letting the library work them
 // out from the standard os/io errors.
 const (
-	ErrSSHFxOk               = fxerr(sshFxOk)
-	ErrSSHFxEOF              = fxerr(sshFxEOF)
-	ErrSSHFxNoSuchFile       = fxerr(sshFxNoSuchFile)
-	ErrSSHFxPermissionDenied = fxerr(sshFxPermissionDenied)
-	ErrSSHFxFailure          = fxerr(sshFxFailure)
-	ErrSSHFxBadMessage       = fxerr(sshFxBadMessage)
-	ErrSSHFxNoConnection     = fxerr(sshFxNoConnection)
-	ErrSSHFxConnectionLost   = fxerr(sshFxConnectionLost)
-	ErrSSHFxOpUnsupported    = fxerr(sshFxOPUnsupported)
+	ErrSSHFxOk                      = fxerr(sshFxOk)
+	ErrSSHFxEOF                     = fxerr(sshFxEOF)
+	ErrSSHFxNoSuchFile              = fxerr(sshFxNoSuchFile)
+	ErrSSHFxPermissionDenied        = fxerr(sshFxPermissionDenied)
+	ErrSSHFxFailure                 = fxerr(sshFxFailure)
+	ErrSSHFxBadMessage              = fxerr(sshFxBadMessage)
+	ErrSSHFxNoConnection            = fxerr(sshFxNoConnection)
+	ErrSSHFxConnectionLost          = fxerr(sshFxConnectionLost)
+	ErrSSHFxOpUnsupported           = fxerr(sshFxOPUnsupported)
+	ErrSSHFxInvalidHandle           = fxerr(sshFxInvalidHandle)
+	ErrSSHFxNoSuchPath              = fxerr(sshFxNoSuchPath)
+	ErrSSHFxFileAlreadyExists       = fxerr(sshFxFileAlreadyExists)
+	ErrSSHFxWriteProtect            = fxerr(sshFxWriteProtect)
+	ErrSSHFxNoMedia                 = fxerr(sshFxNoMedia)
+	ErrSSHFxNoSpaceOnFilesystem     = fxerr(sshFxNoSpaceOnFilesystem)
+	ErrSSHFxQuotaExceeded           = fxerr(sshFxQuotaExceeded)
+	ErrSSHFxUnknownPrincipal        = fxerr(sshFxUnknownPrincipal)
+	ErrSSHFxLockConflict            = fxerr(sshFxLockConflict)
+	ErrSSHFxDirNotEmpty             = fxerr(sshFxDirNotEmpty)
+	ErrSSHFxNotADirectory           = fxerr(sshFxNotADirectory)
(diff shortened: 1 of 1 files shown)


request-errors.go
```diff
@@ -28,27 +51,76 @@ const (
 	ErrSshFxNoConnection     = ErrSSHFxNoConnection
 	ErrSshFxConnectionLost   = ErrSSHFxConnectionLost
 	ErrSshFxOpUnsupported    = ErrSSHFxOpUnsupported
+	ErrSshFxInvalidHandle    = ErrSSHFxInvalidHandle
```
puellanivis: These error codes are deprecated. We should not be adding any new ones.

request-errors.go
```diff
@@ -28,27 +51,76 @@ const (
...
+		return "A badly formatted packet or other SFTP protocol incompatibility was detected."
 	case ErrSSHFxNoConnection:
-		return "no connection"
+		return "There is no connection to the server."
 	case ErrSSHFxConnectionLost:
-		return "connection lost"
+		return "The connection to the server was lost."
 	case ErrSSHFxOpUnsupported:
-		return "operation unsupported"
+		return "An attempted operation could not be completed by the server because the server does not support the operation."
```
puellanivis: Do not change the error strings returned by any previously provided errors, as this would be a breaking change of behavior.

request-errors.go
```diff
@@ -28,27 +51,76 @@ const (
...
+	case ErrSSHFxDeletePending:
+		return "An operation was attempted on a file for which a delete operation is pending."
+	case ErrSSHFxFileCorrupt:
+		return "The file is corrupt; an filesystem integrity check should be run."
+	case ErrSSHFxOwnerInvalid:
+		return "The principal specified can not be assigned as an owner of a file."
+	case ErrSSHFxGroupInvalid:
+		return "The principal specified can not be assigned as the primary group of a file."
+	case ErrSSHFxNoMatchingByteRangeLock:
+		return "The requested operation could not be completed because the specified byte range lock has not been granted."
```
puellanivis: https://go.dev/wiki/CodeReviewComments#error-strings Do not capitalize error strings, and do not replace short simple codes with long verbose explanations of those errors.

puellanivis (commented):
Error codes starting from 32 can also possibly be returned from the server. But we’re not supporting them because they are not yet defined.

In a similar vein, this package specifically only supports https://filezilla-project.org/specs/draft-ietf-secsh-filexfer-02.txt for which these errors are also not defined, and so no server using this package should be returning them, and since this package’s client has not negotiated an upgrade of version, a properly compliant server should also never return these codes.

If anyone does want to use these extended error codes, then I would recommend they instead switch to the `dev-v2` branch, which has taken an intentional approach towards supporting these errors in its `encoding/ssh/filexfer` package.

request-errors.go
theraphim: Done

request-errors.go
theraphim: Done

request-errors.go
theraphim: I have removed the capitalization and the dot at the end. As for the text, I think there's an opportunity to return a meaningful message if we do not need to care about backwards compatibility. That said, if you insist, I can put in some boring names based on the error constant.

puellanivis: > As for the text, I think there's an opportunity to return a meaningful message if we do not need to care about backwards compatibility.

Hyrum’s law dictates that we cannot ever know exactly who is depending on what. That’s why I’m recommending keeping the old strings. If you’ll notice, `dev-v2` switches the error strings. This is the appropriate place to recommend new error strings, where the v1 compat guarantees are not applicable.

With the original strings for the original messages in place, now the overly verbose new messages do not fit in context. While I can understand the depiction of these strings as “boring”, that’s what is already here. Having such a dichotomy between the two kinds of errors is dissonant, and a violation of the principle of least surprise.

puellanivis:
I repeat that adding this feature to the `v1` branch seems inappropriate when this functionality is already provided intentionally by the `v2` branch. If we had more testers using the `dev-v2` branch, then we could be confident enough to release a `v2.0.0-alpha` (perhaps I should just do this anyways).

With a `v2` in the wings, I feel a feature freeze of the `v1` implementation is appropriate. There are already bugs that impact `v1` branch but not the `v2` branch, because implementing them in the `v1` branch would require major refactoring that would be essentially equivalent to just redoing most of the `dev-v2` branch anyways.

theraphim:
Some of my colleagues use this package, to some extent because I used it 10 years ago, to talk to various network devices (and you know how quirky the protocol support can be for these; there should be a commit in the history with my name on it related to that).
While I moved on from that project, the ongoing maintenance is done by someone else. I don't think it's realistic to ask them to rewrite that piece of infrastructure on a short notice to a branch which is essentially unpublished and untested (at least what the readme says). So I thought that exposing the error codes should be relatively simple change - it doesn't add any functionality per se, just makes it possible to match on errors, and I thought it will be useful to community.

It's your decision after all, if this is not accepted I'll just update my private fork.

Thanks.

puellanivis:
Yeah, I understand how simple the change is, and why it might be useful… it’s just more of a “yeah, but technically all of those are explicitly unsupported by the version we’re supporting.” They might be defined, but also… not really. 😩 

It’s a rough choice especially when just about everything stopped supporting any version except the one we’re on. Like, I believe they’re trying to start up a newest version, and the plan is basically to reset back to this version, because no one was really ever supporting anything beyond this version. 🤦‍♀️ 

@drakkan What’s your perspective on this? I don’t want to be heavy handed here, especially when I’m conflicted even myself.

drakkan:
I’d like to better understand your use case. As @puellanivis explained, the error codes are defined by the SFTP v3 specification that we implement. OpenSSH, which follows the same specification as this package, defines the same set of [error codes](https://github.com/openssh/openssh-portable/blob/d01efaa1c9ed84fd9011201dbc3c7cb0a82bcee3/sftp-server.c#L529). Since we advertise SFTP protocol version 3, a compliant implementation is expected to return only the error codes defined in the negotiated specification.

Could you share which SFTP implementation is returning error codes from later specification versions? Is this something we can test ourselves, or is it a proprietary implementation? Thank you

PR #645, opened 2026-04-02
Add build tag to omit the x/crypto/ssh dependency

espadolini:
The only use of `golang.org/x/crypto/ssh` (and, transitively, of any package in `crypto/...`) is in the `NewClient` function, which is a small wrapper around running a subsystem request with an `ssh.Client` to then run a sftp client on the channel.

For binaries that don't otherwise depend on `x/crypto/ssh` (for example utility binaries that only do sftp through stdio) this can be a somewhat significant chunk of binary size; for example a stripped cgoless build of `server_standalone` (darwin arm64, go 1.26.1) goes from 2.9 to 2.0 MiB, which is (relatively) significant.

This PR moves `NewClient` to a new file which is omitted from the build if a new `pkg_sftp.omit_ssh` build tag is set during the build, so users of the library that don't use `x/crypto/ssh` can omit that dependency.

### DIFF
--- a/client.go
+++ b/client.go
@@ -16,7 +16,6 @@ import (
 	"time"
 
 	"github.com/kr/fs"
-	"golang.org/x/crypto/ssh"
 
 	"github.com/pkg/sftp/internal/encoding/ssh/filexfer/openssh"
 )
@@ -190,34 +189,6 @@ type Client struct {
 	useConcurrentWrites    bool
 	useFstat               bool
 	disableConcurrentReads bool
-}
-
-// NewClient creates a new SFTP client on conn, using zero or more option
-// functions.
-func NewClient(conn *ssh.Client, opts ...ClientOption) (*Client, error) {
-	s, err := conn.NewSession()
-	if err != nil {
-		return nil, err
-	}
-
-	pw, err := s.StdinPipe()
-	if err != nil {
-		return nil, err
-	}
-	pr, err := s.StdoutPipe()
-	if err != nil {
-		return nil, err
-	}
-	perr, err := s.StderrPipe()
-	if err != nil {
-		return nil, err
-	}
-
-	if err := s.RequestSubsystem("sftp"); err != nil {
-		return nil, err
-	}
-
-	return newClientPipe(pr, perr, pw, s.Wait, opts...)
 }
 
 // NewClientPipe creates a new SFTP client given a Reader and a WriteCloser.
--- a/client_ssh.go
+++ b/client_ssh.go
@@ -1,0 +1,33 @@
+//go:build !pkg_sftp.omit_ssh
+
+package sftp
+
+import "golang.org/x/crypto/ssh"
+
+// NewClient creates a new SFTP client on conn, using zero or more option
+// functions.
+func NewClient(conn *ssh.Client, opts ...ClientOption) (*Client, error) {
+	s, err := conn.NewSession()
+	if err != nil {
+		return nil, err
+	}
+
+	pw, err := s.StdinPipe()
+	if err != nil {
+		return nil, err
+	}
+	pr, err := s.StdoutPipe()
+	if err != nil {
+		return nil, err
+	}
+	perr, err := s.StderrPipe()
+	if err != nil {
+		return nil, err
+	}
+
+	if err := s.RequestSubsystem("sftp"); err != nil {
+		return nil, err
+	}
+
+	return newClientPipe(pr, perr, pw, s.Wait, opts...)
+}


puellanivis (commented):
I mean… this doesn’t remove it from the `go.mod` so it’s going to get pulled in anyways?

Dead-code analysis should let this be removed anyways?

If someone has serious concerns about 900 KiB of binary size, then they’ll probably want to vendor/fork and do significantly more manual dead code removal than add an esoteric, and poorly discoverable build flag option?

espadolini:
> Dead-code analysis should let this be removed anyways?

DCE can't change the semantics of the code, and there's enough init time code in `crypto/...` to prevent meaningful DCE.

> I mean… this doesn’t remove it from the go.mod so it’s going to get pulled in anyways?

Adding a build tag seemed less invasive than moving all code except for `NewClient` into a subpackage and reexporting everything from the top-level package. 😅

I'd happily take that as an option tho, it would work way better.

> If someone has serious concerns about 900 KiB of binary size, then they’ll probably want to vendor/fork and do significantly more manual dead code removal

DCE seems to work quite well for sftp - at least in the case of a full featured server such as the one in server_standalone, I'm not really seeing any client code and `github.com/kr/fs` is also fully elided from the build - so it's just this one spurious dependency for `x/crypto/ssh` that's adding to the binary size.

900KiB is not a lot in absolute terms but it's _a third_ of the binary, in the case of server_standalone, so it's just a little unfortunate, that's all.

puellanivis:
I mean, the `standalone_server`  is not actually completely full featured, though I suppose that’s more of a nitpick of details over really a statement on the coverage of it vs `sftp-server`.

> 900KiB is not a lot in absolute terms but it's a third of the binary, in the case of server_standalone, so it's just a little unfortunate, that's all.

I understand that it is a third of the binary, but that scaling does not really follow when someone is including pkg/sftp as a dependency. It’s reasonable to presume that it would be about ~900 KiB regardless of how large one’s binary is, it just so happens that `server_standalone` is so minimal on its own that the `x/crypto/ssh` contributes a significant amount of the binary size.

> Adding a build tag seemed less invasive than moving all code except for NewClient into a subpackage and reexporting everything from the top-level package. 😅

It’s not that I’m fundamentally against improving the package weight when someone doesn’t need an SFTP client, and therefore wouldn’t need to import the whole `ssh` package otherwise. 🤔 Just kind of trying to think of the right way to enable it without presenting a novel and unexpected way of turning the imports off.

Perhaps this is something that could be solved with `v2`? (Before we cement the API and have to support it indefinitely.) Putting the `NewClient` code into its own subdirectory?  That would allow someone to import in the whole SSH in an controlled way? Would just a simple subdirectory be a sufficient disconnect, or would it need to be in a separate submodule?

espadolini:
> Perhaps this is something that could be solved with v2? (Before we cement the API and have to support it indefinitely.) Putting the NewClient code into its own subdirectory? That would allow someone to import in the whole SSH in an controlled way? Would just a simple subdirectory be a sufficient disconnect, or would it need to be in a separate submodule?

That'd be great actually, a different package would even let callers get rid of the indirect dependency in their `go.mod` rather than just not include it in their builds.

espadolini:
@puellanivis would it change your mind if I also mentioned the immense convenience of avoiding `crypto/...` entirely (if possible, which in my usecase it is) so that FIPS 140-3 compliance is automatic? I just realized that for a tiny sftp server binary we're going to ship embedded in a larger binary we can avoid even the last bit of `crypto/rand`, and not having to figure out if we need a FIPS compliant build at all would be really nice.

puellanivis:
https://github.com/pkg/sftp/pull/647 Take a look at the PR proposed for `v2`. I realized we could use interfaces to isolate the `ssh.Client` out entirely, but then realized that it’s going to need generics, because `NewSession` returns a type that would rely on the `ssh` package. But with generics, we can extract that out into an interface for a type constraint.

Now, it’s disconnected from `x/crypto/ssh` unless someone actually passes in a `x/crypto/ssh.Client`. 🎉 

All code works the same as it did before, and anyone not otherwise using `x/crypto/ssh` won’t get it.

puellanivis:
> @puellanivis would it change your mind if I also mentioned the immense convenience of avoiding `crypto/...` entirely (if possible, which in my usecase it is) so that FIPS 140-3 compliance is automatic? I just realized that for a tiny sftp server binary we're going to ship embedded in a larger binary we can avoid even the last bit of `crypto/rand`, and not having to figure out if we need a FIPS compliant build at all would be really nice.

You don’t need to sell me on the idea. I’m down. Just had to find the right way to get it in. The new PR for `v2` is even better, because it _entirely removes_ `x/ssh/crypto` from the main module, not just hiding it behind a build tag. Which should make automatic FIPS compliance even easier sell.

PR #647, opened 2026-04-17
Ioslate `ssh/x/crypto` with generics

puellanivis:


client.go
```diff
@@ -579,9 +577,24 @@ func (cl *Client) getDataBuf(size int) []byte {
...
+	StdoutPipe() (io.Reader, error)
+	StderrPipe() (io.Reader, error)
+
+	RequestSubsystem(subsystem string) error
+
+	Close() error
+	Wait() error
+}
+
+type SSHClient[Session SSHSession] interface {
```
puellanivis, line 591: Have to use a generic type here, because `NewSession` returns a type that is specific to `ssh` package, which would require us to pull in that package as well… but now we have generics, so we can say, “I know it returns a type that matches this interface, so use that as a type constraint.”

examples/go.mod
puellanivis, line 1: We add this sub-module to isolate the import of `x/crypto` because the examples need the import, even though the rest of the module doesn’t.

PR #650, opened 2026-05-15
fix: ineffective assignments

alrs:
This fixes two functions that had a non-pointer receiver, preventing them from effectively updating fields on the `orderedResponse{}` and `orderedRequest{}` structs.

In looking at this, I realize that both of these functions are unused. I'd be happy to take the next step and remove them both.

### DIFF
--- a/packet-manager.go
+++ b/packet-manager.go
@@ -62,8 +62,7 @@ type orderedRequest struct {
 func (s *packetManager) newOrderedRequest(p requestPacket) orderedRequest {
 	return orderedRequest{requestPacket: p, orderid: s.newOrderID()}
 }
-func (p orderedRequest) orderID() uint32       { return p.orderid }
-func (p orderedRequest) setOrderID(oid uint32) { p.orderid = oid }
+func (p orderedRequest) orderID() uint32 { return p.orderid }
 
 type orderedResponse struct {
 	responsePacket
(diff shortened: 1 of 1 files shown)


packet-manager.go
```diff
@@ -62,8 +62,8 @@ type orderedRequest struct {
 func (s *packetManager) newOrderedRequest(p requestPacket) orderedRequest {
 	return orderedRequest{requestPacket: p, orderid: s.newOrderID()}
 }
-func (p orderedRequest) orderID() uint32       { return p.orderid }
-func (p orderedRequest) setOrderID(oid uint32) { p.orderid = oid }
+func (p orderedRequest) orderID() uint32        { return p.orderid }
```
puellanivis: If we’re changing one receiver to a pointer receiver, then we should generally convert all of them to pointer receivers. (Unless there’s a good reason, like how lots of shallow clones are simply value receivers that return a pointer to their receiver… but now with `new(*myPointer)` not sure how useful that pattern is anymore.)

puellanivis (commented):
Great catch! It’s probably because they’re unused that the bug was never spotted.

Just two tiny one-character pieces of lint, and 🤷‍♀️ yeah, if they’re unused, we can at least _try_ deleting them. Being that the type itself is unexported, there’s really no reason to implement setters like this unless it needs to match an interface, and the interface doesn’t contain it… so… yeah. I say, chuck it and then we see what breaks. (Easy enough to put back.)

alrs:
I've rebased the commit to remove the two `orderID()` functions.

PR #655, opened 2026-07-12
harden client reply parsers against malformed server replies

drakkan:
The client response decoders parsed server replies with the unchecked unmarshalUint32/unmarshalString helpers and with fixed-offset reads in unmarshalStatus. A malicious or compromised server could return a truncated or oversized reply and panic the client goroutine with an index- or slice-bounds-out-of-range. Convert these paths to the bounds-checked Safe variants and clamp the data-length before slicing, so a hostile reply returns an error instead of panicking.

Reported by @mjbommar

### DIFF
--- a/client.go
+++ b/client.go
@@ -397,15 +397,25 @@ func (c *Client) ReadDirContext(ctx context.Context, p string) ([]os.FileInfo, e
 		}
 		switch typ {
 		case sshFxpName:
-			sid, data := unmarshalUint32(data)
-			if sid != id {
-				return nil, &unexpectedIDErr{id, sid}
+			data, err = unmarshalSID(id, data)
+			if err != nil {
+				return nil, err
 			}
-			count, data := unmarshalUint32(data)
-			for i := uint32(0); i < count; i++ {
+			var count int
+			count, data, err = unmarshalCount(data)
+			if err != nil {
+				return nil, err
+			}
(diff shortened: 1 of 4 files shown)


client.go
```diff
@@ -397,15 +397,25 @@ func (c *Client) ReadDirContext(ctx context.Context, p string) ([]os.FileInfo, e
...
+				return nil, err
+			}
+			var count int
+			count, data, err = unmarshalCount(data)
+			if err != nil {
+				return nil, err
 			}
-			count, data := unmarshalUint32(data)
-			for i := uint32(0); i < count; i++ {
+			for i := 0; i < count; i++ {
```
puellanivis, line 409: We can switch this to `range i` since we’re here.

client_test.go
```diff
@@ -181,6 +181,272 @@ func TestClientShortPacket(t *testing.T) {
...
+	}
+	n, _ := unmarshalUint32(hdr)
+	body := make([]byte, n)
+	if _, err := io.ReadFull(r, body); err != nil {
+		return 0, 0, false
+	}
+	typ = fxp(body[0])
+	if len(body) >= 5 {
+		id, _ = unmarshalUint32(body[1:])
+	}
```
puellanivis, line 207: I think if this is false, then this would be an error… at least a `ok == false` situation.

PR #657, opened 2026-07-23
expose the SSH_FXP_READDIR longname

mickael-kerjean:
Each entry in a SSH_FXP_NAME reply carries three fields: filename, longname and attrs ([source](https://datatracker.ietf.org/doc/html/draft-ietf-secsh-filexfer-02#section-7)). The client currently ignore the longname field with:
```
_, data = unmarshalString(data) // discard longname
```

I have a use case which require to display things similarly to what the open ssh cli shows:
```
sftp> ls -l
drwxrwxr-x    ? mickael  mickael      4096 Jul  3 18:34 Android
drwxr-xr-x    ? mickael  mickael      4096 Jun  3 00:03 Desktop
drwxr-xr-x    ? mickael  mickael      4096 Jul  8 14:14 Documents
```
As of today, this is not possible and this PR fixes that

### DIFF
--- a/attrs.go
+++ b/attrs.go
@@ -54,6 +54,7 @@ type FileStat struct {
 	UID      uint32
 	GID      uint32
 	Extended []StatExtended
+	Longname string
 }
 
 // ModTime returns the Mtime SFTP file attribute converted to a time.Time
--- a/client.go
+++ b/client.go
@@ -405,12 +405,14 @@ func (c *Client) ReadDirContext(ctx context.Context, p string) ([]os.FileInfo, e
 			for i := uint32(0); i < count; i++ {
 				var filename string
 				filename, data = unmarshalString(data)
-				_, data = unmarshalString(data) // discard longname
+				var longname string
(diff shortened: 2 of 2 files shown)


attrs.go
```diff
@@ -54,6 +54,7 @@ type FileStat struct {
 	UID      uint32
 	GID      uint32
 	Extended []StatExtended
+	Longname string
```
puellanivis, line 57: `longname` is not a part of this structure, it’s a part of the `SSH_FXP_NAME` name entry sub-structure: https://datatracker.ietf.org/doc/html/draft-ietf-secsh-filexfer-02#section-7

If you have need to access this and want to get it added in, it should be put in the `fileInfo` struct: https://github.com/pkg/sftp/blob/fc82c354c0d87349411e30a08bef297c9f132105/attrs.go#L23-L26 and then we can expose a `LongName() string` receiver method that you can then test with `.(interface{ LongName() string })`  to use it.

PR #660, opened 2026-08-25
fix: don't silently drop data on short reads in concurrent WriteTo

### LINKED ISSUE #658: `File.WriteTo` silently drops data on short reads when the server's max packet size is smaller than the client's
**Version:** v1.13.10 (also affects earlier versions with the concurrent `WriteTo` implementation)

**Description**

When a client is configured with `MaxPacketUnchecked(N)` for N larger than the server's `maxTxPacket` (e.g. a `pkg/sftp` server at its 32 KiB default), the concurrent path of `File.WriteTo` — which is what `io.Copy` uses — silently skips data instead of detecting the short reads:

- The dispatcher issues `SSH_FXP_READ` requests for `chunkSize` (= client `maxPacket`) bytes at offsets spaced `chunkSize` apart, and advances the offset unconditionally (`client.go`, the "Slice" goroutine in `WriteTo`).
- The server clamps each response to its `maxTxPacket`, a legal short read.
- The worker accepts whatever length arrives without comparing it to the requested length (`client.go`, `case sshFxpData:` in the `WriteTo` map phase), and the reduce phase writes it and moves on.

With a 128 KiB client packet size against a 32 KiB server, every chunk delivers only its first 32 KiB and the remaining 96 KiB is skipped. The copy returns **exactly 1/4 of the file with a nil error** — truncated and corrupted output with no indication of failure. (The sequential paths are not affected: `Read`/`writeToSequential` treat a short read as EOF via `readChunkAt`, and `ReadFrom` sizes writes from the data it has.)

Found in the wild: Incus clients use `MaxPacketUnchecked(128*1024)` against `pkg/sftp`-based servers with the default `maxTxPacket`; `incus file pull` of a ~3 GiB file "completed" with ~768 MiB of corrupt output.

**Reproduction** (self-contained, v1.13.10):

```go
package main

import (
	"crypto/rand"
	"fmt"
	"io"
	"net"
	"os"

	"github.com/pkg/sftp"
)

func main() {
	data := make([]byte, 4*1024*1024)
	rand.Read(data)
	tmp, _ := os.CreateTemp("", "sftp-repro-")
	defer os.Remove(tmp.Name())
	tmp.Write(data)
	tmp.Close()

	c1, c2 := net.Pipe()
	go func() {
		server, _ := sftp.NewServer(c2) // default maxTxPacket = 32768
		server.Serve()
	}()

	client, _ := sftp.NewClientPipe(c1, c1, sftp.MaxPacketUnchecked(128*1024))
	f, _ := client.Open(tmp.Name())

	n, err := io.Copy(io.Discard, f) // delegates to File.WriteTo
	fmt.Printf("source size: %d\ncopied:      %d\nerror:       %v\n", len(data), n, err)
}
```

Output:

```
source size: 4194304
copied:      1048576
error:       <nil>
```

**Expected behavior**

A short read mid-file should never lose data silently. `WriteTo` should either re-request the remainder of the chunk, or return an error. Returning fewer bytes than the file contains with `err == nil` is the worst possible outcome for a copy primitive.

I understand `MaxPacketUnchecked` is documented as "use at your own risk", but the failure mode here is silent corruption rather than an error, and it only manifests on the concurrent code path — the sequential paths handle the same server behavior safely.

ChrisJr404:
Fixes #658.

The concurrent `File.WriteTo` path (what `io.Copy` uses) dispatches reads at offsets spaced `chunkSize` apart and advances the offset unconditionally. Since a server is free to return a short read, and does whenever the client max packet size is larger than the server's, every chunk kept only its first short read and the rest of that chunk was skipped. The copy came back truncated with a nil error, so it was silent corruption rather than an error.

The sequential paths already handle this by looping in `readChunkAt`. This does the same thing in the worker: when the pre-dispatched read comes back short, fill the rest of the chunk before handing it on, so the pre-computed offsets stay aligned. A genuine EOF ends the fill and is carried through as the chunk's error. The full-read path (server returns the whole chunk) is unchanged.

Added a regression test that opens a file through an in-memory server with a smaller max packet size than the client. It fails on master (WriteTo returns ~1/4 of the file) and passes with this change. `go test`, `go test -race` and `go vet` are green on the package.


client.go
```diff
@@ -1528,8 +1527,24 @@ func (f *File) WriteTo(w io.Writer) (written int64, err error) {
...
+							// allowed to return fewer bytes than requested, e.g.
+							// when it clamps to a smaller max packet size) would
+							// leave a hole. Fill the rest of the chunk before
+							// handing it on, the same way the sequential path
+							// does, or we silently drop the skipped bytes. A
+							// genuine EOF ends the fill and is carried as the
+							// chunk's error.
+							if n < chunkSize {
+								var m int
+								m, err = f.readChunkAt(readWork.res, b[n:], readWork.off+int64(n))
```
puellanivis: The problem is this is that we’ve already dispatched reads to later offsets, catch up reads like this result in non-sequential access, which has been a result of various bugs here and there.

Namely, some ssh implementation automatically delete a file once the whole file has been read the first time, this backfilling catch up would break under this situation.

Also, backing up and backfilling can cause significant performance degradation as servers are tuned for sequential read access, not really for random access.

ChrisJr404:
Thanks, that is a good catch and not a case I had considered. If a catch-up read backfills an earlier offset after later reads have already been dispatched, that breaks the sequential-access assumption, and on servers that drop a file once it has been fully read the backfill would fail outright. That is worse than the short-read gap I was trying to close.

Given that, silently backfilling is the wrong fix. Would you prefer I detect the short read and return an explicit error to the caller rather than trying to paper over it, or is there an existing pattern here you would rather I follow? Happy to rework it in whichever direction you think is safe.

puellanivis:
Yes, we should prefer returning an explicit error.

ChrisJr404:
Agreed, backfilling was the wrong fix since the catch-up reads reintroduce exactly the non-sequential access you flagged. Reworked it: the worker code is back to the original, and the reduce step now tracks whether a chunk came back short. If a short read is followed by another chunk that still has data, the fixed-offset stream can no longer be reassembled, so WriteTo returns an explicit error rather than silently dropping the skipped bytes. A short final chunk is unaffected since it is only followed by the EOF packet. Updated the test to assert the loud failure and that whatever was written is a correct prefix of the source. Thanks for the guidance.

client.go
```diff
@@ -1559,12 +1559,23 @@ func (f *File) WriteTo(w io.Writer) (written int64, err error) {
...
 		if !ok {
 			return written, errors.New("sftp.File.WriteTo: unexpectedly closed channel")
 		}
 
+		// The reads are dispatched at fixed offsets (off, off+chunkSize, ...), so a
+		// server returning fewer bytes than requested mid-stream leaves a gap that the
+		// following chunk cannot fill. If a short read is followed by another chunk that
+		// still has data, the stream can no longer be reassembled, so fail loudly rather
+		// than silently drop the skipped bytes. A short final chunk is fine: it is
+		// followed only by the EOF packet, which carries no data.
```
puellanivis, line 1574: [style] Do not break lines just to keep a specific line length: https://go.dev/wiki/CodeReviewComments#line-length

If you’re looking for guidance on how I recommend breaking up comments into multiple lines, I recommend [Semantic Breaking](https://sembr.org/). The additional advantage of semantic breaking of comments is that editing of a single idea in the comment will not cause a cascade formatting change of the rest of the comment, which generates unnecessary PR churn.

client.go
```diff
@@ -1559,12 +1559,23 @@ func (f *File) WriteTo(w io.Writer) (written int64, err error) {
...
+		if shortRead && len(packet.b) > 0 {
+			return written, errors.New("sftp: server returned a short read mid-stream, cannot reassemble concurrent WriteTo")
```
puellanivis, line 1576: I would suggest better `"concurrent WriteTo was aborted"` rather than “cannot reassemble concurrent WriteTo” The later says nothing especially useful about what happened, the former says definitively: I refused to continue under these conditions.

client_test.go
```diff
@@ -198,3 +198,70 @@ func TestClientNoSid(t *testing.T) {
...
+	}
+	go server.Serve()
+
+	client, err := NewClientPipe(cr, cw, MaxPacketUnchecked(128*1024))
+	if err != nil {
+		t.Fatal(err)
+	}
+	// Close the client first (LIFO), so its receive loop sees the server go away.
+	defer client.Close()
+	defer server.Close()
```
puellanivis, line 230: This comment is misleading. `defer` statements are executed as noted in LIFO order. So, we are _not_ closing the client first, we are setting up the client to close after the server.

Rework this to a `defer func() { …  }()` so that the proper intended ordering off the instruction is both clear, and identical to the natural order a human would expect events to happen in.

client_test.go
```diff
@@ -198,3 +198,70 @@ func TestClientNoSid(t *testing.T) {
...
+// dispatches reads at fixed offsets, it cannot reassemble the stream across a
+// mid-stream short read, so it must fail loudly rather than return a truncated
+// copy that looks successful.
+func TestClientWriteToShortReads(t *testing.T) {
+	cr, sw := io.Pipe()
+	sr, cw := io.Pipe()
+
+	// The default server max packet size is 32768, so a bigger client packet
+	// size makes every read come back short.
+	server, err := NewServer(struct {
```
puellanivis, line 215: Since this test runs against a server, it is properly an integration test, not a quick test. Since it only runs against our own server implementation, it should also not run when we’re running with `-testserver`.

client_test.go
```diff
@@ -198,3 +198,70 @@ func TestClientNoSid(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer f.Close()
+
+	var buf bytes.Buffer
+	n, err := f.WriteTo(&buf)
+	if err == nil {
+		t.Fatalf("WriteTo succeeded but should have reported the short read; wrote %d of %d bytes", n, len(want))
+	}
```
puellanivis, line 261: We’ve setup the _current_ conditions that would cause this error to arise, but we’re inherently relying here upon the implementation of the server.

Reasonably, there should be a fix to the server that ensures we do not generally return short packets. In such a case, this test would then fail, because the underlying issue was addressed on both ends: the client returns an error on a short read, but the server also does not return short reads in the first place.

puellanivis (commented):
As an additional comment. The language of responses has hinted at some amount of LLM contribution. 

We haven’t yet established any specific policy on LLM contributions for this specific package, but I still recommend that people follow https://docs.kernel.org/process/coding-assistants.html when contributing any amount of LLM code in general. Specifically, calling out `Assisted-By:` in the commit, so that contributions are properly documented.

FYI: Comments in the review have been addressed with the presumption that they will be acted on by an LLM. If an LLM is used to respond back, do not repeat shared context, do not answer in full formal essay responses. Present your new information and only new information concisely, like one would to someone who already knows what’s going on.

PR #328, opened 2019-12-24
Fix panics when attempting to get file attributes from too small byte slices

vansante:
Fixes #325

Issue #325, opened 2019-12-03
Request.Attributes() can panic when called with the Attr slice empty


vansante:
As mentioned in the title, I got a panic in production.

Is this supposed to happen and am I supposed to check the byte slice myself prior to calling `Attributes()`, or can we just check that in the function (that would be my preference) ? Or is the byte slice always supposed to be filled? (My guess this is some weird client that did not comply to the SFTP standards).

Panic here:

```
panic: runtime error: index out of range [3] with length 0

goroutine 261287 [running]:
github.com/pkg/sftp.unmarshalUint32(...)
        /go/pkg/mod/github.com/pkg/sftp@v1.10.1/packet.go:73
github.com/pkg/sftp.getFileStat(0x138, 0xc0016125df, 0x4, 0x4, 0x3, 0x8, 0x30, 0x8)
        /go/pkg/mod/github.com/pkg/sftp@v1.10.1/attrs.go:118 +0x53d
github.com/pkg/sftp.(*Request).Attributes(...)
        /go/pkg/mod/github.com/pkg/sftp@v1.10.1/request-attrs.go:61
source-dir/sftp.(*RequestHandler).handleSetStat(0xc001324cc0, 0xc000dcd620, 0x6, 0x6, 0xc0006fe540, 0xc0011d1030, 0x6, 0x6)
        /source-dir/sftp/request_handler.go:296 +0x65
source-dir/sftp.(*RequestHandler).Filecmd(0xc001324cc0, 0xc0006fe540, 0xc00045bba0, 0x10)
        /source-dir/sftp/request_handler.go:276 +0x6e2
github.com/pkg/sftp.filecmd(0xfd6d00, 0xc001324cc0, 0xc0006fe540, 0xfe00c0, 0xc0023aba40, 0xfdf180, 0x40df9b)
        /go/pkg/mod/github.com/pkg/sftp@v1.10.1/request.go:267 +0xd4
github.com/pkg/sftp.(*Request).call(0xc0006fe540, 0xfd6d40, 0xc001324cc0, 0xfd6d60, 0xc001324cc0, 0xfd6d00, 0xc001324cc0, 0xfd6d20, 0xc001324cc0, 0xfe00c0, ...)
        /go/pkg/mod/github.com/pkg/sftp@v1.10.1/request.go:162 +0x389
[...]

PR #662, opened 2026-09-19
fix: do not dereference nil attributes in the example Filecmd

### LINKED ISSUE #661: Example server panics on a SETSTAT whose attributes are shorter than its flags promise (request-example.go:160)
### Summary

The example server shipped in `request-example.go` dereferences `Request.Attributes()` without a nil check, so a single malformed `SSH_FXP_SETSTAT` packet panics the handler goroutine and takes the process down.

`Request.Attributes()` returns `nil` when the attribute blob does not hold what the flags promise — that behaviour is deliberate, from #325 / #328, which replaced a panic inside the parser with a `nil` return. The example does not account for it:

https://github.com/pkg/sftp/blob/master/request-example.go#L159-L161

```go
if r.AttrFlags().Size {
	return file.Truncate(int64(r.Attributes().Size))
}
```

A client that sets `SSH_FILEXFER_ATTR_SIZE` but sends fewer than 8 bytes of attributes reaches that line with `Attributes() == nil`. The panic happens in a `packetWorker` goroutine, so it cannot be recovered by the caller: the whole server process ends. Anyone who used the example as a starting point for their own `Filecmd` inherits it — that is how I found it, through fuzzing my own server.

### Reproducing

Against v1.13.11, driving `InMemHandler()` over `net.Pipe`:

```go
// OPEN /a.txt (WRITE|CREAT), WRITE one byte, CLOSE, then:
// SETSTAT /a.txt with flags = SSH_FILEXFER_ATTR_SIZE and only 4 bytes of attributes.
body := append(str("/a.txt"), binary.BigEndian.AppendUint32(nil, 0x00000001)...)
body = append(body, 0, 0, 0, 0)
conn.Write(packet(9 /* SSH_FXP_SETSTAT */, 4, body))
```

Output:

```
INIT -> type 2
OPEN -> type 102
WRITE -> type 101
CLOSE -> type 101
[signal 0xc0000005 code=0x0 addr=0x0 pc=0x7ff6ceb1d2f5]

goroutine 20 [running]:
github.com/pkg/sftp.(*root).Filecmd(0x2597acbda9f0, 0x2597ace1e000)
	github.com/pkg/sftp@v1.13.11/request-example.go:160 +0x4b5
github.com/pkg/sftp.filecmd({...}, 0x2597ace1e000, {...})
	github.com/pkg/sftp@v1.13.11/request.go:507 +0x157
github.com/pkg/sftp.(*Request).call(...)
	github.com/pkg/sftp@v1.13.11/request.go:312 +0x391
github.com/pkg/sftp.(*RequestServer).packetWorker(...)
	github.com/pkg/sftp@v1.13.11/request-server.go:318 +0xc4a
```

The full program is a single file; I can attach it if that is useful.

### Environment

- github.com/pkg/sftp v1.13.11 (also present on master at the time of writing)
- go1.27.1, windows/amd64

### Suggested fix

Two small changes, which I would be glad to send as a PR:

1. `request-example.go`: return an error (`sftp.ErrSSHFxBadMessage`, or `os.ErrInvalid`) when `Attributes()` is nil, instead of dereferencing it.
2. `request-attrs.go`: say in the doc comment on `Attributes()` that it returns nil when the attributes are malformed, so callers know they must check. The current comment ("parses file attributes byte blob and return them in a FileStat object") gives no hint, and the example reinforces the unchecked usage.

While reporting this: the related unbounded pre-allocation in `unmarshalFileStat` is already fixed by c1bc2ff and released in v1.13.11, so this issue is only about the nil dereference.

labeedahmad-debug:
Fixes #661.

`Request.Attributes()` returns nil when the attribute blob does not hold every field its flags promise. That is deliberate — #325 / #328 replaced a panic inside the parser with a nil return — but `request-example.go` dereferences it:

```go
if r.AttrFlags().Size {
	return file.Truncate(int64(r.Attributes().Size))
}
```

A `SSH_FXP_SETSTAT` that sets `SSH_FILEXFER_ATTR_SIZE` and carries fewer than eight bytes of attributes reaches that line with nil. The panic happens in a `packetWorker` goroutine, so the caller cannot recover it and the process ends. I found it by fuzzing a server whose `Filecmd` started from this example.

### What this changes

- `request-example.go`: answer `ErrSSHFxBadMessage` when `Attributes()` is nil, instead of dereferencing it.
- `request-attrs.go`: say in the doc comment that `Attributes()` returns nil for malformed attributes, so callers know they have to check. The comment gave no hint, and the example taught the unchecked pattern.
- `request-example_test.go` (new): a regression test that drives the example handler with a short attribute blob. It panics without the fix and passes with it. It calls the handler directly rather than through `clientRequestServerPair`, so it also runs on Windows and Plan 9.

The library's own parsing is unchanged; only the example and a doc comment.

### Verification

```
go test -run TestInMemHandlerSetstatShortAttributes -v .
=== RUN   TestInMemHandlerSetstatShortAttributes
--- PASS: TestInMemHandlerSetstatShortAttributes (0.00s)
```

Reverting only the `request-example.go` hunk makes that test panic at `request-example.go:160`, which is the reported crash.

`go test ./...` is otherwise unchanged by this PR. Note that `TestCleanPath` already fails on master on windows/amd64, before and after this change, so it looks unrelated and platform-specific.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


request-example_test.go
```diff
@@ -0,0 +1,24 @@
...
+
+// A SETSTAT whose flags promise a size the packet does not carry leaves Attributes nil. The
+// example handler must answer it, not dereference the nil and take the process down with it.
+func TestInMemHandlerSetstatShortAttributes(t *testing.T) {
+	fs := InMemHandler().FileCmd.(*root)
+
+	_, err := fs.Filewrite(&Request{Method: "Put", Filepath: "/foo", Flags: sshFxfWrite | sshFxfCreat})
+	require.NoError(t, err)
+
+	err = fs.Filecmd(&Request{
```
puellanivis: There is no reason to type assert out the `*root` here. We can just use the `FilePut` and ̀ FileCmd` fields directly:
```suggestion
	fs := InMemHandler()

	_, err := fs.FilePut.Filewrite(&Request{Method: "Put", Filepath: "/foo", Flags: sshFxfWrite | sshFxfCreat})
	require.NoError(t, err)

	err = fs.FileCmd.Filecmd(&Request{
```

request-example.go
```diff
@@ -157,7 +157,13 @@ func (fs *root) Filecmd(r *Request) error {
...
 
 		if r.AttrFlags().Size {
-			return file.Truncate(int64(r.Attributes().Size))
+			attrs := r.Attributes()
+			if attrs == nil {
+				// The attribute flags promise fields the packet does not carry.
+				return ErrSSHFxBadMessage
+			}
+
+			return file.Truncate(int64(attrs.Size))
```
puellanivis: I’m going to recommend we provide a better example here, because if one were supporting more than one setstat operation, it would need to be repeated. Instead, let’s test the attributes up front:
```go
		attrs := r.Attributes()
		if attrs == nil {
			// Something went wrong parsing attributes
			return ErrSSHFxBadMessage
		}

		if r.AttrFlags().Size {
			return file.Truncate(int64(attrs.Size))
		}
```

puellanivis (commented):
Please do not use `Co-Authored-By` for an LLM. Legally, an LLM cannot be an author in any copyright jurisdiction that I am aware of. Instead, it is suggested to use `Assisted-by:`: https://docs.kernel.org/process/coding-assistants.html#attribution

